PR #22, opened 2015-07-28
Cfservice

ramirezg:
- Added the cloudfoundry service to make code cleaner and facilitate testing
- using karma and jasmine for unit testing
- setup angular code to be tested with travis
- testing is still incomplete, but this PR puts the framework in place.


static/js/app/cloudfoundry.js
```diff
@@ -0,0 +1,57 @@
+(function() {
+    // CloudFoundry Service
+    var cloudfoundry = function($http) {
+
+        // Get current authentication status from server
+        var getAuthStatus = function() {
+            return $http.get('/v2/authstatus')
+                .then(function(response) {
```
jcscottiii: is this like success and failure? if they do they the same thing, do we need two different anon functions?

static/js/app/cloudfoundry.js
```diff
@@ -0,0 +1,57 @@
...
+
+        // Get organizations
+        var getOrgs = function() {
+            return $http.get('/v2/organizations')
+                .then(function(response) {
+                    return response.data.resources;
+                });
+        };
+        // Get organization spaces details
+        var getOrgSpaceDetails = function(org) {
```
jcscottiii: is it possible for org or any of it's member variables to be null?

static/js/app/cloudfoundry.js
```diff
@@ -0,0 +1,57 @@
...
+            return $http.get(org.entity.spaces_url)
+                .then(function(response) {
+                    var data = response.data;
+                    data.org_name = org.entity.name;
+                    return data;
+                });
+        };
+        // Get space details
+        var getSpaceDetails = function(space) {
```
jcscottiii: is it possible for space or any of it's member variables to be null?

static/js/app/cloudfoundry.js
ramirezg: `org` should not be null because it's populated from [a list of orgs](https://github.com/18F/cf-console/blob/master/static/js/app/partials/dashboard.html#L20). Without an org the element that calls this function doesn't exist. Also, I think every org contains the elements requested.

static/js/app/cloudfoundry.js
ramirezg: This should work the same way the `org` variable works. [List of spaces](https://github.com/18F/cf-console/blob/cfservice/static/js/app/partials/spaces.html#L2)

static/js/app/cloudfoundry.js
ramirezg: Yep the first is for a success and the second is for a failure. In this case we need two of the same functions because we want the site to return the status in both cases. Still, I cleared up the code a bit.

PR #29, opened 2015-07-30
Marketplace

ramirezg:
- PR from [UI: View service marketplace](https://trello.com/c/zymQs8QO/168-ui-view-service-marketplace)

screen shot 2015-08-05 at 11 07 51 am

screen shot 2015-08-05 at 11 09 17 am

### DIFF
--- a/acceptance/app_structure_test.go
+++ b/acceptance/app_structure_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/sclevine/agouti"
 	. "github.com/sclevine/agouti/matchers"
 
+	//"fmt"
 	"net/http/httptest"
 )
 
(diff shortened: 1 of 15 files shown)


jcscottiii:
Looking good! :+1: :)

acceptance/app_structure_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/sclevine/agouti"
 	. "github.com/sclevine/agouti/matchers"
 
+	//"fmt"
```
jcscottiii: remove

static/js/app/controllers.js
```diff
@@ -19,8 +19,18 @@
         };
         // Get data for a specific org
         $scope.showOrg = function(org) {
-            $location.path('/dashboard/org/' + org.metadata.guid);
+            if (org.metadata)
```
jcscottiii: what causes the case for when the guid is not in the org.metadata object?

static/js/app/controllers.js
```diff
@@ -19,8 +19,18 @@
...
+                $location.path('/dashboard/org/' + org.metadata.guid);
+            else
+                $location.path('/dashboard/org/' + org.guid);
         };
+        $scope.showOrgMarketplace = function(org) {
```
jcscottiii: similar from above

static/js/app/controllers.js
jcscottiii: just a comment can help out

PR #30, opened 2015-07-30
Individual pages

ramirezg:
- creating pretty searchable tables for spaces and orgs
- setting up system for efficiently rendering individual pages for orgs. 
tldr - org data from the `/v2/organizations` endpoint is only loaded on entry, if prompted, or if an error occurs. This should keep costly api calls down especially when we create space/app/service pages.


static/js/app/cloudfoundry.js
```diff
@@ -1,6 +1,7 @@
 (function() {
     // CloudFoundry Service
-    var cloudfoundry = function($http, $location) {
+    angular.module('cfconsole').service('$cloudfoundry', function($http, $location, $log) {
```
jcscottiii: change this to cfdeck

static/index.html
```diff
@@ -22,7 +22,7 @@
 			<div class="site-wrapper-inner">
 				<div class="container">
 				<!-- Where our view will load -->
-	    		<div ng-view></div>
+	    		<div ng-view ng-controller="DashboardCtrl"></div>
```
jcscottiii: Why are you putting this into the base file and not the partials/index.html?

static/js/app/controllers.js
```diff
@@ -11,61 +11,62 @@
...
-                });
+    app.controller('OrgCtrl', function($scope, $cloudfoundry, $location, $routeParams) {
+        // Display Space
+        $scope.showSpace = function(space) {
+            $location.path($location.path() + '/spaces/' + space.guid)
+        };
+        // Render the org information on the page
+        var renderOrg = function(orgData) {
+            if (orgData['code'] == 30003) {
+                $scope.activeOrg = "404";
```
jcscottiii: Comment as to what the 30003 is and why we should set activeOrg to 404 and what does sending a 404 do.

jcscottiii: guessing it has to do with https://github.com/18F/cf-deck/pull/30/files#diff-60bb1f6b95d957acd2a186fed9e2db20R3

static/js/app/partials/organizations.html
```diff
@@ -1,12 +1,75 @@
...
-    <div ng-switch-default>
-        <ng-include src="'js/app/partials/spaces.html'"></ng-include>
+    <div ng-switch-when="404">
+        <div class="page-header">
+            <h3 class="text-center">Sorry, we couldn't find that org</h3>
+        </div>
+    </div>
+
+    <div ng-switch-when="undefined">
+        <!-- show nothing while loading -->
```
jcscottiii: In the future, we could do a loading spinner. But for now, put something like "Loading..."

static/js/app/partials/organizations.html
```diff
@@ -1,12 +1,75 @@
...
+                        <td>{{space.name}}</td>
+                        <td>{{space.app_count}}</td>
+                        <td>{{space.mem_dev_total}}</td>
+                        <td>{{space.mem_prod_total}}</td>
+                    </tr>
+                </tbody>
+            </table>
+        </div>
+        <div class="text-center" ng-show="spaces.length == 0">
+            <h2>This org has no spaces</h2>
```
jcscottiii: org -> organization

static/js/app/partials/organizations.html
ramirezg: let's use ng-cloak

static/js/app/partials/organizations.html
jcscottiii: org -> organization

static/js/app/controllers.js
ramirezg: yep

ramirezg:
@jcscottiii This should be done for now. The coverage on the front end dropped a little but we can add the small tests later.

PR #37, opened 2015-08-05
Create service instances

ramirezg:
- still working on testing, language, and sending request to make the service 

screen shot 2015-08-05 at 2 06 13 pm

### DIFF
--- a/static/js/app/controllers.js
+++ b/static/js/app/controllers.js
@@ -84,9 +84,9 @@
     app.controller('MarketCtrl', function($scope, $cloudfoundry, $location, $routeParams) {
         // Render the active org
         var renderActiveOrg = function(org) {
-            $scope.activeOrg = org;
-        }
-        // Show a specific service details by going to service landing page
+                $scope.activeOrg = org;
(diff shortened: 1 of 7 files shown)


ramirezg:
@jcscottiii this should be done for now. It's not the best looking form, but I figured we could fix it later.

static/js/app/controllers.js
```diff
@@ -84,9 +84,9 @@
     app.controller('MarketCtrl', function($scope, $cloudfoundry, $location, $routeParams) {
         // Render the active org
         var renderActiveOrg = function(org) {
-            $scope.activeOrg = org;
-        }
-        // Show a specific service details by going to service landing page
+                $scope.activeOrg = org;
+            }
```
jcscottiii: spacing

static/js/app/controllers.js
```diff
@@ -84,9 +84,9 @@
...
+            // Show a specific service details by going to service landing page
```
jcscottiii: spacing

PR #38, opened 2015-08-05
App Live Status and App Actions Buttons

jcscottiii:


jcscottiii:
image

ramirezg:
I really like the live data. :+1:

static/js/app/cloudfoundry.js
```diff
@@ -139,6 +139,64 @@
...
+	};
+	// Wrapper function that will submit a request to stop an app.
+	this.stopApp = function(app) {
+		return this.changeAppState(app, "STOPPED");
+	};
+	// Wrapper function that will submit a request to restart an app.
+	this.restartApp = function(app) {
+		return this.changeAppState(app, "STOPPED")
+			.then(function() {
+				return changeAppState(app, "STARTED");
```
ramirezg: The restart function isn't working for me b/c `changeAppState` isn't defined. I think you'll have to declare `ChangeAppState` with var in order to make it accessible to the scope on line 197. Like below: 

```
	// Internal generic function that actually submits the request to backend to change the app.
	var changeAppState = function(app, desired_state) {
		setPollAppStatusProperty(false); // prevent UI from refreshing.
		return $http.put("/v2/apps/" + app.guid + "?async=false&inline-relations-depth=1", {"state":desired_state})
			.then(function(response) {
				// Success
				// Set the state immediately to stop so that UI will force a load of the new options.
				// UI will change the buttons based on the state.
				app.state = desired_state;
			}, function(response) {
				// Failure
			}).finally(function() {
				setPollAppStatusProperty(true); // allow UI to refresh via polling again.
			});
	}
	// Wrapper function that will submit a request to start an app.
	this.startApp = function(app) {
		return changeAppState(app, "STARTED");
	};
	// Wrapper function that will submit a request to stop an app.
	this.stopApp = function(app) {
		return changeAppState(app, "STOPPED");
	};
	// Wrapper function that will submit a request to restart an app.
	this.restartApp = function(app) {
		return changeAppState(app, "STOPPED")
			.then(function() {
				return changeAppState(app, "STARTED");
			});
	};

```

jcscottiii:
Merged into #45

PR #39, opened 2015-08-05
WIP: Move things around

dlapiduz:
🚧🚨🚧🚨🚧🚨🚧🚨 DO NOT MERGE 🚧🚨🚧🚨🚧🚨🚧🚨🚧
WIP


static/app/controllers.js
```diff
@@ -11,23 +11,15 @@
             .then(renderStatus);
     });
 
-    app.controller('DashboardCtrl', function($scope, $cloudfoundry, $location) {
+    app.controller('MainCtrl', function($scope, $cloudfoundry, $location, $rootScope) {
```
jcscottiii: We should try to stay away from using rootscope
http://www.thinkingmedia.ca/2015/01/learn-how-to-use-scopes-properly-in-angularjs/#keep-the-rootscope-clean-of-variables
Instead we should use an injected service.

static/app/controllers.js
```diff
@@ -39,25 +31,23 @@
         $cloudfoundry.getOrgsData(renderOrgs);
     });
 
-    app.controller('OrgCtrl', function($scope, $cloudfoundry, $location, $routeParams) {
-        // Display Space
-        $scope.showSpace = function(space) {
-            $location.path($location.path() + '/spaces/' + space.guid)
-        };
+    app.controller('OrgCtrl', function($scope, $cloudfoundry, $location, $routeParams, $rootScope) {
```
jcscottiii: Same comment as above

static/index.html
```diff
@@ -6,28 +6,29 @@
 		<meta http-equiv="X-UA-Compatible" content="IE=edge">
 		<meta name="viewport" content="width=device-width, initial-scale=1">
 
-		<title>CF-Deck</title>
+		<title>18F Cloud Deck</title>
```
jcscottiii: For the community, we should leave as "Deck"

static/index.html
```diff
@@ -6,28 +6,29 @@
...
+		<script src="app/controllers.js"></script>
+		<script src="app/filters.js"></script>
 	</head>
-	<body>
-		<div class="site-wrapper">
-			<div class="site-wrapper-inner">
-				<div class="container">
-				<!-- Where our view will load -->
-	    		<div ng-view></div>
+	<body ng-controller="MainCtrl">
```
jcscottiii: More of a question: What data / functions do we need from the MainCtrl here? Could a service be used that is injected in the views here?

static/app/views/partials/header.html
```diff
@@ -20,8 +23,8 @@
                     <a class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true"
                     aria-expanded="false" id="org-dropdown">{{ (activeOrg | orgNameFilter) || "Organizations" }}&nbsp;<span class="caret"></span></a>
                     <ul class="dropdown-menu">
-                        <li ng-repeat="org in orgs" ng-click="showOrg(org)">
-                            <a class="org-name">{{org | orgNameFilter}}</a>
+                        <li ng-repeat="org in $parent.orgs">
```
jcscottiii: Should avoid using "$parent.".
These articles talk about ways to avoid having that problem.
http://blog.thoughtram.io/angularjs/2015/01/02/exploring-angular-1.3-bindToController.html
or 
http://toddmotto.com/opinionated-angular-js-styleguide-for-teams/
A little different from each other but accomplish the same thing in the end.

static/index.html
jcscottiii: Hmm, I think you're referring to the data here: https://github.com/18F/cf-deck/pull/39/files#r36405385

static/app/controllers.js
ramirezg: So originally, I depended on root scope for the active org but it made bookmarking/refreshing a problem. For example, if you refresh on a space page the activeOrg will stay undefined because the activeOrg is set with the org controller. Hence, the findActiveOrg method... There might be a better way though.

static/app/views/partials/sidebar.html
```diff
@@ -0,0 +1,18 @@
+<ul class="nav nav-sidebar">
+  <ul class="nav navbar-nav">
+      <li class="dropdown">
```
ramirezg: Sidebar looks great and it does make it easier to navigate. I'm assuming one of the org dropdown will be removed later.

static/index.html
ramirezg: So originally, I had the MainCtrl in here but that forces an call to /v2/organizations before the user is logged in. I like this organization better though.

static/app/services.js
```diff
@@ -0,0 +1,6 @@
+(function() {
+    'use strict';
```
ramirezg: Do you plan on putting the cloudfoundry service here?

static/app/views/partials/header.html
ramirezg: Do we need $parent here? Isn't the navbar called inside `MainCtrl`?

ramirezg:
@dlapiduz I really like how you flattened the code and the side bar makes this app much more usable.

static/index.html
dlapiduz: I think that we shouldn't allow any views without being logged in. We should redirect from the go app to the login page.

static/app/services.js
dlapiduz: To be honest, I had a service here and removed it. But we can keep something like this around so we have somewhere to put services...

static/app/views/partials/header.html
dlapiduz: I think I need it because it is an ng-include.... It is a bit of a hack, we can improve it.

PR #54, opened 2015-08-10
Removing 3rd party libraries

ramirezg:
- All front end dependencies can be installed with 
```
npm install
```
- App now work with a multi-buildpack deploy
- Test cleanup


jcscottiii:
i really like this. try deploying an app to CF to see how much stuff will be downloaded automatically vs how much we will have script to download upon deploying

PR #75, opened 2015-08-17
adding quota usage

ramirezg:
screen shot 2015-08-17 at 3 46 10 pm

### DIFF
--- a/static/app/cloudfoundry.js
+++ b/static/app/cloudfoundry.js
@@ -72,6 +72,23 @@
                 });
         };
 
+        // Get quota usage data
+        this.getQuotaUsage = function(org) {
+            return $http.get('/v2/organizations/' + org.guid + '/memory_usage')
+                .then(function(response) {
+                    var used_memory = response.data.memory_usage_in_mb;
+                    return $http.get('/v2/organizations/' + org.guid)
+                        .then(function (response) {
+                            return $http.get(response.data.entity.quota_definition_url)
+                                .then(function (response) {
+                                   org.quota = response.data;
+                                   org.quota.entity.used_memory = used_memory;          
+                                });
+                        });
+                 });
+        };
+
+
         // Get space details
         this.getSpaceDetails = function(spaceGuid) {
             return $http.get('/v2/spaces/' + spaceGuid + '/summary')
--- a/static/app/cloudfoundry.js
+++ b/static/app/cloudfoundry.js
@@ -72,22 +72,40 @@
                 });
         };
 
-        // Get quota usage data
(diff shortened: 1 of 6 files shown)


static/app/cloudfoundry.js
```diff
@@ -76,6 +76,23 @@
                 });
         };
 
+        // Get quota usage data
+        this.getQuotaUsage = function(org) {
+            $http.get('/v2/organizations/' + org.guid + '/memory_usage')
```
msecret: I think the deep nesting can be solved with using promises a little better:

```
var toReturn = org,
      scope = {};
$http.get('/v2/organizations/' + org.guid + '/memory_usage')
  .then(function(response) {
    toReturn.used_memory = response.data.memory_usage_in_mb;
    return $http.get('/v2/organizations/' + org.guid);
  }).then(function(response) {
    return $http.get(response.data.entity.quota_definition_url);
  }).then(function(response) {
    toReturn.quota = response.data;
    toReturn.quota.entity.used_memory = toReturn.used_memory;
  }).catch(function(error) {
    // Do nothing
  });
```

I didn't get to test that this actually worked, but I got some of the ideas from here: http://solutionoptimist.com/2013/12/27/javascript-promise-chains-2/. I'm also not a promise expert by any means, so somebody else might have even better ideas.

static/app/views/partials/sidebar.html
```diff
@@ -1,7 +1,8 @@
 <ul class="nav nav-sidebar">
+
     <li class="dropdown">
         <a class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true"
-        aria-expanded="false" id="org-dropdown">Organizations&nbsp;<span class="caret"></span></a>
+        aria-expanded="false" id="org-dropdown">Change Organization&nbsp;<span class="caret"></span></a>
```
jcscottiii: Did you try to shrink the window? Did this overflow to two lines?

static/app/cloudfoundry.js
jcscottiii: @msecret great article! We have this kind of deep nesting in a few places. i really like this pattern because of the ease of error handling.

ramirezg: @msecret  :+1: , I'll definitely, keep this article around. Just refactored the code to make it a bit clearer and catch errors.

static/app/views/partials/sidebar.html
ramirezg: Something odd is happening to my side bar, at the small screen it doesn't overflow but it does at the medium.

PR #82, opened 2015-08-23
Fix Memory Leak

jcscottiii:
The console has been refusing requests due to using all the memory available after awhile. This PR should fix that. Also is a cleanup patch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-- 1.4.2
+- 1.5
 before_install:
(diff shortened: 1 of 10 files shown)


jcscottiii:
The memory leak was due to
1) some API calls being lost to the ether despite timeouts
2) some session mishandling

This patch fixes (1) by timing out eventually using Go's TimeoutHandler
and fixes (2) by using gorilla's Context ClearHandler to ensure session
data was released after each incoming request.

The number of goroutines leaked as a result of the external API calls never coming back. Even though 1) is fixed, we also limit the number of goroutines by adding a timeout to the HTTP client

PR #86, opened 2015-08-25
Reorg

ramirezg:
To be merged after #85

Reorganizing frontend to make sure that: 
1. code can be minified correctly
2. Controllers use the controller as format
3. Data is removed from $scope and placed into specific controller scopes.


ramirezg:
@msecret as we talked about earlier, I have no qualms about changing frameworks. But if we do happen to stay with Angular, this might be a worthwhile branch to expand. In short, it's a start of a refactor based on some of the lessons learned from our first iteration.

PR #88, opened 2015-08-26
Revert "Api limit"

jcscottiii:
Reverts 18F/cf-deck#85
A user that may refresh sometimes when data doesn't load fast enough, this patch may cause user to be redirected to the home screen.


jcscottiii:
will reopen #85 cc: @ramirezg

PR #96, opened 2015-09-03
Add status page light to top of page

jcscottiii:


jcscottiii:
Uses a custom directive. now you can just put <a cloud-status></a> anywhere and it will create it for you

PR #106, opened 2015-09-11
Acceptance test updates

ramirezg:


static/app/views/partials/header.html
```diff
@@ -48,7 +48,7 @@
                         </li>
 
                         <li>
-                            <a href="/v2/logout">Logout</a>
+                            <a id="logout" href="/v2/logout">Logout</a>
```
msecret: I want to come up with a simple standard for how we "select" things with our tests. I see in tests above we use FindByName, or FindByButton with the text of the button. Is there a reason we're using an ID for this logout?

PR #120, opened 2015-09-21
Create tests for routes

ramirezg:
Tests creating and deleting a route.


acceptance/routes_test.go
```diff
@@ -0,0 +1,75 @@
...
+		user = StartUserSessionWith(testEnvVars)
+
+		// Log user in
+		user.LoginTo(page)
+	})
+
+	It("Allow the user to create and destroy app routes", func() {
+		By("going to the app page", func() {
+			user.OpenDropdownOfOrgsOn(page)
+			DelayForRendering()
```
jcscottiii: I think these individual calls have delays already in them. are they needed here too?

ramirezg:
looks like the [route test](https://travis-ci.org/18F/cf-deck/builds/81461586#L992) passed but 4 others failed

PR #127, opened 2015-10-09
Adding foxchat to angular code to provide loading after authorization

ramirezg:
This is a temporary solution, until we write a more elegant one.


jcscottiii:
@msecret any thoughts on this?

msecret:
This look fine as a quick hack.

PR #132, opened 2015-10-21
Staging alpha

msecret:


jcscottiii:
To get this to deploy, you'll need to push to the 18F branch of "staging-alpha"

static/index.html
```diff
@@ -1,51 +1,73 @@
-<!DOCTYPE html>
-<html ng-app="cfdeck">
-	<head>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <link rel="stylesheet" type="text/css" href="assets/style.css">
+  </head>
+  <style>
```
jcscottiii: Move this to own css file

static_src/actions/app_actions.js
```diff
@@ -0,0 +1,19 @@
+
+import AppDispatcher from '../dispatcher.js';
+import { appActionTypes } from '../constants';
+
```
jcscottiii: comments

static_src/actions/error_actions.js
```diff
@@ -0,0 +1,9 @@
+
+import errorActionTypes from '../constants.js';
+import Dispatcher from '../dispatcher.js';
+
```
jcscottiii: comments

static_src/actions/login_actions.js
```diff
@@ -0,0 +1,13 @@
+
+import AppDispatcher from '../dispatcher.js';
+import cfApi from '../util/cf_api.js';
+import { loginActionTypes } from '../constants';
+
```
jcscottiii: comments

static_src/actions/org_actions.js
```diff
@@ -0,0 +1,44 @@
+
+import AppDispatcher from '../dispatcher.js';
+import cfApi from '../util/cf_api.js';
+import { orgActionTypes } from '../constants';
+
```
jcscottiii: comments

static_src/actions/org_actions.js
```diff
@@ -0,0 +1,44 @@
...
+  fetch(orgGuid) {
+    AppDispatcher.handleViewAction({
+      type: orgActionTypes.ORG_FETCH,
+      orgGuid: orgGuid
+    });
+  },
+
+  fetchAll() {
+    setTimeout(() => {
```
jcscottiii: What should happen when the timeout expires?

static_src/actions/service_actions.js
```diff
@@ -0,0 +1,33 @@
+
+import AppDispatcher from '../dispatcher.js';
+import { serviceActionTypes } from '../constants';
+
```
jcscottiii: comment

static_src/actions/space_actions.js
```diff
@@ -0,0 +1,22 @@
+
+import AppDispatcher from '../dispatcher.js';
+import cfApi from '../util/cf_api.js';
+import { spaceActionTypes } from '../constants.js';
+
```
jcscottiii: comment

static_src/actions/user_actions.js
```diff
@@ -0,0 +1,19 @@
+
+import AppDispatcher from '../dispatcher.js';
+import { userActionTypes } from '../constants';
+
```
jcscottiii: comment

static_src/constants.js
```diff
@@ -0,0 +1,45 @@
+
+import keymirror from 'keymirror';
+
+
```
jcscottiii: Comment over each var to tell what these constants are used for?

static_src/components/app_list.jsx
```diff
@@ -0,0 +1,91 @@
...
+function stateSetter(props) {
+  var space = SpaceStore.get(props.initialSpaceGuid);
+  return {
+    apps: space && space.apps || [],
+    currentOrgGuid: props.initialOrgGuid,
+    currentSpaceGuid: props.initialSpaceGuid
+  }
+}
+
+export default class AppList extends React.Component {
```
jcscottiii: You have two Lists. This one and the service list. In case we make anymore lists, should he have a parent BaseList class to inherit from to help with extensibility?

static_src/stores/app_store.js
```diff
@@ -0,0 +1,47 @@
+
+import AppDispatcher from '../dispatcher';
+import BaseStore from './base_store.js';
+import cfApi from '../util/cf_api.js';
+import LoginStore from './login_store.js';
+import { appActionTypes } from '../constants.js';
+
+class AppStore extends BaseStore {
```
jcscottiii: A comment at the top of each of these stores to tell me what are the cases / capabilities of each.

static_src/dispatcher.js
```diff
@@ -0,0 +1,19 @@
+
+import {Dispatcher} from 'flux';
+
+class AppDispatcher extends Dispatcher {
+  handleViewAction(action) {
+    action.source = 'VIEW_ACTION';
```
jcscottiii: who actually checks / compares for these different types of actions?

static_src/main.js
```diff
@@ -0,0 +1,82 @@
...
+
+
+function login() {
+  React.render(<App><Login/></App>, mainEl);
+}
+
+function dashboard() {
+  React.render(<App>
+    <h3>Welcome to CF-Deck</h3>
+    <h5>Pick an organization to get started</h5>
```
jcscottiii: A new thing to think about maybe for the future. Let's say someone gets an account created via our self-service thing (This actually happened) and they have 0 orgs since they did that. Should have an ```
if orgs.count <= 0 { "You have zero orgs. Please talk to your org manager to get started")}
else {"Pick an organization to get started"}

static_src/test/global_setup.js
```diff
@@ -0,0 +1,242 @@
...
+      return '';
+    } else if (otherArgs.length > 1) {
+      return jasmine.pp(otherArgs);
+    } else {
+      return jasmine.pp(otherArgs[0]);
+    }
+  }
+};
+
+var sinonMatchers = [
```
jcscottiii: Are using all of these matchers?

jcscottiii: why do we have to translate between sinon and jasmine matchers? why not only have one?

static_src/test/global_setup.js
```diff
@@ -0,0 +1,242 @@
...
+function createJasmineSinonMatchers(matchers) {
+  var matcher, jasmineSinonMatchers = {};
+  for (var i = 0, len = matchers.length; i < len; i++) {
+    matcher = matchers[i];
+    jasmineSinonMatchers[matcher.jasmineName] = createMatcher(matcher);
+  }
+  return jasmineSinonMatchers;
+}
+
+/*
```
jcscottiii: in your final PR, remove this if not being used

static_src/test/unit/actions/service_actions.spec.js
```diff
@@ -0,0 +1,87 @@
...
+
+      serviceActions.receivedInstances(expected);
+
+      expect(spy).toHaveBeenCalledOnce();
+      let arg = spy.getCall(0).args[0];
+      expect(arg.type).toEqual(serviceActionTypes.SERVICE_INSTANCES_RECEIVED);
+      expect(arg.serviceInstances).toEqual(expected);
+    });
+  });
+
```
jcscottiii: we need a case to simulate a failed delete whether it be from it already deleted from someone else or you suddenly lost perms. but just one test case.

static_src/actions/org_actions.js
msecret: I'll explain with a TODO. I had to make the code asynchronous.

static_src/components/app_list.jsx
msecret: Yea I could see that, and we will have more lists.

So there will be a component for each list, like applist, service list, who's responsibility will be to get and manage the data(the apps or services) and pass down to a basic list component. I'm not sure I want an inheritance model here as react usually recommends composition. 

The one issue that I'd have to figure out is some lists have slightly different ways to render columns and rows so I might have abstract that out somehow.

static_src/dispatcher.js
msecret: Right now they're more for debugging purposes. They're also there if the system gets more complex, the stores would potentially use them, but I'm not thinking of that right now.

For now, I like having them in tests because it's being very explicit about where actions are coming from, making the tests more descriptive (and a little easier to make them fail).

static_src/main.js
msecret: Yea I like that. I'll make a note for the design team.

static_src/test/global_setup.js
msecret: So Jasmine comes with some simple matchers. I'm adding on sinon matchers to making writing asserts in tests a little easier but mainly for better error messaging in tests. So the difference:

```
expect(spy.toHaveBeenCalledOnce).toBe(true);
vs
expect(spy).toHaveBeenCalledOnce();
```

The second option will actually tell you "function blahblah was supposed to have been called once but was called 0 times", while the first option will say something like "false is not true".

msecret:
@jcscottiii I think I got all the comments.

jcscottiii:
@msecret looks good!! :100:

PR #134, opened 2015-10-26
389 view all users

msecret:
View commit messages for more information


static_src/constants.js
```diff
@@ -53,10 +53,12 @@ var appActionTypes = keymirror({
...
 var userActionTypes = keymirror({
-  // Action to fetch a single user from the server.
+  // Action to fetch users belonging to a organization from the server.
+  ORG_USERS_FETCH: null,
+  // Action to fetch users belonging to a space from the server.
   SPACE_USERS_FETCH: null,
-  // Action when a user was received from the server.
-  SPACE_USERS_RECEIVED: null
+  // Action when all users were received from the server.
+  USERS_RECEIVED: null
```
jcscottiii: USERS_RECEIVED -> SPACE_USERS_RECEIVED

jcscottiii:
just had one suggestion.

jcscottiii:
I like the commit to separate the two actions now. 

The reason the tests are failing are because only a PR coming from a branch within the 18F repo can decrypt the variables needed for testing. Your PR is coming from a branch on your fork. @msecret

msecret:
@jcscottiii no they won't because I think theres still things they're testing that I haven't completed. But I think this is getting to the point now where I should get any relevant tests passing for each PR. What do you think the best way to do this is? Have the tests run on CI but allow failures? Or I could just run them locally.

jcscottiii:
@msecret move the run of the acceptance test from the travis.yml to the `codecheck.sh` file.
then do
```
set +e && cd acceptance; go test -tags acceptance; cd .. && set -e
```
This should go at the end but before the return at the end.

This *should* (fingers crossed) allow us to still run the acceptance tests but not fail the build

msecret:
@jcscottiii I'm attempting to fix the acceptance tests a bit and ran into a confusing go problem:

So we're importing `github.com/18F/cf-deck/acceptance/views` in `user_login_test`. So when I change something in the `acceptance/views` files it doesn't get actually updated in my local code because it's using the `18F` repo to pull the code down.

Is there some obvious go thing here I'm missing?

jcscottiii:
@msecret what is your $GOPATH?
If that's all set up right, you should be able to modify the Go code even in your fork and it will affect the acceptance tests.

We should pair on this. GOPATH can be confusing.

jcscottiii:
Just two minor comments. after those two modifications, i agree, let's merge

PR #135, opened 2015-10-26
386 fix frontend tests

msecret:
This fixes the front end test that have to run in phantom js on travis CI. This fix has them working locally, not sure about travis ci yet.

For more information about why, look at commit messages.


msecret:
Phantom is still failing, for some reason the karma launcher doesn't work on linux. I can investigate including phantomjs2 somehow and referencing the binary with a env var. Will have to explore this on a linux.

jcscottiii:
FYI: sudo was turned off to use the faster travis container infrastructure.

glad you were able to get it to work without sudo

PR #137, opened 2015-11-04
Fix Org Users Roles bug and provide feedback on err

jcscottiii:
This goes from using the /users API to the new /user_roles API. which will include the roles when getting the users list for the Org. This helps fix the user bug when trying to retrieve org user roles. Instead of doing 3 calls per user (1 for billing_manager, 1 for org_manager, 1 for org auditor), this patch does 1 single call to the `/user_roles` API which gets all the roles for the users in the org. This is a new API introduced by cloud foundry.

Org Users Menu is a similar place as Space Users Menu
image

image

Error when trying to remove org user:
When the user has other org roles.
image

Error directly from API surfaced to the user when trying to remove user when they have roles inside spaces within the org
image

Error for inviting non-exisiting user to org
image


jcscottiii:
@geramirez @msecret take a look

jcscottiii:
Didn't put too much into this angular stuff since it's going to be replaced with the React refactor

static/app/views/partials/org_tabs.html
```diff
@@ -0,0 +1,10 @@
+<div class="page-header">
+    <h3 class="text-center">{{activeOrg | orgNameFilter}} Spaces</h3>
+</div>
+<!-- Nav tabs -->
+<ul class="nav nav-tabs nav-justified">
```
geramirez: Changing this to `<ul class="nav nav-pills nav-justified">`  might help with the visual, the spaces sections are pills.

static/app/views/manage_org.html
```diff
@@ -1,9 +1,6 @@
...
-    <h3 class="text-center">{{activeOrg | orgNameFilter}} Management</h3>
-</div>
-
-<div>
-
+<ng-include src="'app/views/partials/org_tabs.html'"></ng-include>
+<div class='text-center' ng-hide="loadComplete.status">Loading...</div>
+<div ng-show="loadComplete.status">
     <!-- Nav tabs -->
     <ul class="nav nav-tabs nav-justified" role="tablist">
```
geramirez: cross what i said before, maybe the pill should go here.

static/app/controllers.js
```diff
@@ -92,136 +92,110 @@
...
+                    if (response.status !== 201) {
+                        $scope.removeOrgUserError = response.data.description;
+                    } else {
+                        $scope.showTab('current_org_users')
+                    }
                 })
         };
         // Show a specific tab
         $scope.showTab = function(tab) {
+            $scope.removeOrgUserError = null;
```
geramirez: might want to clear `$scope.inviteOrgUserError` here so that the message doesn't persist.

jcscottiii:
Also, this goes from using the `/users` API to the new `/user_roles` API. which will include the roles when getting the users list for the Org

geramirez:
@jcscottiii looks great! that was a fast overhaul! 
**Curses version 195 of the CF API**

msecret:
This looks good to me, especially considering it will be refactored at some point.

mogul:
@jcscottiii Is there any chance we can hook the generation of this message to point people at the invitation link on login.cloud.gov?
user doesn't exist

PR #138, opened 2015-11-05
390 remove user from org

msecret:
Functionality to remove a user from an org, from the org management page which you can get to from the space page (I realize this is different). Also displays error messages from cloud foundry.

### DIFF
--- a/static_src/components/user_list.jsx
+++ b/static_src/components/user_list.jsx
@@ -93,6 +93,7 @@ export default class UserList extends React.Component {
 
 UserList.propTypes = {
   initialUsers: React.PropTypes.array,
+  // Set to a function when there should be a remove button.
   onRemove: React.PropTypes.function
 };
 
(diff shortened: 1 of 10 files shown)


static_src/components/user_list.jsx
```diff
@@ -22,19 +27,59 @@ export default class UserList extends React.Component {
...
+  }
+
   get columns() {
-    return [
+    var columns = [
       { label: 'Name', key: 'username' },
       { label: 'Date Created', key: 'created_at' }
     ];
+
+    if (this.props.onRemove) {
```
jcscottiii: when will this not be set?. there are similar conditionals elsewhere. if you're planning for something in the future where it won't be set. put a TODO.

jcscottiii:
just had one comment. otherwise looks good. @msecret

static_src/components/user_list.jsx
msecret: Ha so currently it's not set for the space user list and it will add a remove button to the list. I added a note to the UserList propTypes on link 94. I think that's a good place to comment as it's kinda defining the type for the component. Let me know what you think.

PR #139, opened 2015-11-12
Remove duplication in action tests

msecret:
An attempt to remove duplication in action tests, open to
opinions.

WIP, wanted feedback

### DIFF
--- a/static_src/test/unit/actions/service_actions.spec.js
+++ b/static_src/test/unit/actions/service_actions.spec.js
@@ -2,7 +2,7 @@
 import '../../global_setup.js';
 
 import AppDispatcher from '../../../dispatcher.js';
-import { assertAction } from '../helpers.js';
+import { assertAction, setupViewSpy, setupServerSpy } from '../helpers.js';
 import cfApi from '../../../util/cf_api.js';
 import serviceActions from '../../../actions/service_actions.js';
 import { serviceActionTypes } from '../../../constants.js';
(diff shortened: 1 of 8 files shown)


static_src/test/unit/actions/service_actions.spec.js
```diff
@@ -6,6 +6,15 @@ import cfApi from '../../../util/cf_api.js';
 import serviceActions from '../../../actions/service_actions.js';
 import { serviceActionTypes } from '../../../constants.js';
 
+function assertAction(spy, type, params) {
+  expect(spy).toHaveBeenCalledOnce();
+  let arg = spy.getCall(0).args[0];
```
jcscottiii: change `arg` to `actualOrgGuidArg` or something that lets the reader know that the arg being assigned should the org.

jcscottiii: my only worry is the indices used for `getCall` and `args` i really doubt the order of the call or args will ever matter. but i'm just thinking how this will be used in case some other developer comes along trying to add tests.

jcscottiii: with that said. i'm fine with it for now, and we can worry about it later when that time comes.

static_src/test/unit/actions/service_actions.spec.js
```diff
@@ -17,17 +18,40 @@ describe('serviceActions', function() {
     sandbox.restore();
   });
 
+  function setupViewSpy() {
+    return sandbox.spy(AppDispatcher, 'handleViewAction');
```
jcscottiii: if you find yourself referring to the string `'handleViewAction'` or `'handleServerAction'` elsewhere in the code, I would make those constants. i'm not sure if you do use them more than here. i remember i did that in the angular stuff and wished i didn't. too many places to find and replace when refactoring. your call on this one.

PR #140, opened 2015-11-12
Refactor stores shared functionality

msecret:
Moves all `get()` and `getAll()` functionality to the base store and moves tests there. Also tests the rest of the functionality of the base store.

This was done to reduce repetition and increase unit test coverage for important and highly used code.


jcscottiii:
really like this patch

PR #143, opened 2015-11-18
477 service plan list

msecret:
List of service plans under each service. I had to use a normal table for it rather the Reactable because it lacks functionality, I plan on solving that more fully when we have a greater sense of design. Also had to do some fixing of the async HTTP code as it was masking errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "karma-sourcemap-loader": "^0.3.5",
     "karma-webpack": "^1.7.0",
     "node-libs-browser": "^0.5.2",
+    "promises-done-polyfill": "^1.0.0",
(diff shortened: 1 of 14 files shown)


msecret:
@jcscottiii `done` will return undefined rather then another promise object. It will also not swallow errors, any error that happens in the then function stack will float back up to be reported in the browser. You should still use `then` when there's further things to be done with the data, because it returns a Promise object. You can call `done` when it's the completed value, ready to be used.

Unfortunately, this information is ahrd to find online because `done` isn't in the actual Promise implementation right now, hence the polyfill.

http://jamesknelson.com/are-es6-promises-swallowing-your-errors/

PR #144, opened 2015-12-08
428 add service instance

msecret:
An add service instance form.

Does a good amount of react form work so it won't need to be done again.

### DIFF
--- a/static_src/actions/service_actions.js
+++ b/static_src/actions/service_actions.js
@@ -45,6 +45,22 @@ export default {
     });
   },
 
+  createInstance(name, spaceGuid, servicePlanGuid) {
+    AppDispatcher.handleViewAction({
+      type: serviceActionTypes.SERVICE_INSTANCE_CREATE,
+      name: name,
+      spaceGuid: spaceGuid,
+      servicePlanGuid: servicePlanGuid
+    });
+  },
+
+  createdInstance(serviceInstance) {
+    AppDispatcher.handleServerAction({
+      type: serviceActionTypes.SERVICE_INSTANCE_CREATED,
+      serviceInstance: serviceInstance
+    });
+  },
+
   receivedInstances(serviceInstances) {
     AppDispatcher.handleServerAction({
       type: serviceActionTypes.SERVICE_INSTANCES_RECEIVED,
--- a/static_src/actions/service_actions.js
+++ b/static_src/actions/service_actions.js
@@ -45,9 +45,9 @@ export default {
     });
   },
 
-  createInstanceDialog(serviceGuid, planGuid) {
+  createInstanceForm(serviceGuid, planGuid) {
     AppDispatcher.handleViewAction({
-      type: serviceActionTypes.SERVICE_INSTANCE_CREATE_DIALOG,
+      type: serviceActionTypes.SERVICE_INSTANCE_CREATE_FORM,
       serviceGuid: serviceGuid,
       servicePlanGuid: planGuid
     });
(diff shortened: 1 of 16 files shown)


static_src/actions/service_actions.js
```diff
@@ -45,6 +45,37 @@ export default {
     });
   },
 
+  createInstanceDialog(serviceGuid, planGuid) {
```
jcscottiii: is the `createInstance_Dialog_` the best place for it in the rest of the service actions?

static_src/components/create_service_instance.jsx
```diff
@@ -0,0 +1,107 @@
...
+  _onValidForm = (values) => {
+    serviceActions.createInstance(
+      values.name,
+      values.space,
+      this.props.servicePlan.guid
+    ); 
+  }
+
+  get serviceName() {
+    return this.props.service.label || 'unknown';
```
jcscottiii: `unknown` -> `Unknown Service Name`

static_src/components/create_service_instance.jsx
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  get servicePlanName() {
+    return this.props.servicePlan.name || 'unknown';
```
jcscottiii: `unknown` -> `Unknown Service Plan Name`

static_src/components/form.jsx
```diff
@@ -0,0 +1,257 @@
+
```
jcscottiii: This is great. I only have one worry, is there an existing react library that creates these kind of form elements already? Just trying to think long term since we would have to possibly add additional `FormElements` in the future

jcscottiii: If not, I think this element stuff should be spun of into a library so that others could use when starting their own React projects.

jcscottiii: This whole thought came from me thinking about how to test this. It's generic enough for others to use it. But the tests for this would stick out from the tests for the rest of the code. (But I'm not sure if it's too opinionated).
Ok I'm done with this rant lol.

static_src/actions/service_actions.js
jcscottiii: If it does belong here, maybe `SERVICE_INSTANCE_CREATE_FORM` instead?

static_src/util/cf_api.js
```diff
@@ -14,6 +14,28 @@ import userActions from '../actions/user_actions.js';
 const APIV = '/v2';
 
 export default {
+  version: APIV,
+
+  fetch(url, action, multiple) {
```
jcscottiii: For a separate PR, we need to take care of pagination of the returned resources.

jcscottiii:
Finished review @msecret.
Any other thoughts @geramirez ?

static_src/actions/service_actions.js
msecret: Yea totally agree, I didn't like dialog.

static_src/components/form.jsx
msecret: Yea I had similar thoughts. I think ideally, we'd be able to use a 3rd-party react forms library that allowed us to have control of the html for the forms. This would allow us to style our forms how we want while, ensure they're accessible, and use the logic of a maintained library. I think it would take a good amount of research to find the best 3rd party library for this.

I propose that the next time we have to create a form for the deck, if the form requires more functionality to be added to what we have here, then we'll commit to researching third party solutions rather then adding to this code. From there, we'll refactor any current form code to use the 3rd party library. If we really can't find a 3rd party library that meets our needs, we'll then commit to moving this form code into it's own 18f library and import it in.

msecret:
Moving on now, @geramirez will review if he has time on the closed PR.

PR #145, opened 2015-12-08
Some bug fixes from refactor work

msecret:
I caused some bugs in my refactoring work. The shared fetch methods
should be able to pass extra params through to the actions because some
require to do so. Fixed more react warnings.


jcscottiii:
These all look good. we need some tests to make sure these bugs don't come back in. but i'm guessing these would be caught by the acceptance tests. (which will be redone at a later point)

PR #146, opened 2015-12-11
425 user permission roles

msecret:
Adds ability to modify (remove and add) user roles from a user. The UI to do this is a simple list of checkboxes below each user in the user list. This is changed from the current app as to not make more UI that may be unnecessary.

More info in commit messages.

### DIFF
--- a/static_src/stores/user_store.js
+++ b/static_src/stores/user_store.js
@@ -23,8 +23,20 @@ class UserStore extends BaseStore {
         cfApi.fetchOrgUsers(action.orgGuid);
         break;
 
+      case userActionTypes.ORG_USER_ROLES_FETCH:
+        cfApi.fetchOrgUserRoles(action.orgGuid);
+        break;
+
       case userActionTypes.SPACE_USERS_FETCH:
         cfApi.fetchSpaceUsers(action.spaceGuid);
+        break;
+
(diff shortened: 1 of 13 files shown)


static_src/test/unit/util/cf_api.spec.js
```diff
@@ -473,13 +473,32 @@ describe('cfApi', function() {
       expect(actual).toMatch(new RegExp(expected));
       expect(actual).toMatch(new RegExp('organizations'));
       expect(actual).toMatch(new RegExp('users'));
-      actual = spy.getCall(0).args[1];
+      actual = spy.getCall(0).args[1]
```
jcscottiii: what happened to the semicolon?

static_src/stores/user_store.js
```diff
@@ -23,10 +30,83 @@ class UserStore extends BaseStore {
...
+        });
+        break;
+
+      case userActionTypes.USER_ROLES_DELETED:
+        var user = this.get(action.userGuid);
+        if (user) {
+          let role = resourceToRole[action.roles] || action.roles;
+          let idx =  user.organization_roles && 
+            user.organization_roles.indexOf(role);
+          if (idx > -1) {
```
jcscottiii: just want to clarify, `indexOf` returns that -1 if not found right?

msecret: yes

PR #154, opened 2015-12-17
Change navbar from dropdown to simpler

msecret:
This allows me to do navigate around in acceptance testing
in a much easier way. It also moves towards to community UI way
of the nav.


static_src/main.js
```diff
@@ -32,7 +32,10 @@ function dashboard() {
 function org(orgGuid) {
   orgActions.changeCurrentOrg(orgGuid);
   cfApi.fetchOrg(orgGuid);
-  React.render(<App><SpaceList initialOrgGuid={ orgGuid } /></App>, mainEl);
+  React.render(
+    <App currentOrgGuid={ orgGuid }>
```
jcscottiii: Why do we need to store the org guid in the app tag? and why not also the space or app guid

static_src/main.js
```diff
@@ -45,7 +48,7 @@ function space(orgGuid, spaceGuid, potentialPage) {
     cfApi.fetchOrgUserRoles(orgGuid);
   }
   React.render(
-    <App>
+    <App currentOrgGuid={ orgGuid }>
```
jcscottiii: same

static_src/main.js
```diff
@@ -60,7 +63,7 @@ function marketplace(orgGuid, serviceGuid, servicePlanGuid) {
     serviceActions.createInstanceForm(serviceGuid, servicePlanGuid);
   }
   React.render(
-    <App>
+    <App currentOrgGuid={ orgGuid }>
```
jcscottiii: same

msecret: Yea, for this current navbar, it doesn't matter at all, I think I'll remove

PR #157, opened 2015-12-22
Using path.resolve for node_modules path

msecret:


msecret:
There will be an error message in the build I can find that will determine if it worked or not.

PR #168, opened 2016-03-22
Added user management tests

msecret:
This ensures we catch regressions of changing user permissions.

### DIFF
--- a/acceptance/user_managment_test.go
+++ b/acceptance/user_managment_test.go
@@ -101,6 +101,38 @@ var _ = Describe("UserManagement", func() {
 		})
 	})
 
+	It("should allow an org manager to change a space users pemissions", func() {
+		var userRow *agouti.Selection
+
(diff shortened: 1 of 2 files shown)


acceptance/user_managment_test.go
```diff
@@ -101,6 +101,38 @@ var _ = Describe("UserManagement", func() {
 		})
 	})
 
+	It("should allow an org manager to change a space users pemissions", func() {
+		var userRow *agouti.Selection
+
+		By("allowing the user to navigate to the space users page", func() {
+			Expect(page.Navigate(fmt.Sprintf(testEnvVars.Hostname+
```
jcscottiii: Mixing of concatenation with string specifiers
`fmt.Sprintf("%s/#/org/%s/spaces/%s/users", testEnvVars.Hostname, testOrg, testSpace)`

msecret: Good catch

PR #171, opened 2016-04-06
Feature incorporate cg style

msecret:
Work to get cloudgov-style working with webpack. Work imports only base cloudgov tag styles, component styles will be brought directly into react.

TODO: make this independent of using USWDS or not.
TODO: get cloudgov-style components working (requires setting up CSS modules again)

See commit messages for details.


msecret:
Closed for local branch (rather then remote) https://github.com/18F/cg-deck/pull/175

PR #176, opened 2016-04-06
Add airbnb eslint

msecret:
Ensures that it checks airbnbs linting rules againts es6 and
JSX.

Is testing for the guild:
https://github.com/18F/frontend/issues/114#issuecomment-206037302


jeremiak:
:+1: works great

PR #241, opened 2016-04-13
Incorporate cg style

msecret:
Refs #243


package.json
```diff
@@ -11,6 +11,7 @@
     "build": "npm run clean && node_modules/webpack/bin/webpack.js",
     "codecov": "cat ./coverage/coverage-final.json | ./node_modules/codecov.io/bin/codecov.io.js",
     "clean": "rm -rf ./static/assets/*",
+    "install-git": "npm install git+https://git@github.com/18F/cg-style.git#master && cd node_modules/cloudgov-style && npm install && npm run build; cd ../../",
```
jeremiak: @msecret are you wedded to calling this script "install-git"? It is a bit confusing, I think something like "cloudgov-style" would be a bit more obvious what this script is doing.

In addition, do we want to add this command into the docs or do we want this to run in something like `scripts:postinstall` so other devs don't need to remember to run this command?

package.json
```diff
@@ -63,18 +70,14 @@
     "babel-runtime": "^6.3.19",
     "bootstrap": "^3.3.6",
     "classnames": "^2.1.5",
+    "cloudgov-style": "git+https://git@github.com/18F/cg-style.git#master",
```
jeremiak: It makes me a bit nervous to pull straight off of `#master` since that can change without us being aware and break the build if the project changes significantly. This happened to me on USWDS.

Could we start doing releases on cloudgov-style or is that too much overhead?

msecret: I think we could. I've been pretty good about using semantic versioning on `cg-style`, and we'll end up using it to the extent where we'll need a good process to do things like this.

This webpack is a big change, which changes a lot of cg-style, so I'm doing things in a slower manner. My goal with cg-style is: you see a need for a change in cg-deck, you make the change in cg-style, you npm link to ensure the change worked, you test cg-style to make sure nothing broke and then you push up a new version. So it'll be quick and cg-deck will mostly use the new versions of cg-style.

That's my ideal, and we can see how well that pans out.

But for this PR, which I plan in updating a minor, cg-style to 0.5, it's a little slower.

package.json
msecret: I'm fine calling it whatever. And we should add to docs. Maybe it's a command we should keep around long-term for future times we need it like this?

jeremiak: Cool, I added f91e568 then

Issue #172, opened 2016-04-06
Remove bootstrap CSS and classes from all code


msecret:
In order for developers to not have to deal with unused CSS code AND to improve performance of the app, bootstrap CSS should be removed from the app and dependencies, AND there should be no bootstrap classes in the components.

Acceptance criteria
- [ ] Bootstrap should not be in the bundled CSS output file
- [ ] Bootstrap should not be a dependency in `npm`
- [ ] None of the components should have bootstrap classes in their `className` attribute.

jeremiak:
Hey @msecret I think this should've been closed with 242

msecret:
I put it into awaiting acceptance.

jeremiak:
Oh right, I forgot that PRs can get merged and then their original cards go to awaiting acceptance. Good call

PR #242, opened 2016-04-14
Remove bootstrap from the project

jeremiak:
This removes bootstrap from the code base. It explicitly removes it from the `package.json` as well as `npm-shrinkwrap.json` and any references to bootstrap have been removed from the code and build steps.

Acceptance criteria from #172:

- [X] Bootstrap should not be in the bundled CSS output file
- [X] Bootstrap should not be a dependency in `npm`
- [X] None of the components should have bootstrap classes in their `className` attribute.

### DIFF
--- a/static_src/components/app_list.jsx
+++ b/static_src/components/app_list.jsx
@@ -69,13 +69,13 @@ export default class AppList extends React.Component {
     var content = <h4 className="test-none_message">No apps</h4>;
     // TODO format rows in table
     if (this.state.apps.length) {
-      content = <Table data={ this.getRows(this.state.apps) } 
+      content = <Table data={ this.getRows(this.state.apps) }
         columns={ this.columns }
-        sortable={ true } className="table" />;
+        sortable={ true } />;
     }
 
     return (
-      <div className="tableWrapper"> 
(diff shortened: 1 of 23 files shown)


static_src/components/app_list.jsx
```diff
@@ -69,13 +69,13 @@ export default class AppList extends React.Component {
     var content = <h4 className="test-none_message">No apps</h4>;
     // TODO format rows in table
     if (this.state.apps.length) {
-      content = <Table data={ this.getRows(this.state.apps) } 
+      content = <Table data={ this.getRows(this.state.apps) }
         columns={ this.columns }
-        sortable={ true } className="table" />;
+        sortable={ true } />;
```
msecret: I unfortunately *may* have used ".table" in the acceptance tests. They also seem to be failing on travis, although that may be some unrelated error (I tried re-running them). Did they run successfully locally?

msecret: Yea it looks like a bunch of the tests are failing.

jeremiak: I'll try adding the table classes and see if it passes

PR #245, opened 2016-04-15
Added more from main cg products

msecret:
Rather then redefining things here, I wanted to just link to the official docs and augment with a few more items here.


jeremiak:
Looks good to me

PR #247, opened 2016-04-18
Ignore acceptance tests if a PR

jcscottiii:


jeremiak:
Thanks @jcscottiii! I'm not going to merge this because if an acceptance test is failing we should try to fix it for a short period of time and then disable that individual test per [@msecret's new changes](https://github.com/18F/cg-deck/commit/ac4409434b7673ab75194b38c160afeeb7998070) to the CONTRIBUTING guide.

PR #248, opened 2016-04-19
Update orgs for deploy and tests

jcscottiii:
For deployment:
space deck-prod was removed.
space deck-stage became space deck-prod
space deck-staging-alpha became space deck-prod

For testing
The org cf-deck became cf-deck-testing 
This means we need to update the test


jeremiak:
Sweet, thanks @jcscottiii! This is what was causing my tests to error out

Issue #173, opened 2016-04-06
Webpack import of components into react files


msecret:
In order developers to be able to use their own CSS for the deck AND for cloudgov users to have a shared visual experience when using the deck, the deck should incorporate the cg-style CSS components into it's own components

Acceptance criteria
- [ ] Any existing cg-style components that are in use on the deck should be importable.
- [ ] Components CSS should be imported by using CSS modules, minimizing global classes.
- [ ] Remove all sass processing from webpack config.

msecret:
The style side of this task has been worked on here: https://github.com/18F/cg-style/pull/49

PR #249, opened 2016-04-20
Use webpack to load css modules from cg-style

jeremiak:
To load CSS modules from webpack, it needs to be configured with the proper loaders and the proper paths. Because cg-style is still under active development, we don’t install it from the npm registry.

`npm link` (by itself) isn’t sufficient as its just a symlink. When webpack goes to try and resolve things, it will use the absolute path so we need to include that in the loader.

That path has been abstracted away into an environment variable called `CG_STYLE_PATH`, which should be an absolute path to the location of the cg-style repo.

The README has been updated to also reflect the new environment variable.

Refs #173

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,13 +3,15 @@ const path = require('path');
 
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 
+const CG_STYLE_PATH = process.env.CG_STYLE_PATH;
+
 const srcDir = './static_src';
 const compiledDir = './static/assets';
 
 module.exports = {
   entry: [
     'babel-polyfill',
-    srcDir + '/main.js'
+    `${srcDir}/main.js`
   ],
 
   output: {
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -30,20 +32,15 @@ module.exports = {
...
         }
       },
-      { test: /\.css$/,
-        include: path.resolve(__dirname, 'static_src/css'),
+      {
+        test: /\.css$/,
+        include: [
+          path.resolve(__dirname, 'static_src/css'),
+          path.resolve(__dirname, 'node_modules/cloudgov-style'),
+          CG_STYLE_PATH
```
msecret: If the `CG_STYLE_PATH` env var isn't set, will anything break?

webpack.config.js
```diff
@@ -3,13 +3,15 @@ const path = require('path');
...
+const CG_STYLE_PATH = process.env.CG_STYLE_PATH;
+
 const srcDir = './static_src';
 const compiledDir = './static/assets';
 
 module.exports = {
   entry: [
     'babel-polyfill',
-    srcDir + '/main.js'
+    `${srcDir}/main.js`
```
msecret: awesome

webpack.config.js
jeremiak: Yes, everything will break. That's because the JSX components require the styles and if they can't resolve those file names the webpack process will fail

msecret: oh I mean also in the use case where we have a pinned cloudgov-style rather the the symlinked folder

jeremiak: Gotcha. No, I don't think so, because the proper files will be resolved in `node_modules/cloudgov-style`. Just so that its not undefined though I should probably put in a default string or something.

PR #252, opened 2016-04-26
Ms cgstyle svg icons

msecret:
An react component implementation of icons. User must set the name of the icon, which maps directly to the filename and then can set modifier classes for different colors and such.

The main point of this PR is to iron out how icons would work so we can continue with a cg-style 0.4 release. Please think about any considerations for the future that we could change now. Like: does the naming scheme make sense?

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,8 +61,6 @@
     "sass-loader": "^3.2.0",
     "scss-loader": "0.0.1",
     "style-loader": "^0.13.0",
-    "svgo": "^0.6.6",
-    "svgo-loader": "^1.1.0",
     "url-loader": "^0.5.7",
     "watch": "^0.16.0",
     "webpack": "^1.12.9"
--- a/package.json
+++ b/package.json
@@ -61,6 +61,8 @@
     "sass-loader": "^3.2.0",
     "scss-loader": "0.0.1",
     "style-loader": "^0.13.0",
+    "svgo": "^0.6.6",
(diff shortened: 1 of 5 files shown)


msecret:
Another change included in this that isn't relevant to cg-style: I changed the alias of cloudgov-style to itself, without the `css` directory. This is because we now want to import from other directories, such as `img`. I think this makes overall sense.

msecret:
The issue has something to do with not being able to install wget from apt, reminding myself to look into this:

```
Step 2 : RUN apt-get install -y wget
 ---> Running in 2fd1ed7296d6
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package wget
The command '/bin/sh -c apt-get install -y wget' returned a non-zero code: 100
```

msecret:
Seems like it's in apt: https://apps.ubuntu.com/cat/applications/wget/

acceptance/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM ubuntu:14.04
 MAINTAINER James C. Scott III <jcscott.iii@gmail.com>
 
+RUN apt-get -qq update
```
msecret: @jcscottiii this fixed the build problems I was having in the main convo, but would like your feedback on it.

jeremiak:
This looks good to me, I like using the more explicit asset path from `cloudgov-style`.

Issue #192, opened 2016-04-08
Two-level nested navigation


mogul:
In order to present orgs and space compactly and enable easy switching, we want to provide collapsing two-level navigation of the available orgs and spaces in the console nav.

Acceptance Criteria
- [ ] Orgs and spaces are alphabetized within their scope
- [ ] First level of nav is org, second is space
- [ ] Orgs have a synthetic "overview" entry as the first in their list of spaces, which brings up the org overview
- [ ] Responsive should collapse the menu, similar to docs.cloud.gov

PR #258, opened 2016-05-03
Use cloudgov-style sidenav for organizations and spaces

jeremiak:
To address #192, this uses the `cloudgov-style` assets to style the sidenav. Includes modifying the Navbar React component to use the correct HTML structure and data, updating the API utility module to also get all space information, and toggle the display of menus in the sidenav. All newly added code should have appropriate testing.

This is the first component that has some specific UI related state so we didn't have a pattern for handling that. After some discussion, @msecret and I decided to have UI actions that change state create actions to be dispatched and handled by the stores so that components can be as stateless as possible. To accommodate this, I added a new type of action creator that dispatches `UI_ACTION` sourced actions. This pattern was utilized to toggle the displays of submenus in the navigation. The component creates an `ORG_TOGGLE_SPACE_MENU` type action that the OrgStore registers to handle. The OrgStore adds a `space_menu_open` key to each organization and toggles it when told to.

Keeping this data in the store enables us to keep the components as stateless as possible. There was some talk about nesting all UI related state elements under a key like `ui` in each object, but we decided against it for the time being.

### DIFF
--- a/static_src/app.jsx
+++ b/static_src/app.jsx
@@ -58,8 +58,11 @@ export default class App extends React.Component {
       </nav>
       <div className={ titleBarStyles['title_bar'] }>
         <div className="nav_toggle">
-          <i className="fa fa-bars nav_toggle-icon"></i>
-          <div className="icon-reorder tooltips" data-original-title="Toggle Navigation" data-placement="bottom">
+          <i className="nav_toggle-icon"></i>
+          <div className="icon-reorder tooltips"
+            data-original-title="Toggle Navigation"
+            data-placement="bottom"
+          >
           </div>
         </div>
         <h1 className={ titleBarStyles['title_bar-title'] }>Organizations</h1>
--- a/static_src/app.jsx
+++ b/static_src/app.jsx
@@ -2,8 +2,9 @@
 import classNames from 'classnames';
 import React from 'react';
 
-// import styles from './css/main.css';
-import sidenavStyles from 'cloudgov-style/components/sidenav.css';
+import cgBaseStyles from 'cloudgov-style/css/base.css';
+import sidenavStyles from 'cloudgov-style/css/components/sidenav.css';
+import titleBarStyles from 'cloudgov-style/css/components/title_bar.css';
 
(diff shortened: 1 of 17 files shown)


static_src/app.jsx
```diff
@@ -23,6 +25,10 @@ export default class App extends React.Component {
     LoginStore.addChangeListener(this._onChange);
   }
 
+  componentWillUnmount() {
+    LoginStore.removeChangeListener(this._onChange);
```
msecret: 👍

static_src/app.jsx
```diff
@@ -53,22 +56,29 @@ export default class App extends React.Component {
...
-      <div className="container-fluid">
-        <div className="row">
-          <nav className={ sidebarClasses }>
-            { sidebar }
-          </nav>
-          <main className={ mainClasses }>
-            { content }
-          </main>
+      <div className={ titleBarStyles['title_bar'] }>
+        <div class="nav_toggle">
```
msecret: Aren't these `class`es not actually being applied in React? Can they be removed?

static_src/app.jsx
```diff
@@ -53,22 +56,29 @@ export default class App extends React.Component {
...
+          <i class="fa fa-bars nav_toggle-icon"></i>
+          <div class="icon-reorder tooltips" data-original-title="Toggle Navigation" data-placement="bottom">
```
msecret: Are those data attributes valid JSX code?

static_src/app.jsx
jeremiak: I think we want them so I'll change the attribute to `className` as React expects

static_src/app.jsx
jeremiak: Ya, apparently data attributes are really the only valid HTML and JSX attributes.

[Relevant JSX documentation](https://facebook.github.io/react/docs/jsx-gotchas.html#custom-html-attributes)

msecret:
Lgtm after that one last class thing

jeremiak:
@msecret You're right, I removed the FA classes

Issue #187, opened 2016-04-07
Add cloud.gov top navigation


msecret:
In order for users to navigate to the other cloud.gov product pages while in the deck, there should be a global cloud.gov top nav in the deck.

Acceptance criteria
- [x] There should be a topnav that has a similar look and feel to docs and landing sites.
- [x] All the links in the top nav should go to their respective places
- [x] The main link in the menu should stay on the deck page.

msecret:
One potential difficulty here: landing and deck have a max width and deck does not (nor should it in many ways). Do we (a) modify the navigation to work in this full screen site or do we (b) modify the deck header to by the same size as the other sites?

Considerations:
- (b) means that switching between sites would include a visual "jump" for users because the layout of the header, navigation and links would be different.
- (a) might not look like a cohesive experience
- (a) what happens when we add search? Global cloud.gov search probably doesn't make sense on the deck (deck specific search would only work)
- (b) would modifying the layout require extra work to make each break point work?

jeremiak:
I think this should come before #237, otherwise we'll do some integration work and then have to replicate some amount of it to work with the search. Also, sounds like we're not 100% sure how the search would fit in, though I think that should go in #237.

Here are how some other sites handle it:

* Facebook has a max width, and therefore has space on each side on a wide screen. They make their left hand nav blend in by keeping it the same bg color as the page
* AWS stretches the main content to fill the space and has their left side element stick to the left side of the window
* Heroku takes the same approach as Facebook, with the left nav detaching from the edge of the window but being the same color

I personally like the Facebook/Heroku approach of having a max-width and making the background of the elements fill the available space.

@mogul where are those design artifacts/comparison of the other Cloud Foundry implementors? I couldn't find them in `cg-style` or `cg-product`

msecret:
I'd agree with the Facebook/Heroku approach. This will work well for cloud.gov as its a simple white background in the header.

mogul:
@jeremiak they're [in Google Drive](https://drive.google.com/drive/folders/0BwLqM4Nicmq-bUt0NjRjclFMUEU)... Sorry for the delayed response!

PR #268, opened 2016-05-11
[WIP] Adding the global cloud.gov header

jeremiak:
This is a work in progress PR that isn't yet ready to be merged. It should implement the cloud.gov navigation per #187, using style from 18f/cg-style


jeremiak:
Whoops, this should've been against `staging-alpha` not `master`

PR #269, opened 2016-05-11
Add the global cloud.gov header

### LINKED ISSUE #187: Add cloud.gov top navigation

jeremiak:
This PR should implement the cloud.gov navigation using style from 18f/cg-style and close #187.

I accomplished this by adding a new `<Disclaimer />` component and a `<Header />` component, which import the proper styles from [cg-style](/18f/cg-style).

I'm also trying out a new pattern for using CSS modules with the `createStyler` function which is in 6400b48 and I would like feedback on it. I like it because it seems to make the components a bit more readable, and more CSS like while also removing a bunch of ugly variable declaration statements at the top of `render()` ([such as in `navbar.js`](https://github.com/18F/cg-deck/blob/staging-alpha/static_src/components/navbar.jsx#L69-L92)). 

The color the navigation links is not correct, but will be once 18F/cg-style#70 is merged and a new version of the library is published.


static_src/util/create_styler.js
```diff
@@ -0,0 +1,17 @@
+import classNames from 'classnames';
+
+function uniqueOnly(value, index, self) {
+  return self.indexOf(value) === index;
+}
+
+export default function createStyler(...args) {
```
msecret: What happens if there are two styles from separate files that are the same? Or is that something that would've even happen because of CSS Modules?

static_src/components/header.jsx
```diff
@@ -0,0 +1,60 @@
...
+import logoStyle from 'cloudgov-style/css/components/logo.css';
+
+export default class Header extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.styler = createStyler(headerStyle, navStyle, logoStyle);
+  }
+
+  getImagePath(iconName) {
```
msecret: Is this something that would make sense to have globally?

static_src/components/disclaimer.jsx
```diff
@@ -0,0 +1,40 @@
...
+import createStyler from '../util/create_styler';
+
+export default class Disclaimer extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.styler = createStyler(baseStyle);
+  }
+
+  getImagePath(iconName) {
```
msecret: Is this being used in the current file?

static_src/components/header.jsx
jeremiak: I'm not sure I understand what you mean by globally. Do you mean in `<App/>` (or something similar) instead of as its own component?

static_src/util/create_styler.js
jeremiak: Ya, I don't think that would happen. I think the hashing on the end of CSS module classes is there to prevent that

static_src/components/header.jsx
msecret: Yea I can't think of a good place for it, never mind. I just meant that it seems like a thing many components will use.

msecret:
This looks awesome, I think it's merge ready after my comment: https://github.com/18F/cg-deck/pull/269#discussion_r63803822

static_src/components/header.jsx
jeremiak: Possibly. I think some components will include similar, if not the same, CSS files but that shouldn't be a problem with CSS modules.

I tried to get the `createStyler` function to be most of the re-usable parts but we can always refactor again later if see a common pattern emerge

static_src/components/disclaimer.jsx
jeremiak: Hmm, nope! not sure how that wasn't picked up by the linter but will remove

jeremiak:
@msecret Awesome, will merge when the build turns green now that `getImagePath` is removed

PR #277, opened 2016-05-11
Change policy on acceptance tests based on retro

msecret:
Due to acceptance tests failures related to separate tests interfering with each other and tests not working because they don't have a clean data state, the acceptance tests should be stated as not a focus right now.


jeremiak:
I made a tiny modification to be more explicit but looks good, I’ll merge now

PR #282, opened 2016-05-17
Change promise code to remove non-standard `done`

msecret:
Removes all instances of the `done()` method on promises favoring the use of `then()` and `catch()`. Changes the bulk of test code in `cf_api` tests to use a resolved or rejected promise, meaning the assertions have to be done in the `then`.

### DIFF
--- a/static_src/stores/user_store.js
+++ b/static_src/stores/user_store.js
@@ -4,16 +4,15 @@
  * server.
  */
 
-import AppDispatcher from '../dispatcher';
 import BaseStore from './base_store.js';
 import cfApi from '../util/cf_api.js';
 import userActions from '../actions/user_actions.js';
 import { userActionTypes } from '../constants.js';
 
 const resourceToRole = {
-  'managers': 'org_manager',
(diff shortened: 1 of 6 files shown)


static_src/stores/user_store.js
```diff
@@ -25,133 +24,152 @@ class UserStore extends BaseStore {
   }
 
   _registerToActions(action) {
-    switch(action.type) {
-      case userActionTypes.ORG_USERS_FETCH:
+    switch (action.type) {
+      case userActionTypes.ORG_USERS_FETCH: {
         cfApi.fetchOrgUsers(action.orgGuid);
         break;
+      }
```
jeremiak: @msecret did the linter complain about not having these curly brackets?

msecret: No it told me to put them there actually.

jeremiak:
This looks pretty good to me, nice work.

I think we should remove `promises-done-polyfill` from `package.json` and regenerate `npm-shrinkwrap.json`.

jeremiak:
🚢

Issue #272, opened 2016-05-11
Format time and date in the deck


msecret:
In order for deck users to clearly understand the time or date of an item, dates and times in the deck should be formatted in `hours:minutes`

Acceptance criteria:
- [ ] All datetime stamps should be formatted in the same way in the app
- [ ] They should be formatted like: `03/21/2016 10:39am PDT`

msecret:
@jeremiak is there anything more specific we can say here?

PR #284, opened 2016-05-19
Format datetime stamp utility function

jeremiak:
Provides a utility function to format date strings just like "03/21/2016 10:39am PDT". The function takes two arguments:

0. A datetime string such as the ones returned from the API (i.e. "2015-02-25T02:11:46Z")
0. An optional timezone (that defaults to UTC) so that we can potentially allow users to toggle between timezones in the future

Unit tests are added to ensure that the function guards against bad input and to ensure that the returned string is properly formatted.

Also includes the addition of `moment-timezone` to `package-.json` and adjustments to `webpack.config.js` so that the timezone data is loaded properly.

Use `formatDateTime()` in `<ServiceInstanceList/>`
Use `formatDateTime()` in `<UserList/>`

This refs #272 but does not close it because I am only using this new function in two components. To close the issue, the function needs to be used across the application. I figure that once this utility function is available, we can change the other components as we touch them.


webpack.config.js
```diff
@@ -48,6 +48,10 @@ module.exports = {
       },
       { test: /\.(ttf|woff2?|eot)$/,
         loader: 'url-loader?limit=1024&name=font/[name].[ext]'
+      },
+      {
+        test: /\.json$/,
```
msecret: Where is this needed?

msecret:
LG although tests are failing. I honestly don't get why they're failing, but it could be the app isn't running correctly on them for some reason. I'm trying to restart them.

webpack.config.js
jeremiak: This is needed because `moment-timezone` loads the timezone data through a `.json` file

jeremiak:
I think the problem is [that `moment-timezone` isn't included](https://travis-ci.org/18F/cg-deck/builds/131292752#L2409), I'll try and fix that

jeremiak:
Alright, I needed to update `npm-shrinkwrap.json` and now the build is green. I also added `npm run shrinkwrap` so that we can use it to consistently include `devDependencies`

Issue #220, opened 2016-04-08
App page shows basic information about the app: name, state, status, last push, build pack, stack and disk limit.


mogul:
In order for a space manager to see each apps name and state about their app there should be an app page with basic information on the app.

Acceptance criteria
- [ ] GIVEN a space manager SM AND the space manager has space SA AND SA has an app A WHEN  SM visits the SA page THEN SM should see a link to the A page with the name of A.
- [ ] GIVEN a space manager SM AND the space manager has space SA AND SA has an app A WHEN SM clicks on the A link on the SA page THEN SM should see an app page for A with
  - [x] name of app
  - [x] app state (running, paused, etc)
  - [x] date of last push
  - [x] which build pack was used (can this be multiple?)
  - [ ] stack limit (?)
  - [ ] disk limit

msecret:
Requires a design.

msecret:
I actually think we can ignore the design on this story for now, waiting until a designer gets ramped up on the project. So the point of this story will be more to create an app page, link to it and show some basic information.

msecret:
@jcscottiii I'm a little unsure of how to get the current memory/disk and the memory/disk limit that's currently on the app page. Do you happen to know where that is in the API?

jcscottiii:
There are two APIs.
This API will give you the limits for an app regardless of what state it is in: https://apidocs.cloudfoundry.org/237/apps/get_app_summary.html

When the app is in started mode, you can the current levels: https://apidocs.cloudfoundry.org/237/apps/get_detailed_stats_for_a_started_app.html

Issue #289, opened 2016-05-19
Add view of events and logs in app page


msecret:
In order for developers to see information of what is happening in their app, there should be a UI element that shows data for events and logs from logs.cloud.gov

Acceptance criteria
MVP:
- [ ] There's links in the events and log sections of the app page which take the developer to the correct page on the logs.cloud.gov site.

Post-MVP:
- [ ] The relevant information from logs.cloud.gov is shown for events and logs on the app page (likely through the use of an iframe)

msecret:
@jeremiak assigning you as this would be a good thing to look into next. I think it might be best to investigate Post-MVP first to see if it's feasible.

jeremiak:
What is the "correct page" defined as? It sounds to me like clicking on the link would take you to the "Dashboard" view on logs.cloud.gov and all of the events would be filtered by that application's name. How far off the mark am I?

jeremiak:
Also where should the link be exactly? This answer might be dependent on what is decided in #290

msecret:
@jeremiak it should be close to where it is on the current console app page, below all the other information.

msecret:
So I'm not great at using Kibana, but it would be the discover page filtered to the org, space and app, probably 24 hours. If you play around in Kibana, it will give you a URL for the pages you pull up.

@mogul do you know how to differ between events and logs in kibana? I'm having trouble figuring it out.

jeremiak:
Cool, some preliminary investigating:

It looks like Kibana gives us some nice URLs to get around. For instance, to filter by app name we can just use 
```
[...]

Issue #256, opened 2016-04-28
Use immutable data library in stores


msecret:
In order for developers to lose less time due to mutating state AND for an org manager to see deck views that render and perform faster the flux stores should use immutable.js data structures.

msecret:
Basic idea of how this would work:

The `_data` array on each store would be an `Immutable.List`.

When changes to the data are made from store actions, `_data` would be updated through Immutable's `mergeDeep` or `updateIn` methods.

jeremiak:
Acceptance criteria:
- [ ] Attempting to change a store's data from within a component is not allowed

PR #290, opened 2016-05-20
Ms app page

msecret:
Adds an app page with basic information as defined in #220. Uses the app's summary API route for app name, state, last push, buildpack. Uses app's stat API route for current mem/disk vs limit.

Picked the simplest UI possible, by showing this as a horizontal table. This was done to limit having to redo work when we learn more from design.


static_src/app.jsx
```diff
@@ -81,7 +81,7 @@ export default class App extends React.Component {
   }
 }
 App.propTypes = {
-  children: React.PropTypes.element,
+  children: React.PropTypes.any,
```
msecret: This was getting rid of an annoying error because I think we pass different things as children, sometimes an arry of elements, sometimes maybe just an element?

static_src/components/app_page.jsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  formatMb(bytes) {
+    return Math.round(bytes / 1000000);
+  }
+
+  render() {
+    let content = <h4 className="test-none_message">No app</h4>;
+
+    if (this.state.app.name) {
```
msecret: If name isn't there I'm assuming everything else is, might need to be more selective though.

static_src/stores/app_store.js
```diff
@@ -18,23 +16,42 @@ class AppStore extends BaseStore {
...
         this.emitChange();
         break;
 
+      case appActionTypes.APP_STATS_RECEIVED: {
+        const idx = this._data.findIndex((app) => app.guid === action.appGuid);
+        if (idx < 0) {
+          let newApp = {
+            guid: action.appGuid
+          };
+          newApp = Object.assign({}, newApp, action.app);
```
msecret: Is this the correct way to do this? In this case, I'm essentially creating a new app object and putting the stats data in.

static_src/stores/app_store.js
```diff
@@ -18,23 +16,42 @@ class AppStore extends BaseStore {
...
+          this._data.push(newApp);
+        } else {
+          this._data[idx] = Object.assign({}, this._data[idx], action.app);
```
msecret: Is this the correct way to do this?

msecret:
This is what this looks like:

screen shot 2016-05-24 at 4 49 15 pm

@mogul one thing to point out: I'm not using the UI design on the current app page of those little boxes with different information. This is because we'd have to code it anew, we don't have a full screen page like the current deck does anymore and we don't know if it'd be a successful UI as we don't have research behind it.

If you feel that we should layout this page in the same manner as the current deck app page so that our current user's will understand where things are, let us know and we can do that.

static_src/stores/app_store.js
jeremiak: I don't think this is the preferred way once #256 gets merged. It would be better to use the `.push()` or `.merge()` methods

static_src/components/app_page.jsx
```diff
@@ -0,0 +1,94 @@
...
+    super(props);
+    this.props = props;
+    this.state = {
+      app: {},
+      currentAppGuid: this.props.initialAppGuid
+    };
+    this._onChange = this._onChange.bind(this);
+  }
+
+  componentDidMount() {
```
jeremiak: Make sure to clean up any listeners on `componentWillUnmount()`

static_src/components/app_page.jsx
jeremiak: Won't all apps have a name? I think this is a fine logic check as written

static_src/components/app_list.jsx
```diff
@@ -43,9 +44,9 @@ export default class AppList extends React.Component {
   }
 
   appUrl(app) {
-    return `/#/org/${ this.state.currentOrgGuid }
-      /spaces/${ this.state.currentSpaceGuid }
-      /apps/${ app.guid }`;
+    return dedent`/#/org/${ this.state.currentOrgGuid }
```
jeremiak: Nice! I know the bug that this was causing but definitely didn't guess that it was related to multi-line template strings

static_src/stores/app_store.js
msecret: I guess since those don't exist yet, I should just do this?

static_src/components/app_page.jsx
msecret: Yea they will

jeremiak:
Hey @msecret this looks pretty great, nice work.

I think it should be merged after #256 and the `AppStore` internals should be updated to use the new methods like `.push()` and `.merge()`. Other than that I think it works well for now while we're doing research on what data users need to see here.

Also way to kick that weird app-URL-has-funky-spaces bug with `dedent`!

msecret:
That sounds fine, we can work on getting #256 merged

mogul:
I could opine, but I think I'd rather defer to @vz3 and @juliaelman on this... 

>  I'm not using the UI design on the current app page of those little boxes with different information. This is because we'd have to code it anew, we don't have a full screen page like the current deck does anymore and we don't know if it'd be a successful UI as we don't have research behind it.
> 
> If you feel that we should layout this page in the same manner as the current deck app page so that our current user's will understand where things are, let us know and we can do that.

Victor and Julia, this is a read-only interim-rewrite of the top of the existing app page, which looks like this...

image

It's happening because we're trying to do the minimum work of reimplementation necessary to move off of the old code base to the new one. Do you think this is acceptable? If not, are there simple changes you'd bless us doing to make the step down acceptable until we have user-research driving the redesign of this page?

msecret:
@thisisdano take a look at Bret's last comment too.

juliaelman:
I am going to defer to @thisisdano on this, but think that it might be good for us to start thinking about better styling for tables for cloud.gov overall. It's something we're working on for [Compliance Viewer](https://github.com/18F/compliance-viewer) too.

thisisdano:
That new content section/table is ugly af, but a few fixes would improve it a bunch. 
- Lack of padding in the main content container makes everything worse — 12-16px of padding is a necessary start.
- "About" subhed feels redundant since there are no other options/tabs.
- Table itself has too much padding and has unnecessary right/left borders. Leave the borders on the top and bottom of each row, remove the right/left borders on each cell, reduce the padding by 50% on the top/bottom of each cell, and remove the padding from the left of each cell.
- `Memory` and `Disk Limit` are a little difficult to parse as is. They appear to be paired values `[n]/[m]`, but the relationship isn't clear. Is it saying the app is consuming [n] out of [N] total available? Maybe reformat display to read `<b>[n]</b> out of [N] ([p]%)`
- Is `Disk Limit` the same as (what seems to be) `Disk` usage in the `Live` section of the earlier screen? Should the title of the row be `Disk Usage`?
- Together, those changes would immediately improve the legibility of the content and could be an acceptable interim solution

msecret:
I talked with @thisisdano on slack and we decided that we shouldn't focus on wider design issues here and that a table is fine UI for now.

jeremiak:
Looks good to me, nice work on the code. Merging now for #295 and for my upcoming work in #289

PR #291, opened 2016-05-24
Ensure sidenav dropdowns all appear as links

msecret:
I decided to implement this with HTML rather then using the pointer
property in CSS because this problem is specific to the deck.


jeremiak:
Looks great, nice quick fix!

PR #292, opened 2016-05-25
Stores use immutable data structures internally

jeremiak:
This PR should change all stores so that the underlying data they maintain is kept in immutable data structures using [Immutable.JS](facebook.github.io/immutable-js). This was done in a way that only stores need to be concerned with keeping their data structure in immutable types and that any components that get data from those stores are still working with plain JavaScript objects. I've been casually calling this the "immutable barrier". The stores are implemented so that anything returned from a method that begins with `get` such as `getAll()` will convert the immutable data back to regular JS objects before returning them.

I also changed the API of the BaseStore a bit to make dealing with immutable data a bit easier. For example, I added a [`.push()`](https://github.com/18F/cg-deck/blob/0e57e66b12a5719bad4385ec7509e8de7d99384d/static_src/stores/base_store.js#L31-L37) method which takes an argument, converts it to the respective immutable data type and then creates a new list with that item pushed on and sets that array to be `_data` on the store. It also emits a change event as the underlying data has changed.

One of the benefits of using immutability is that we can easily compare nested objects for sameness. This means that we can quickly/cheaply do checks to see if data has changed or not and act accordingly. I added [`dataHasChanged()`](https://github.com/18F/cg-deck/blob/0e57e66b12a5719bad4385ec7509e8de7d99384d/static_src/stores/base_store.js#L53-L56) to the BaseStore which returns a boolean. This gives every store the ability to only emit change events when the data has actually changed. It also provides an interface for components to be able to use the fast equality checking of immutable data types without having to know about them.

I also added two new merge functions that account for the fact that the data is immutable. A nice benefit is that these merges no longer have unexpected side effects. There are [two new merge APIs](https://github.com/18F/cg-deck/blob/0e57e66b12a5719bad4385ec7509e8de7d99384d/static_src/stores/base_store.js#L106-L141) that have very similar function signatures:

* `merge(mergeKey, dataToMerge, optionalCallback)`
* `mergeMany(mergeKey, arrayOfDataToMerge, optionalCallback)`

Each takes three arguments, with the last being an optional callback. The first argument is the key by which to identify similar objects and merge them (generally speaking this is `guid` in our project), the data to merge in, and a callback that is passed `true` or `false` depending on if there was any data that changed. `merge` takes a single object as the data to merge and `mergeMany` takes an array of objects. By default, the callback will call `this.emitChange()` if there is a data change.

Open questions (probably for @msecret):

0. Many of our APIs in this project are promised-based instead of callbacks. For example, all the methods in `cfApi` return promises instead of using callbacks. I wrote the initial version of these merge functions to use callbacks instead of returning a promise, but it might be best to be consistent and use promises throughout. What do you think?
0. Can we remove [`_merge()`](https://github.com/18F/cg-deck/blob/0e57e66b12a5719bad4385ec7509e8de7d99384d/static_src/stores/base_store.js#L77-L90) now that we have these new methods? It doesn't look like its being used at all anymore.

### DIFF
--- a/static_src/stores/base_store.js
+++ b/static_src/stores/base_store.js
@@ -8,11 +8,15 @@ import Immutable from 'immutable';
 
 import AppDispatcher from '../dispatcher.js';
 
+function defaultChangedCallback(changed) {
+  if (changed) this.emitChange();
+}
+
 export default class BaseStore extends EventEmitter {
 
   constructor() {
     super();
-    this._data = Immutable.List();
+    this._data = new Immutable.List();
   }
 
   subscribe(actionSubscribe) {
@@ -29,7 +33,7 @@ export default class BaseStore extends EventEmitter {
   }
 
   push(val) {
-    let newData = Immutable.fromJS(val);
+    const newData = Immutable.fromJS(val);
     this._data = this._data.push(newData);
     this.emitChange();
 
@@ -38,12 +42,13 @@ export default class BaseStore extends EventEmitter {
 
   get(guid) {
     if (guid && !this.isEmpty()) {
-      let item =  this._data.find((space) => {
-        return space.get('guid') === guid;
-      });
+      const item = this._data.find((space) =>
+        space.get('guid') === guid
+      );
 
       if (item) return item.toJS();
     }
+    return undefined;
   }
 
   getAll() {
@@ -51,14 +56,12 @@ export default class BaseStore extends EventEmitter {
   }
 
   dataHasChanged(toCompare) {
-    let c = Immutable.fromJS(toCompare);
+    const c = Immutable.fromJS(toCompare);
     return !Immutable.is(this._data, c);
   }
 
   formatSplitResponse(resources) {
-    return resources.map((resource) => {
-      return Object.assign(resource.entity, resource.metadata);
-    });
+    return resources.map((r) => Object.assign(r.entity, r.metadata));
   }
 
   emitChange() {
@@ -75,17 +78,20 @@ export default class BaseStore extends EventEmitter {
 
   // TODO determine if we can remove this function
   _merge(currents, updates) {
-    if (currents.length) {
-      updates.forEach(function(update) {
-        var same = currents.find(function(current) {
-          return current.guid === update.guid;
-        });
+    if (!currents.length) return updates;
 
-        same ? Object.assign(same, update) : currents.push(update);
-      });
-    } else {
-      currents = updates;
-    }
+    updates.forEach((update) => {
+      const same = currents.find((current) =>
+        current.guid === update.guid
+      );
+
+      if (same) {
+        Object.assign(same, update);
+      } else {
+        currents.push(update);
+      }
+    });
+
     return currents;
   }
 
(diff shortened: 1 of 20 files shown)


msecret:
There's a linting error on travis, it might be causing the build to fail

msecret:
Answers to questions

1. I think callbacks are fine and we can switch to promises if it's required. I think callbacks still have their place in JS, and this is a great simple use of them.
2. we should remove `_merge`, I hate it.

jeremiak:
Cool, sounds good re: callbacks.

I went ahead and fixed linting errors (and made sure to remove all the store files from `.eslintignore` and removed the old `_merge()` and its companion tests.

I'm still getting an error on a test for the `cfApi` file so I'll take a look at that next.

static_src/stores/base_store.js
```diff
@@ -75,17 +78,20 @@ export default class BaseStore extends EventEmitter {
 
   // TODO determine if we can remove this function
   _merge(currents, updates) {
```
msecret: When can we remove this?

jeremiak: I think that's outdated as of ea77ba70d9d2836c521f2f69f56279842465e95d

static_src/stores/user_store.js
```diff
@@ -155,13 +166,15 @@ class UserStore extends BaseStore {
       }
 
       case userActionTypes.USER_DELETED: {
-        const deleted = this.get(action.userGuid);
-        if (deleted) {
-          const index = this._data.indexOf(deleted);
-          this._data.splice(index, 1);
-          this._error = null;
+        const index = this._data.findIndex((d) =>
```
msecret: I could see a `delete` pattern being potentially useful on the base store.

static_src/stores/service_store.js
```diff
@@ -13,28 +14,34 @@ class ServiceStore extends BaseStore {
...
+      case serviceActionTypes.SERVICES_RECEIVED: {
+        const services = this.formatSplitResponse(action.services);
+        const immutableServices = Immutable.fromJS(services);
 
-      case serviceActionTypes.SERVICES_RECEIVED:
-        var services = this.formatSplitResponse(action.services);
-        this._data = services;
-        this.emitChange();
+        if (this.dataHasChanged(immutableServices)) {
+          this._data = immutableServices;
```
msecret: I could see this pattern being a potential for a base store method, `replaceAll` or something.

static_src/stores/user_store.js
jeremiak: Totally. Would it be `.delete(guid)`?

msecret:
I wrote some notes on where I see some repeated patterns that could potentially be moved to the base store, but I see no need for them to be done now. So I would say this is ready to merge and leave it to you if you want move them.

static_src/stores/user_store.js
msecret: Yea sure

jeremiak: Great, I'll implement that. I think that it makes sense as just `delete` because `get` uses `guid` as the index so it will basically be the same convention

jeremiak:
Great comments, totally agree. I went ahead and added `.delete(guid)` to the `BaseStore` but held off on something like `replaceAll` till we see it utilized more as a common pattern.

@msecret want to give it another once over or should  I just merge on green?

PR #293, opened 2016-05-25
Add slack channel build notifications

msecret:
These are set to just the #cloud-gov-liberator-news channel. I wanted to keep most notifications on.


jeremiak:
Merging, this looks like one of those things we really need to merge to test.

PR #295, opened 2016-05-27
Ms routes app page

msecret:
Built on top of #290, that should be closed first.

This adds a table below the basic info on the app page for routes. To do this, a RouteStore has been created with all the things that go along with that (actions, constants).

Currently work in progress.

### DIFF
--- a/static_src/actions/domain_actions.js
+++ b/static_src/actions/domain_actions.js
@@ -1,0 +1,24 @@
+
+/*
+ * Actions for domain entities. Any actions such as fetching, creating,
+ * etc should go here.
+ */
+
+import AppDispatcher from '../dispatcher.js';
+import { domainActionTypes } from '../constants';
+
+export default {
+  fetch(domainGuid) {
+    AppDispatcher.handleViewAction({
+      type: domainActionTypes.DOMAIN_FETCH,
+      domainGuid
+    });
+  },
+
+  receivedDomain(domain) {
+    AppDispatcher.handleServerAction({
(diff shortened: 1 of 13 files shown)


static_src/actions/domain_actions.js
```diff
@@ -0,0 +1,24 @@
...
+    });
+  },
+
+  receivedDomain(domain) {
+    AppDispatcher.handleServerAction({
+      type: domainActionTypes.DOMAIN_RECEIVED,
+      domain
+    });
+  }
+};
```
jeremiak: @msecret what do you think about consolidating these routeActions and domainActions into routeActions?

Here's my thinking: its kind of some CF specific terminology to know the difference between a route and a domain since most of the time you'll actually want both to determine the URL for an application. By having them handled in the same store, but using different action creators and action types it makes it a bit more confusing to hold the difference between a route and a domain in your head.

I think it would make sense to bring all of this into the `route*.js` files. Thoughts?

msecret: Yea I see what you mean. I think I separated them because there is a just Routes and just Domains view in the community UI, so figured we'd need to separate them conceptually at some point.

jeremiak: Ya, we can wait for the research team to tell us otherwise, but I have a feeling that the division is hard for most users to understand and that we'll actually be using them in concert.

I'm open either way, which direction do you want to go?

msecret: Ha the one with less work.

Issue #179, opened 2016-04-06
Pre-launch: Add logout button to UI


mogul:
In order for a user to change which account they're logged into, there should be a logout button in the UI.

jeremiak:
So what if instead of "Deck" we put "Login" (or "Logout") in the nav bar on the deck project? This would:

0. Make the nav link more useful since its pretty clear nobody knows what "Deck" means
0. Put the link on the far right side of the navigation
0. Provide an obvious place to put a "Logout" link

msecret:
Its funny, @vz3 was just talking about this. @vz3, @thisisdano, @juliaelman what do y'all think?

thisisdano:
Sounds simple and logical.

jeremiak:
I did a really simple take in #296, would appreciate your opinions @vz3 @thisisdano @juliaelman and @msecret

PR #296, opened 2016-05-27
Use login/logout link the global navigation

jeremiak:
Quick sketch of what using a header link for login/logout might look like in the global nav. I substituted the previous "Deck" link with a link at the far right of the navigation because user testing revealed that the "Deck" link was of very little value to users.

The link will be "Login" when the user is not yet authenticated and "Logout" when the user is properly authenticated.

Screenshots:

Logged out:

Logged in:

Refs #179

### DIFF
--- a/static_src/components/header.jsx
+++ b/static_src/components/header.jsx
@@ -1,6 +1,8 @@
 
 import React from 'react';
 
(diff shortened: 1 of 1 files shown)


mogul:
:heart:

PR #298, opened 2016-05-31
Add links for Logs and Events

jeremiak:
This PR adds a link for the current logs view of the application. There is no way to embed an iframe until we resolve 18F/logsearch-for-cloudfoundry#5, so in the meantime we are using plain links to direct users to the proper places on logs.cloud.gov. Screenshot:

Right now, the "Logs" link works but I'm having issues with the "Events" link. I'm not quite sure what it should link to. When I follow @mogul's instructions I get an empty dashboard. Screenshot: 
The instructions I am referencing are:

0. Load logs.cloud.gov
0. Select "Load Saved Dashboard"
0. Select "App - Events"
0. No results found

Refs #289

### DIFF
--- a/static_src/components/app_page.jsx
+++ b/static_src/components/app_page.jsx
@@ -29,13 +29,20 @@ export default class AppPage extends React.Component {
     });
   }
 
+  logsLink(appName) {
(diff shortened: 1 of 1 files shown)


msecret:
This looks good, waiting for you to check what Bret said. I also re-ran the tests because it was a random failure.

jeremiak:
Here's a screenshot of the current view:
Screenshot of current implementation of the app page component with event and log links

This is technically ready for review @msecret. Both the logs and events links should go to the proper view on Kibana, with the time bounds set by default to year to date to help ensure that there will be data for the user to look at.

I wanted to raise the issue of the current text not being super helpful to users who don't already have context on what the difference between an "event" and the regular "logs" are. Indeed, I'm not quite sure of this difference myself. @mogul anywhere I can grab some text on what an event is and how it differs from the other logs?

mogul:
@jeremiak said:

> I wanted to raise the issue of the current text not being super helpful to users who don't already have context on what the difference between an "event" and the regular "logs" are. Indeed, I'm not quite sure of this difference myself. @mogul anywhere I can grab some text on what an event is and how it differs from the other logs?

I don't see any good summary in the CF docs, so maybe distill [what you see in this SO answer](http://stackoverflow.com/a/24911041)?

Issue #266, opened 2016-05-10
Add analytics to the deck


msecret:

jeremiak:
@msecret can you be a bit more specific about which actions are "major" actions? It would be helpful to list the actions in [`constants.js`](/18F/cg-deck/blob/staging-alpha/static_src/constants.js) that should be instrumented with analytics

msecret:
@jeremiak it will be all actions that are significant for user research. Here's a list:

- SERVICE_INSTANCE_DELETE
- SERVICE_INSTANCE_DELETED
- SERVICE_INSTANCE_CREATE_FORM
- SERVICE_INSTANCE_ERROR
- SERVICE_INSTANCE_CREATE
- SERVICE_INSTANCE_CREATED
- USER_ROLES_ADD
- USER_ROLES_ADDED
- USER_ROLES_DELETE
- USER_ROLES_DELETED
- USER_DELETE
- USER_DELETED
- ERROR_REMOVE_USER

So it's really every action that isn't standard fetch and receive.

jeremiak:
What data should be sent along with the analytics? As this is for feature prioritization based on usage, do we just want to know that the event occured?

msecret:
I think we could send along extra data if it makes sense, but may want to be careful of user privacy. Like maybe which Role is added/deleted.

msecret:
Checked GA and found events and page views.

PR #299, opened 2016-05-31
Track page views and app actions with Google Analytics

jeremiak:
This PR adds functionality to track page views and app actions using Google Analytics. There are two new helper functions in `/static_src/util/analytics.js`:

0. `trackAction()` - takes an `action` object
0. `trackPageView()` - takes a `url` string

The dispatcher (63ed517) is the centralized place for tracking actions though Google Analytics calls them "events", and the router (f5f8a23) tracks page views.

A few open questions for @msecret and @mogul:

0. Are there any actions we should blacklist? That is to say, are there any actions on which we do _not_ track?
0. What is the proper GA property id? [Currently](https://github.com/18F/cg-deck/blob/974d80219a57c5931e125223a7684062cc27c874/static/index.html#L15) I am using the top level cloud.gov ID but we might want to make a new property for the deck. 

Refs #266


mogul:
> 0. Are there any actions we should blacklist? That is to say, are there any actions on which we do _not_ track?

Not that I know of... Let's try capturing everything first, and see how noisy our data is and decide what to omit later.

> 0. What is the proper GA property id? [Currently](https://github.com/18F/cg-deck/blob/974d80219a57c5931e125223a7684062cc27c874/static/index.html#L15) I am using the top level cloud.gov ID but we might want to make a new property for the deck. 

I believe that's the right one to use... I think we only ever had one for all of cloud.gov, but @msecret may know better.

msecret:
In the original story #266, I said we should ignore actions involved in fetching, and instead ensure we track the whole page view. I don't think it's important that they are ignored, but I do think it would be good to do synthetic page view tracking (because this is a single page app, going to different pages might not trigger a new page view so there's a GA api to track it manually)

msecret:
@jeremiak I was under the impression we should use the one on cg-style, which is the one you have.

static_src/util/analytics.js
```diff
@@ -0,0 +1,19 @@
+export function trackAction(action) {
+  if (!window.ga) return;
+
+  window.ga('send', {
+    hitType: 'event',
+    eventCategory: action.source,
+    eventAction: action.type,
+    eventLabel: action.type
```
msecret: Do we need this if it's repeated? Can't `label` be empty?

jeremiak: Yup, totally can. Will remove and update

jeremiak:
@msecret as far as I understand it, f5f8a23d79bfd2ef7135456217b760107a988cc1 should enable synthetic page view tracking by sending a new page view event on every request that matches a route in the router

jeremiak:
This is now ready for code review @msecret

Issue #294, opened 2016-05-26
Add links to documentation for unimplemented actions on app page


msecret:
In order for deck users to know how to accomplish actions that were in a previous iteration of the deck, there should be links to documentation on how to complete actions with the CLI.

Acceptance criteria:
- [x] There should be a link on how to start/stop/reset the app, https://docs.cloudfoundry.org/devguide/deploy-apps/deploy-app.html (this is the best page I could find, although the instructions don't match up prefectly)
- [x] There should be a link to create/delete a route, https://docs.cloud.gov/apps/custom-domains/#application-routes
- [x] There should be a link on how to (un)bind a service, https://docs.cloud.gov/apps/managed-services/

msecret:
- Bind/Unbind service: https://docs.cloud.gov/apps/managed-services/#bind-the-service-instance

msecret:
Here's an example of what this text will look like:

To modify, create or delete a route, view [application routes](https://docs.cloud.gov/apps/custom-domains/#application-routes) for more information.

msecret:
Because there is no Services view currently on the home page, we'll create a heading for it with just the link and link text under it.

jeremiak:
Looks good as a temporary solution

jeremiak:
This is complete and looks good. Nice work!

I realize that these links are a temporary measure and don't always match up perfectly (stopping/starting isn't really clear in that documentation) but we can always change it as we get better documentation and interactions.

PR #300, opened 2016-05-31
Ms docs links app page

msecret:
See #294

### DIFF
--- a/static_src/components/app_page.jsx
+++ b/static_src/components/app_page.jsx
@@ -82,10 +82,24 @@ export default class AppPage extends React.Component {
                 </tr>
               </tbody>
             </table>
+            <aside>
+              <p>To start or stop an app, view the <a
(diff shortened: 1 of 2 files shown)


static_src/components/app_page.jsx
```diff
@@ -82,10 +82,24 @@ export default class AppPage extends React.Component {
                 </tr>
               </tbody>
             </table>
+            <aside>
+              <p>To start or stop an app, view the <a
+                href="https://docs.cloudfoundry.org/devguide/deploy-apps/deploy-app.html"
+                target="_blank">
+                <span>&nbsp;</span>deployment guide</a> for more information.
```
jeremiak: Why are you using `&nbsp;` here?

msecret: It's to ensure spacing in React. If you move to a new line, if won't always but a space.

jeremiak:
Looks good to me!

Issue #306, opened 2016-06-03
Make a bunch of small fixes


jeremiak:
In getting the `staging-alpha` refactor ready for a merge to `master`, we generated a series of small issues that are expected to take less than five minutes. This issue is to track those at a higher level to make sure we get everything done.

- [x] Set proper view port on deck
- [x] Remove the page title bar (with the dark background color) from the secondary header
- [x] Set `max-width` of 1400px
- [x] Include a new column in the user table for “permissions”
- [x] Style delete action buttons appropriately (red)
- [x] Apps view table doesn’t close at the bottom
- [x] Add units to app list view/table for “memory” and “disk quota” (fixed by #330)
- [x] Add link to documentation for adding users to organization on the user management view (https://docs.cloud.gov/apps/managing-teammates/)
- [x] Add link to role definitions on documentation sites (http://docs.cloudfoundry.org/concepts/roles.html#roles)

jeremiak:
I am closing/accepting this even though I created it because @juliaelman and @msecret did all the work

PR #313, opened 2016-06-03
Ms small design fixes

msecret:
Tracks #306. See individual commit messages for more information.

### DIFF
--- a/static_src/components/users.jsx
+++ b/static_src/components/users.jsx
@@ -146,6 +146,11 @@ export default class Users extends React.Component {
         initialItem={ this.state.currentTab } />
         { errorMessage }
         <div>
+          <p>For more information on different roles and what they mean see <a
+            href="http://docs.cloudfoundry.org/concepts/roles.html#roles"
+            target="_blank">
+            <span>&nbsp;</span>Cloud Foundry's roles information</a>.
+          </p>
           <p>To add a user to an organization see <a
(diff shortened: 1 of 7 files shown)


msecret:
The acceptance tests are failing in travis due to a change in login configuration at a different level of cloud.gov. These tests will be turned off in a separate PR.

static_src/components/users.jsx
```diff
@@ -159,6 +159,11 @@ export default class Users extends React.Component {
                 onRemovePermissions={ this.handleRemovePermissions }
             />
           </div>
+          <p>For more information on different roles and what they mean see <a
```
jeremiak: @msecret what do you think about putting this at the top of the page for now? I think its worthwhile context and its hard to find if there are a decent number of users in an org.

jeremiak:
["Remove user from org"](https://github.com/18F/cg-deck/blob/staging-alpha/static_src/components/user_list.jsx#L69) action  should by styled red since it is a delete action

jeremiak:
The ["title bar"](https://github.com/18F/cg-deck/blob/ms-small_design_fixes/static_src/app.jsx#L66-L74) needs to be removed to satisfy "Remove the page title bar (with the dark background color) from the secondary header"

jeremiak:
Looks great to me, nice work.

PR #315, opened 2016-06-07
Updates to getting started instructions

jeremiak:
@juliaelman and I were pairing on getting her local environment set up and the instructions were a bit out of date.

Mostly, this added some instructions on cloning the repository so that it fits with a typical Go project structure.

I also changed references from "Angular" to "React" so that all the information is accurate for the rewrite branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,13 +13,17 @@ This ["deck"](#whats-a-deck) is a web application to manage cloud.gov organizati
 Learn more about [cloud.gov](https://cloud.gov).
 
 ## Tech Stack
-- `Go` (v1.6.2 required) for the backend server. [![Go Code Coverage Status](https://coveralls.io/repos/18F/cg-deck/badge.svg?branch=master&service=github)](https://coveralls.io/github/18F/cg-deck?branch=master)
 
-- `Node` (v.4.2.3) and `React` for the frontend. [![JS Code Coverage Status](https://coveralls.io/repos/18F/cg-deck/badge.svg?branch=staging-alpha&service=github)](https://coveralls.io/repos/18F/cg-deck/badge.svg?branch=staging-alpha&service=github)
+### Backend Server [![Go Code Coverage Status](https://coveralls.io/repos/18F/cg-deck/badge.svg?branch=master&service=github)](https://coveralls.io/github/18F/cg-deck?branch=master)
+- `Go` (version 1.6.2)
+
+### Front end application [![JS Code Coverage Status](https://coveralls.io/repos/18F/cg-deck/badge.svg?branch=staging-alpha&service=github)](https://coveralls.io/repos/18F/cg-deck/badge.svg?branch=staging-alpha&service=github)
+- `Node` (version 4.2.3)
+- `React` (version ^0.14.0)
 
 ## Setup
 ### Cloning the repository
-If you are unfamiliar with `Go` project directory structure, you want the code in this repository to be in something like `<your-code-directory>/cg-deck-ws/src/github.com/18f/cg-deck`. You can use that exact pattern by cloning the repository with:
+If you are unfamiliar with [`Go` project directory structure](https://golang.org/doc/code.html#Workspaces), you want the code in this repository to be in something like `<your-code-directory>/cg-deck-ws/src/github.com/18f/cg-deck`. You can use that exact pattern by cloning the repository with:
 
 ```
 git clone git@github.com:18F/cg-deck.git cg-deck-ws/src/github.com/18F/cg-deck
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # 18F Cloud Foundry Deck
 
-[![Build Status](https://travis-ci.org/18F/cg-deck.svg?branch=master)](https://travis-ci.org/18F/cg-deck)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -15,10 +15,16 @@ Learn more about [cloud.gov](https://cloud.gov).
 ## Tech Stack
 - `Go` (v1.6.2 required) for the backend server. [![Go Code Coverage Status](https://coveralls.io/repos/18F/cg-deck/badge.svg?branch=master&service=github)](https://coveralls.io/github/18F/cg-deck?branch=master)
 
-- `AngularJS` for the frontend. [![JS Code Coverage Status](http://codecov.io/github/18F/cg-deck/coverage.svg?branch=master)](http://codecov.io/github/18F/cg-deck?branch=master)
-
+- `Node` (v.4.2.3) and `React` for the frontend. [![JS Code Coverage Status](https://coveralls.io/repos/18F/cg-deck/badge.svg?branch=staging-alpha&service=github)](https://coveralls.io/repos/18F/cg-deck/badge.svg?branch=staging-alpha&service=github)
```
juliaelman: I'd recommend putting React on it's own line and reference the version number. Probably something important to understand when debugging and also to reference during installation.

README.md
```diff
@@ -15,10 +15,16 @@ Learn more about [cloud.gov](https://cloud.gov).
...
 
 ## Setup
+### Cloning the repository
+If you are unfamiliar with `Go` project directory structure, you want the code in this repository to be in something like `<your-code-directory>/cg-deck-ws/src/github.com/18f/cg-deck`. You can use that exact pattern by cloning the repository with:
```
juliaelman: Maybe make `project directory structure` hyperlink to: https://golang.org/doc/code.html#Workspaces

README.md
```diff
@@ -15,10 +15,16 @@ Learn more about [cloud.gov](https://cloud.gov).
...
+
+```
+git clone git@github.com:18F/cg-deck.git cg-deck-ws/src/github.com/18F/cg-deck
+```
```
juliaelman: Since (I think) `ws` stands for `workspace`, we might want to mention this and why the clone adds that in. Not 100% sure it HAS to be explicit, but might be helpful for folks not familiar with the Go file structure requirements.

README.md
```diff
@@ -15,10 +15,16 @@ Learn more about [cloud.gov](https://cloud.gov).
...
+
 ### Create a Client with UAAC
```
juliaelman: @jeremiak mentioned to me that this may or may not be necessary for local dev. If so, should we state either why it is helpful and if it is a requirement or not?

README.md
```diff
@@ -34,28 +40,24 @@ uaac client add <your-client-id> \
 - Unable to create an account still? Troubleshoot [here](https://docs.cloudfoundry.org/adminguide/uaa-user-management.html#creating-admin-users)
 
 ### Set the environment variables
-If you are testing locally, export these variables. If you are deploying to cloud foundry, modify the manifest.yml
+If you are testing locally, export these variables. If you are deploying to cloud foundry, modify the `manifest.yml`
```
juliaelman: Should we add advice on best practices for exporting the environment variables?

juliaelman: Or maybe we could add an example file for folks to cp and set?

README.md
```diff
@@ -34,28 +40,24 @@ uaac client add <your-client-id> \
...
 - `CONSOLE_API_URL`: The URL of the API service. i.e. `http://api.domain.com`
 - `CONSOLE_LOG_URL`: The URL of the loggregator service. i.e. `http://loggregator.domain.com`
-- `PPROF_ENABLED`: An optional variable. If set to `true` or `1`, will turn on `/debug/pprof` endpoints as seen [here](https://golang.org/pkg/net/http/pprof/)
-- `CG_STYLE_PATH`: The absolute path to your `cg-style` repo. Used for webpack in development
+- `PPROF_ENABLED`: <optional> If set to `true` or `1`, will turn on `/debug/pprof` endpoints as seen [here](https://golang.org/pkg/net/http/pprof/)
+- `CG_STYLE_PATH`: <optional> The absolute path to your `cg-style` repo. If set, will use a local copy of `cloudgov-style` to build the front end application.
 
 ## Front end
-Install front end dependencies, including the `cloudgov-style` library
+Install front end dependencies:
```
juliaelman: Add note about where in the dir structure to run this cmd from. e.g. `/path/to/folder/cg-deck-ws/src/github.com/18F/cg-deck`

juliaelman:
Looks good and thank you for taking the time to do this! Added a few notes that might help clarify things a bit further. I was able to get up and running with only a few road bumps (which I was able to figure out!). ^5

README.md
jeremiak: Great idea

README.md
jeremiak: I think this is still necessary, I just gave you my generated client tokens when we were getting your environment setup. In fact, if you could test out these instructions to make sure they work that would be really valuable.

jeremiak:
Thanks for the feedback @juliaelman, I think I addressed most of your points. Great idea about a sample file of environment variables. Ready for your review!

juliaelman:
LGTM - thanks @jeremiak!

PR #318, opened 2016-06-08
Updated template for new tab styles

thisisdano:
**What:** Updated class calls in `navbar.jsx` to allow styling from `cg-styles`.

**Why:** The nav template needed styling added through React to correctly read the new tab styles (see https://github.com/18F/cg-style/pull/81) from the stylesheet.

### DIFF
--- a/static_src/components/tabnav.jsx
+++ b/static_src/components/tabnav.jsx
@@ -2,12 +2,16 @@
 import React from 'react';
 
 import classNames from 'classnames';
+import navStyles from 'cloudgov-style/css/components/nav.css';
+import createStyler from '../util/create_styler';
+
 
 export default class Tabnav extends React.Component {
   constructor(props) {
     super(props);
     this.props = props;
     this.state = { current: this.props.initialItem };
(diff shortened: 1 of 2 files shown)


static_src/components/tabnav.jsx
```diff
@@ -2,27 +2,31 @@
...
+    this.styler = createStyler(navStyles);
   }
 
   componentWillReceiveProps(nextProps) {
     this.setState({current: nextProps.initialItem});
   }
 
   render() {
-    var defaultClasses = ['nav', 'nav-tabs', 'nav-justified'];
+    var defaultClasses = [this.styler('nav'), this.styler('nav-tabs'), this.styler('nav-justified')];
```
jeremiak: Nice work. It isn't well documented but you can actually pass in any number of classes to the `this.styler` function at a time. So this could be written as:
`this.styler('nav', 'nav-tabs', 'nav-justified')`

msecret: Jeremia, I told him this way because we hadn't merged my styler changes.

Issue #304, opened 2016-06-03
Adjust layout of checkboxes on forms


msecret:
In order for a user to correctly associate a user permission with a checkbox, the label and checkbox element must visually be together.

Current state:
![checkboxes-current_state](https://cloud.githubusercontent.com/assets/1701077/15789488/516910f2-2981-11e6-85e9-bcc902965260.png)

msecret:
Acceptance criteria:
- [ ] The user role (Manager, auditor, etc) should be on the same line and visual connected to the checkbox that controls it.
- [ ] The group of checkboxes should be in the same row as the user, so it's visually connected to the user.

msecret:
Screenshot of getting 2nd acceptance criteria: 
![screen shot 2016-06-02 at 3 44 34 pm](https://cloud.githubusercontent.com/assets/1701077/15789412/f4fb4772-2980-11e6-99dc-43fd1023a1a9.png)

From here, we should additionally move the label so it's next to the check box rather than the next line.

PR #328, opened 2016-06-08
Je adjust checkbox layout

juliaelman:
Refs #304 and relies on https://github.com/18F/cg-style/pull/82

See screenshot below for layout reference //cc @thisisdano 

screen shot 2016-06-08 at 1 08 21 pm


jeremiak:
This looks good!

It requires a release in `cg-style` to take effect but I'm going to merge it now because it doesn't break the interface and we don't want long running PRs.

PR #330, opened 2016-06-08
Separating out apps data into rows/columns + adding unit def.

juliaelman:
Refs #306.

This PR completes the list of items defined in these small design fixes.

screen shot 2016-06-08 at 4 08 29 pm

### DIFF
--- a/static_src/components/app_list.jsx
+++ b/static_src/components/app_list.jsx
@@ -69,11 +69,37 @@ export default class AppList extends React.Component {
 
   render() {
     var content = <h4 className="test-none_message">No apps</h4>;
-    // TODO format rows in table
     if (this.state.apps.length) {
-      content = <Table data={ this.getRows(this.state.apps) }
-        columns={ this.columns }
-        sortable={ true } />;
+      content = (
+        <table sortable={ true }>
+          <thead>
+          { this.columns.map((column) => {
+            return (
+              <th column={ column.label } className={ column.key }
+                  key={ column.key }>
+                { column.label }
+              </th>
+            )
+          })}
+          </thead>
+          <tbody>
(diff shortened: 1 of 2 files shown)


static_src/components/app_list.jsx
```diff
@@ -69,11 +69,37 @@ export default class AppList extends React.Component {
...
+                <td label="Memory">{ app.memory } MB</td>
+                <td label="Instances">{ app.instances }</td>
+                <td label="State">{ app.state }</td>
+                <td label="Disk quota">{ app.disk_quota } MB</td>
+              </tr>
+            ])
+          })}
+          </tbody>
+        </table>
+      );
```
jeremiak: Nice, I find this to be much more readable than using `Reactable`.

jeremiak:
This looks great, nice work.

Minor quibble: On the app page (`/#/orgs/:guid/spaces/:guid/apps/:guid`) there isn't any space between the number and "MB" so we should be consistent. I think this would also prevent the text from breaking when the memory quantity is four digits. I think being consistent is more important than preventing line breaks so I think either approach is fine

Screenshot of app page:
Screenshot of app page

Screenshot of line break in app list table:
Screenshot of line break in app list table

static_src/components/app_list.jsx
msecret: Yea we gotta get rid of Reactable at some point, I think it sucks.

msecret:
Speaking of @jeremiak 's comment, I like the way on the app list page more, having a space between the number and MD.

jeremiak:
This looks great. I went ahead and formatted the app page to match the app list page. I'll merge when this is green

PR #344, opened 2016-06-15
Ms service delete confirmation

msecret:
Added confirmation UI to service instance delete. Uses actions for the confirm coming up, canceling and confirming the delete which map to a boolean on each service instance to track whether the confirmation is currently showing. This boolean is used to show the confirmation component and disable the delete button.

Other changes:
- Removed Reactable on service actions because it sucks and can't take style attributes.

### DIFF
--- a/static_src/actions/service_actions.js
+++ b/static_src/actions/service_actions.js
@@ -90,6 +90,13 @@ export default {
     });
   },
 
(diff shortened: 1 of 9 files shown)


jeremiak:
Looks good! I'm going to merge this so we can show it for the demo today and so that we can test that our deploy is working again. We'll come back and make sure to address the comments above.

PR #351, opened 2016-06-15
Ms create instance cancel

msecret:
Finishes https://github.com/18F/cg-style/issues/78 by adding the actions and UI to allow for cancelling out of the form.


thisisdano:
Works great for me!

Issue #347, opened 2016-06-15
Marketplace page sometimes doesn't load after visiting other pages


msecret:
I found while using the console-staging deck, when I was clicking around various pages and returned to the Marketplace page, it would say "No services found".

jeremiak:
This looks resolved by #354.

Acceptance criteria:
- [X] A user can navigate to the Marketplace, and then to a space page, and back to the Marketplace page successfully.
- [X] The second marketplace visit should load quickly with cached data from the initial load, before making a request for new data and re-rendering if needed.

PR #354, opened 2016-06-15
Ms react errors

msecret:
Fixes various problems to the JX code and react such as ensuring components start with existing data, HTML in jsx is valid and unmounting components. Some of these were causing actual bugs, such as #347 .


jeremiak:
Looks great, nice work! I had seen that initial state bug before but didn't know what was causing it.

PR #355, opened 2016-06-16
Attempt to get new relic license key in

msecret:
At first I added the license key directly with cf, but found that
it was missing today, likely after the deploys. I'm taking a guess
that the env vars are completely replaced by a deploy process, therefore
need to be set as part of the build? @jcscottiii am I on the right track?


jcscottiii:
@msecret yeah that's right! we need to document that process of adding new env vars :/

PR #357, opened 2016-06-16
Deploy master branch to prod space

### LINKED ISSUE #356: Master branch deploy is broken
Currently, Travis does not deploy the `master` branch to the production route (console.cloud.gov). This is because the `master` branch is set to deploy to our current staging space (`deck-stage`).

I think this is the offending bit of code in [`/deploy/travis_deploy.sh`](https://github.com/18F/cg-deck/blob/master/deploy/travis_deploy.sh#L12-L21)
```
if [ "$TRAVIS_BRANCH" == "master" ]
then
	CF_MANIFEST="manifests/manifest-master.yml"
	CF_SPACE="deck-stage"
elif [ "$TRAVIS_BRANCH" == "production" ]
then
	CF_MANIFEST="manifests/manifest-production.yml"
	CF_SPACE="deck-prod"
fi
```

We no longer have a `production` branch so we should adjust this script and remove the `manifest-production.yml` since that is also out of date.

jeremiak:
Instructs Travis to deploy `master` branch to production space and `staging-alpha` branch to the staging space. I also removed the `manifest-production.yml` and addd `manifest-staging.yml`.
 
To fix #356

### DIFF
--- a/deploy/travis_deploy.sh
+++ b/deploy/travis_deploy.sh
@@ -13,11 +13,11 @@ CF_PATH="."
 if [ "$TRAVIS_BRANCH" == "master" ]
 then
 	CF_MANIFEST="manifests/manifest-master.yml"
-	CF_SPACE="deck-stage"
-elif [ "$TRAVIS_BRANCH" == "production" ]
-then
-	CF_MANIFEST="manifests/manifest-production.yml"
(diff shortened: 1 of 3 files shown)


juliaelman:
LGTM

Issue #308, opened 2016-06-03
There should be no unauthenticated view of the deck


jeremiak:
A user should not see the deck if they are not authenticated. Instead, they should be redirected to the UAA login page, and redirected to the deck after a successful authentication.

Currently, there is an extra click just to get to the login screen, even though the logged out view of the deck provides no user value.

dlapiduz:
I had a pull request a while ago for this: https://github.com/18F/cg-deck/pull/87.

Not sure if it still works or what.

jeremiak:
I cherry picked @dlapiduz's code onto the `staging-alpha` branch and pushed to the [jk-force-authentication](https://github.com/18F/cg-deck/tree/jk-force-authentication) branch, but I am still able to see an unauthenticated home page in an Incognito window.

@jcscottiii or @dlapiduz - my Go foo isn't too strong, did I get everything? From my reading of the code I think this should work

PR #365, opened 2016-06-16
Redirect all unauthenticated requests to the login page

jeremiak:
To reduce the number of steps and clicks a user has to take to use the deck, unauthenticated visitors should be redirected to the login page.

@jcscottiii and I achieved this by removing all caching for the server and with @dlapiduz's `LoginRequired` middleware.

Refs #308


dlapiduz:
👏

Issue #373, opened 2016-06-16
Create instance form has accessibility issues.


msecret:
In order for users to successfully use the create instance for on any type of device, it should be section 508 compliant.

Acceptance criteria:
- [ ] Dropdown options should be grouped with `optgroup`.
- [ ] Fieldset should contain a legend element.
- [ ] Field set element must have an accessible name.
- [ ] Form should have some sort of submit element.

msecret:
>
If this selection list contains groups of related options, they should be grouped with optgroup.

msecret:
>
Fieldset does not contain a legend element. All fieldsets should contain a legend element that describes a description of the field group.

msecret:
>
This fieldset element does not have a name available to an accessibility API. Valid names are: legend element.

msecret:
>
This form does not contain a submit button, which creates issues for those who cannot submit the form using the keyboard. Submit buttons are INPUT elements with type attribute "submit" or "image", or BUTTON elements with type "submit" or omitted/invalid.

jeremiak:
A few questions:

> If this selection list contains groups of related options, they should be grouped with optgroup.

Why does this selection list contain multiple groups of related options? Since a marketplace is underneath an organization, shouldn't just the spaces for the current org show up? Currently the behavior is to show all spaces in the `SpaceStore`, but I think that's actually a bug.

> This fieldset element does not have a name available to an accessibility API. Valid names are: legend element.

How did you generate this error? I don't see anything like it in my `HTML_CodeSniffer` output.

msecret:
I generated it with code sniffer

msecret:
In terms of the grouping, I think you're correct, it doesn't need to be grouped. Myabe we should fix the bug where it uses all spaces in spacestore?

jeremiak:
Agreed @msecret. I've got it fixed locally so I'll include that in this PR.

msecret:
Checked in HTML Codesniffer and errors are all gone.

PR #390, opened 2016-06-23
Fix accessibility errors for the service instance creation form 

jeremiak:
The criteria in #373 are:

0. Dropdown options should be grouped with optgroup.
0. Fieldset should contain a legend element.
0. Field set element must have an accessible name.
0. Form should have some sort of submit element.

**Dropdown options should be grouped with optgroup.**
I didn't implement this requirement. Instead, I altered the space store to keep a reference of the parent org on each store and the service instance creation form in the marketplace only shows spaces for the current organization. This was preferable because the "Marketplace" view (on which the service instance creation form resides) is a child view of an organization. Therefore, it didn't make sense to show spaces that are in other organizations in this view.

**Fieldset should contain a legend element.**
I converted the former `<h4>` into a legend element (f1ddc039b2ebf0ea6a3cc9d1cd5f9983c91e29a8)

**Field set element must have an accessible name.**
I never saw this error using `HTML_CodeSniffer`. Maybe adding a `<legend>` fixed it.

**Form should have some sort of submit element.**
The `<Button />` component was modified to accept a `type` property (d0e9f580dd1ea0f7b63606c13687384a50e80d83) and the form button was given a "submit" type (f9092092426065a834512d7055421041727a9c65).

### DIFF
--- a/static_src/components/button.jsx
+++ b/static_src/components/button.jsx
@@ -21,8 +21,9 @@ export default class Button extends React.Component {
 
   render() {
     var classes = classNames(...this.props.classes);
+    let type = this.props.type || 'button';
     return (
-      <button type="button" className={ classes }
+      <button type={ type } className={ classes }
           aria-label={ this.props.label } onClick={ this._handleClick }
           disabled={this.props.disabled}>
         { this.props.children }
@@ -35,6 +36,7 @@ Button.propTypes = {
   classes: React.PropTypes.array,
   label: React.PropTypes.string,
   onClickHandler: React.PropTypes.func,
+  type: React.PropTypes.string,
   disabled: React.PropTypes.bool
 };
 
--- a/static_src/components/button.jsx
+++ b/static_src/components/button.jsx
@@ -20,8 +20,8 @@ export default class Button extends React.Component {
   }
 
   render() {
-    var classes = classNames(...this.props.classes);
-    let type = this.props.type || 'button';
+    const classes = classNames(...this.props.classes);
+    const type = this.props.type;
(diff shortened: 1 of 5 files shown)


controllers/root.go
```diff
@@ -79,7 +79,7 @@ func (c *Context) LoginRequired(rw web.ResponseWriter, r *web.Request, next web.
 	// revisit the server, you will get a bad view due to a caching issue.
 	// for now, we clear the cache for everything.
 	// TODO: revist and cache static assets.
-	rw.Header().Set("cache-control", "priviate, max-age=0, no-cache")
+	rw.Header().Set("cache-control", "no-cache, no-store, must-revalidate, private")
```
msecret: Does this need to be rebased?

static_src/components/button.jsx
```diff
@@ -21,8 +21,9 @@ export default class Button extends React.Component {
 
   render() {
     var classes = classNames(...this.props.classes);
+    let type = this.props.type || 'button';
```
msecret: If this is a new prop, I think it should be added to propTypes and defaultProps

msecret:
LG

controllers/root.go
jeremiak: Yup, I think it does now. I'll do that

static_src/components/button.jsx
jeremiak: Ya, that's a better way to do this. I'll update accordingly

jeremiak:
@msecret I think I addressed your comments now

Issue #383, opened 2016-06-20
Show loading state


msecret:
In order for a deck user to understand that they're data isn't loaded yet, they should see a loading indicator on the page.

Acceptance criteria
- [ ] They don't see contradictory information: no apps and then apps.
- [ ] The single loading indicator shows up somewhere consistent in the content section.
- [ ] The loading indicator goes away when it's confirmed that theres no data OR there is data that comes back.

msecret:
Where should we place the spinner?

msecret:
- Have empty state in content area, when event for getting info is fired, the entire content area will be taken over. As soon as that's done, it will render the non-loading state.
- The empty state is should be not appear before page load before loading state.
- We coudl play with it so we fire the event before that.
- If there is any data already, it won't do the loading spinner.
  - So first time on org, you might see flash of content.
- Prefer to give a loading thing and then give you an empty state.
- They are different states but compositie in visual sense. if it comes back and theres nothing, show empty state, if it comes back with something, show data.
- As long as it doesn't show contradictory info.

jeremiak:
List of loading states to implement:

- [ ] When loading the apps for a space
- [ ] When loading the services for a space
- [ ] When loading any list of users for an organization/space
- [ ] When loading possible services in the Marketplace view
- [ ] When loading an app detail view

Any that I'm missing @msecret?

msecret:
thinking of an example for the apps store:

```
function app() 
  appActions.fetch();
  appActions.fetchStats()

AppStore
on FETCH:
  this._data[loading] = true;

on RESV:
  this._data[loading] = false

AppPage.jsx
render()
  let content = (this.state.loading) ? <span class="spinner">loading</span> : <h4>No apps</h4>
  if (this.state.app) //data exists
    content = <stuff>
```

msecret:
@vz3 @thisisdano are you guys going to be working on the spinner for this?

jeremiak:
Ya, I think that makes a lot of sense for the first implementation @msecret. I'll try it out in the app page first and we can see if we like it

jeremiak:
[...]

PR #398, opened 2016-06-28
[WIP] Adding loading states to the UI

jeremiak:
Refs #383

### DIFF
--- a/static_src/components/app_page.jsx
+++ b/static_src/components/app_page.jsx
@@ -4,6 +4,8 @@ import React from 'react';
 import AppStore from '../stores/app_store.js';
 import RouteList from './route_list.jsx';
 
+window.A = AppStore;
+
 import sectionStyle from 'cloudgov-style/css/components/section.css';
 import createStyler from '../util/create_styler';
 
@@ -14,7 +16,8 @@ export default class AppPage extends React.Component {
     this.props = props;
     this.state = {
       app: AppStore.get(this.props.initialAppGuid) || {},
-      currentAppGuid: this.props.initialAppGuid
+      currentAppGuid: this.props.initialAppGuid,
+      loading: AppStore.fetching
     };
     this._onChange = this._onChange.bind(this);
     this.styler = createStyler(sectionStyle);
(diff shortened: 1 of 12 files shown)


static_src/components/app_page.jsx
```diff
@@ -4,6 +4,8 @@ import React from 'react';
 import AppStore from '../stores/app_store.js';
 import RouteList from './route_list.jsx';
 
+window.A = AppStore;
```
msecret: Remember to remove before finalizing

static_src/components/app_page.jsx
```diff
@@ -48,12 +53,15 @@ export default class AppPage extends React.Component {
 
   render() {
     let content = <h4 className="test-none_message">No app</h4>;
```
msecret: So just to think through, I'd guess this is something like this:

```
let content = <h4>{ (this.state.loading) ? 'Loading...' : 'No app' }</h4>

if (!this.state.loading && this.state.app.name) { // not loading and app has data
  content = 
  ...
```

Not sure if that's valid react now that I look at it. But the point would be to always show loading state if loading and show no data if app data is missing.

static_src/test/unit/util/cf_api.spec.js
```diff
@@ -470,6 +470,29 @@ describe('cfApi', function() {
...
+      expect(fetchAppSpy).toHaveBeenCalledOnce();
+      expect(fetchAppStatsSpy).toHaveBeenCalledOnce();
+    });
+
+    it('should call the receivedAppAll app action', function (done) {
+      const guid = 'shouldCallActionCreatorGuid';
+      const actionCreatorSpy = sandbox.spy(appActions, 'receivedAppAll');
+
+      cfApi.fetchAppAll(guid).then(() => {
+        expect(actionCreatorSpy).toHaveBeenCalled();
```
jeremiak: @msecret this test would always fail if I was expecting the spy `toHaveBeenCalledOnce` with a message that it had been called twice, but I can't figure out why.

From looking at the way this test executes and the code in `cfApi.fetchAppAll()` can you see what makes the `appActions.receivedAppAll` get called twice?

msecret: Hmm, I can't think of anything. When I have problems with things like this, I'll put a `debugger` statement in the action or maybe the spy and then run in in the browser with `watch-test`. Then I'll try and get a good stack trace of whos calling(it doesn't always give me a good stack track). We can check it out tomorrow.

PR #399, opened 2016-06-28
Adding an update notice

brittag:
I don't have a local version of this current version of the Deck set up to test this, but here's a try at putting in a little notice to let people know that a new Deck is coming up.


msecret:
I don't think that emoji will render but otherwise this looks good

brittag:
Ha, I'm curious why it wouldn't render (at least on systems that usually support emoji, such as OS X and iOS), but I'll take it out just in case. :)

msecret:
You're right, it would probably have rendered.

Issue #403, opened 2016-06-30
Service instance creation and viewing bug


msecret:
When a user creates a service instance on the marketplace page, and then goes to the service instance page to see the created service instance, there's a bug due to incomplete information.

To reproduce:
- Go to any org -> marketplace.
- Create a "sql" or similar service instance.
- Give the instance any name that's memorable
- Assign the instance to any space you can view.
- After the instance is created (the create form disappears) go the to space you selected.
- Go to the service instances tab
- Observe that the service instance is not there and there is an error in the JS console.

jeremiak:
Resolved.

PR #405, opened 2016-06-30
Create service bug

msecret:
There was a bug where if you create a service instance and then go to the service instances page, it would fail because of missing data for `updated_at` and `created_at`. This is because in the CF API, when you create a service instance, it delivers back an incomplete representation of it.

I solved this by just fetching the service instance from the server rather then trying to deal with the incomplete one. I did this because it's easier to reason about a complete object that has the same format rather then trying to deal with a differently formatted object, which would split the logic of the codebase.

ref #403

### DIFF
--- a/static_src/stores/service_instance_store.js
+++ b/static_src/stores/service_instance_store.js
@@ -69,6 +69,7 @@ class ServiceInstanceStore extends BaseStore {
       }
 
       case serviceActionTypes.SERVICE_INSTANCE_CREATED: {
+        cfApi.fetchServiceInstance(action.serviceInstance.guid);
         this._createInstanceForm = null;
         this.emitChange();
         break;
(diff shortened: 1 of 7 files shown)


static_src/stores/service_instance_store.js
```diff
@@ -37,6 +37,15 @@ class ServiceInstanceStore extends BaseStore {
         break;
       }
 
+      case serviceActionTypes.SERVICE_INSTANCE_RECEIVED: {
+        const instance = this.formatSplitResponse(
+          [action.serviceInstance])[0];
```
jcscottiii: question (it's been awhile since i've reviewed some of this code so i could be off or forgot something)
below we use `fetchServiceInstance` which uses `fetchOne`. Why do we need to still treat it as an array and return the first one via `[0]`.

jcscottiii:
@msecret looks good. had one question for my own curiosity.

static_src/stores/service_instance_store.js
msecret: @jcscottiii I thought `formatSplitResponse` only works with arrays but you might be right.

Issue #404, opened 2016-06-30
Go test failures should fail the build


jmcarp:
Errors like this should fail the build (I assume!) but don't: https://travis-ci.org/18F/cg-deck/builds/140568911#L3636. @jcscottiii mentioned that `codecheck.sh` might be ignoring test exit codes and probably knows the solution already.

msecret:
yea that's my fault, I wanted to turn some tests off because we needed to get some messaging out and the broken acceptance tests were blocking us. Speaking of which, I'm not quite sure why we're getting failures, weren't not actively developing the go code on this branch.

msecret:
This was fixed by #414

PR #414, opened 2016-07-01
Fix false positive for tests

jcscottiii:
Addresses #404 

There are now two functions
testCmd will just check for the error code of the command.
testCmdOutput will wrap a command in `fgt`. This is useful for commands
that may not return non-zero themselves but will just output something
an error. fgt will take any output and say its non zero.

Also fix the tests from regression from #365


jcscottiii:
@msecret @jmcarp

PR #415, opened 2016-07-01
Remove privileged proxy for converting ids between e-mail addresses

jcscottiii:
Originally, you needed the user id to set roles. Back when this was
corrected the only way to do that was to query for the user info first
by e-mail address to get the id. Now, there are APIs that make it
possible to set roles by e-mail address. This patch removes the need
for the privileged query call. 

Example of the new `by Username` apis to use:
image


jcscottiii:
@msecret

jcscottiii:
I figured that this is the time to do it since the ability to invite someone to an org via e-mail does not exist in the new deck. so when we finally do it, we can use the new api.

Issue #413, opened 2016-07-01
While managing space users, I expect to see space-level permission options rather than org-level options


brittag:
On the staging Dashboard, if I look at "current space users", I see options to change "Org Manager” / "Billing Manager” / "Org Auditor” permissions:

![screen shot 2016-06-30 at 4 58 41 pm](https://cloud.githubusercontent.com/assets/391313/16508062/f225de7c-3ee3-11e6-9f24-e663442bc0be.png)

But if I look at the current Deck/Console/Dashboard, I see "Space Manager” / "Space Developer” / "Space Auditor” in those spots:

![screen shot 2016-06-30 at 4 58 30 pm](https://cloud.githubusercontent.com/assets/391313/16508061/f1091b76-3ee3-11e6-8eb6-3647f9eac3fa.png)

It makes more sense to see the second set of options there, so I suspect the staging Dashboard has a mistake here. Let's check that and fix it.

msecret:
I have to change the code so the type of user is passed down to user role control, which then maps the role to change based on space or org.

jeremiak:
Closed by ^^

PR #422, opened 2016-07-02
Space level perms

msecret:


jeremiak:
Is this for #413?

msecret:
Yes for #413

jeremiak:
Nice work. This is an improvement as the roles are related to the tab interface and so I'm merging now.

There are a few improvements to make in the future:
* the interface hangs when you select a box (it triggers the API request)
* the authorization error isn't pronounced.

PR #423, opened 2016-07-02
Implement basic loading component

jeremiak:
Changes the stores to have a `.fetching` attribute, which will be a boolean. How the stores implement the toggling of this value is up to each individual store. Then the components can check for requests that are currently in flight and use the `<Loading />` component to display a standardized loading message.

We still need to design the loading component, as it is pretty stark right now but it is a first pass.

Refs #383

### DIFF
--- a/static_src/components/loading.jsx
+++ b/static_src/components/loading.jsx
@@ -1,0 +1,25 @@
+
+import React from 'react';
+
+const propTypes = {
+  text: React.PropTypes.string
+};
+
+const defaultProps = {
+  text: 'Loading'
+};
+
+class Loading extends React.Component {
+  render() {
+    return (
+      <div>
+        { this.props.text }
+      </div>
+    );
+  }
+}
+
+Loading.propTypes = propTypes;
+Loading.defaultProps = defaultProps;
+
+export default Loading;
--- a/static_src/actions/app_actions.js
+++ b/static_src/actions/app_actions.js
@@ -35,5 +35,19 @@ export default {
       appGuid,
       app
     });
+  },
+
+  fetchAll(appGuid) {
+    AppDispatcher.handleViewAction({
+      type: appActionTypes.APP_ALL_FETCH,
+      appGuid
+    });
+  },
+
+  receivedAppAll(appGuid) {
+    AppDispatcher.handleServerAction({
+      type: appActionTypes.APP_ALL_RECEIVED,
+      appGuid
+    });
   }
 };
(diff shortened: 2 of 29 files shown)


jeremiak:
This is ready for review @msecret

jcscottiii:
@jeremiak This might be more for another PR / I'm missing something but I'm wondering does this account for a certain model type (e.g. users) doing multiple fetching at the same time on the page. I'm sure it doesn't happen now but just in case it happens in the future.

jeremiak:
@jcscottiii it is a great point. Just to clarify, you're saying that if:
* Request A is made and `fetching` is set to true
* Request B is made and `fetching` is set to true, which means it doesn’t change
* Request A comes back and changes fetching to false and therefore the store is `fetching` false but there is still a request in flight

Therefore, the store is actually waiting for a request but it's `fetching` value doesn't reflect that. This is definitely a problem with this design/implementation. I was trying to keep the mechanism really simple and made that tradeoff. I haven't seen it be a problem with any of the stores yet, but it could just be that my network connection has worked.

Even if we were to keep track of each request going out, it would still be hard to know if every request an interface needed had come back with fresh data or if it was stale. Thoughts on how we might want to do this otherwise?

jcscottiii:
let me think of it this way:
```
# single store emits requests all while rendering a single fetch 
requests (1..N) <- single store (contains) single fetching <-> render single loading state
```

instead let's do it this way make the fetching as part of the component instead of the store. thoughts @msecret @jeremiak ?

msecret:
According to our design, only the store can keep state, such as fetching. But the components can be listening to the store.

jcscottiii:
@msecret ahh yeah. hmmm. so could each component get a way to know that the store is changing it's particular fetching state?

msecret:
Yea I mean, we coudl keep fetching in the stores but separate different fetching out more so each fetching is more specific? Then components could be more specific about what fetching they're waiting for?

jcscottiii:
@msecret i got ya!

msecret:
If this works, we should use it for the time being, and get these merge conflicts fixed. @jcscottiii if you're interested in taking a look at improving the fetching tracking in the next few days as we just discussed, you definitely should, I think it could be a good improvement.

jeremiak:
@msecret @jcscottiii I fixed the merge conflicts so this should be ready for review now

static_src/components/loading.jsx
```diff
@@ -0,0 +1,25 @@
+
+import React from 'react';
+
+const propTypes = {
```
msecret: Is there a reason you defined these into separate variables rather then directly like other components:

```
Loading.proptTypes = {
  text
  ...
```

Is this a better pattern we should adopt?

Issue #396, opened 2016-06-25
I see "Remove User from Org" buttons where I don't have authorization to do that


brittag:
When I look at "All organization users" in the sandbox organization, I see "Remove User from Org" for the other people in the org, but I don't have authorization to remove them, so I'm mildly confused to see that series of buttons. I'd expect to just not see that button if I can't do that. :)

![screen shot 2016-06-24 at 5 49 59 pm](https://cloud.githubusercontent.com/assets/391313/16353829/4bce3dc8-3a36-11e6-8007-f9864d90bac2.png)

If I try clicking it, I get "You are not authorized to perform the requested action", which makes sense, although that error is in a place I might not notice:

![screen shot 2016-06-24 at 6 08 25 pm](https://cloud.githubusercontent.com/assets/391313/16353845/b0531606-3a36-11e6-83af-ed6701762bfe.png)

brittag:
Related, I'm not sure why these permissions have checkboxes for me (under both "current space users" and "all organization users") - I can click the checkboxes but don't have a way to save changes, which is somewhat confusing.

msecret:
@brittag I think the changes should save automatically, but we need more UI to express that. I think it's a separate issue.

brittag:
Thanks, filed at https://github.com/18F/cg-deck/issues/409

msecret:
@brittag do you suggest just removing the buttons, or is there anything else that has to be done with the UI?

brittag:
To solve this bug, I suggest just removing the "remove" buttons for everyone who doesn't have permission to remove others.

I'd also suggest making error messages show up more visibly when they happen (for example, in red text and closer to the action you tried to perform), but that's a different problem and should be handled more generally at some point.

brittag:
Status - this isn't a blocker for switchover as long as we communicate in copy that you might see buttons when you can't actually remove people.

msecret:
The deprecated deck had a uaa codebase that could do things like get permissions for the org:

https://github.com/18F/cg-deck/blob/deprecated/static/app/uaa.js#L61

There also seems to be a `/uaa` route to get current user info:

https://github.com/18F/cg-deck/blob/deprecated/static/app/uaa.js#L33.

So it seems like we could 

```
return http.get('/uaa/userinfo').then(function (response) {
  // Store user data
  userInfoStorage = response;
  return response
}).then(function(userInfo) {
  http.get('/uaa/Users', {
    "userName": user.userName
  })
[...]

PR #438, opened 2016-07-07
Limit current user actions

msecret:
Added code that:
- Get's the current user with guid.
- Adds the current user to the UserStore.
- Allow to check if current user has a certain role
- Has the user UI check the current user for the "org_manager" role before showing the user delete button

This role checking functionality will only work on this users page, because otherwise, the role information for a user may have not been fetched yet and may not be there. I think this is fine for now, as its the only place we need it, and can come up with a more robust solution after.

I'm not 100% certain what user role allows you to delete a user, I'm pretty sure it's org_manager.

An additional thing to fix here would be to also check for this role for the role checkboxes, and allow for role checking in the space for those checkboxes.

Ref #396

### DIFF
--- a/static_src/actions/user_actions.js
+++ b/static_src/actions/user_actions.js
@@ -120,5 +120,13 @@ export default {
       type: userActionTypes.USER_CHANGE_VIEWED_TYPE,
       userType
     });
+  },
+
+  receivedCurrentUserInfo(user) {
+    AppDispatcher.handleServerAction({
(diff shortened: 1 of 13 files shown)


jcscottiii:
rebase please!

jcscottiii:
> I'm not 100% certain what user role allows you to delete a user, I'm pretty sure it's org_manager
this is correct

PR #458, opened 2016-07-11
Govcloud deploy

jcscottiii:
Deploy to both govcloud and east-west region
- Separates manifests by region.
Moves to using cf UPS for credentials like client id and client secret.
- This prevents the .travis.yml from bulking up in terms of encrypted variables. (uses go-cfenv to get those variables, hence the new import and glide.lock and glide.yaml updates)

How manifests are split out now
```bash
manifests
├── eastwest
│   ├── manifest-master.yml
│   ├── manifest-prod.yml
│   ├── manifest-region-base.yml
│   └── manifest-staging.yml
├── govcloud
│   ├── manifest-master.yml
│   ├── manifest-prod.yml
│   ├── manifest-region-base.yml
│   └── manifest-staging.yml
└── manifest-base.yml
```

The govcloud environment has a different org name `cloud-gov` from the eastwest region `cf` There's a new environment variable `CF_ORGANIZATION_GC` to capture that.

The deployer username is the same however the passwords are different hence `CF_PASSWORD_GC`

Finally, the api to login to is different hence `CF_API_GC`

Go code changes: 
Due to the JSON format that the CF environment puts credentials for the UPS, we don't want the typical dev to bootstrap their local development machines with that style. In order to keep the current development setup the same with environment variables, we keep the loading into `envVars` the same. Then there is a new function `loadUPSVars` which will override the `envVars` that load from UPS.


jcscottiii:
@msecret @jeremiak take a look!

msecret:
LGTM, does the Atlas team need to check?

jcscottiii:
@msecret i don't think so. not on concourse.

PR #463, opened 2016-07-12
Content updates for tables, marketplace and app

msecret:
- Updated a bunch of table header titles to include the type of
name.
- Updated marketplace with new content at top and create service form.
- Updated app page to include the space and org name you are in.

@brittag might want to check for content updates, to ensure I didn't mess any up.

### DIFF
--- a/static_src/components/app_page.jsx
+++ b/static_src/components/app_page.jsx
@@ -15,8 +15,8 @@ function stateSetter(current) {
   return {
     app: AppStore.get(current.currentAppGuid) || {},
     currentAppGuid: current.currentAppGuid,
-    currentOrgName: OrgStore.currentOrgName(),
-    currentSpaceName: SpaceStore.currentSpaceName(),
+    currentOrgName: OrgStore.currentOrgName,
+    currentSpaceName: SpaceStore.currentSpaceName,
     loading: AppStore.fetching
   };
 }
--- a/static_src/components/app_page.jsx
+++ b/static_src/components/app_page.jsx
@@ -3,37 +3,47 @@ import React from 'react';
 
 import AppStore from '../stores/app_store.js';
 import Loading from './loading.jsx';
+import OrgStore from '../stores/org_store.js';
+import SpaceStore from '../stores/space_store.js';
 import RouteList from './route_list.jsx';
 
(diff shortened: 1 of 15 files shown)


msecret:
We can ignore the `/pr` tests right? I think we were ignoring them on cg-style because they were wrong?

static_src/stores/org_store.js
```diff
@@ -99,6 +99,12 @@ class OrgStore extends BaseStore {
     return this._currentOrgGuid;
   }
 
+  currentOrgName() {
```
jeremiak: @msecret I think this should use the same get accessor patter that `currentOrgGuid` does

static_src/stores/space_store.js
```diff
@@ -58,6 +58,12 @@ class SpaceStore extends BaseStore {
   get currentSpaceGuid() {
     return this._currentSpaceGuid;
   }
+
+  currentSpaceName() {
```
jeremiak: Same here, let's keep all of these as values instead of returned from functions

static_src/components/app_page.jsx
```diff
@@ -3,37 +3,47 @@ import React from 'react';
...
 import createStyler from '../util/create_styler';
 
+function stateSetter(current) {
+
+  return {
+    app: AppStore.get(current.currentAppGuid) || {},
+    currentAppGuid: current.currentAppGuid,
+    currentOrgName: OrgStore.currentOrgName(),
+    currentSpaceName: SpaceStore.currentSpaceName(),
+    loading: AppStore.fetching
```
jeremiak: Now that we have two more stores in here should loading actually be something more like below?

`loading: AppStore.fetching || OrgStore.fetching || SpaceStore.fetching`

brittag:
Hmm, my [local test version](https://github.com/18F/cg-deck#run-locally-without-needing-go) doesn't have any marketplace data, so I can't easily check that page - but guessing based on the code, these changes look good from my content perspective. Thanks!

static_src/components/app_page.jsx
msecret: I'm not sure it should because it's just the name of the thing, so we'd actually prefer to just show the data when app store is available. I've also accounted for the case when the other two stores are still fetching by just not displaying that part of the name.

Issue #455, opened 2016-07-08
No unit labels for memory numbers on org page


brittag:
On my org page, the "Total Development Memory" and "Total Production Memory" items have numbers without unit labels (MB / GB / etc). Not sure if this has always been the case or if this is a regression.

My account:
![screen shot 2016-07-08 at 12 20 39 pm](https://cloud.githubusercontent.com/assets/391313/16699040/680760b6-4506-11e6-9fc3-fb2b84e35c6d.png)

Local test:
![screen shot 2016-07-08 at 12 20 11 pm](https://cloud.githubusercontent.com/assets/391313/16699045/6c71441e-4506-11e6-88b2-61ac3d64a32a.png)

msecret:
Acceptance criteria:
- [ ] The data in the column should have ` MB` after the unit.

PR #464, opened 2016-07-12
Display memory units to the spaces of an org

jeremiak:
Add "MB" to the memory related columns in the space list on an organization view.

Also refactor so that we don't use `Reactable` in this component anymore.

Refs #455

screen shot 2016-07-12 at 3 18 08 pm


msecret:
Think this needs a rebase

jeremiak:
@msecret rebased and now ready for review

Issue #331, opened 2016-06-08
Simplify maintaining npm dependencies by updating npm version


msecret:
In order for staging deck maintainers to easily update, add, remove npm dependencies, npm should be on version > 3 so the npm commands automatically update the shrinkwrap file

- [x] Run the npm installation, build and tests with node version > 5.0 and npm > 3.0 and ensure it works.
- [x] Add a `.nvmrc` file with new node version.
- [x] The engines hash in `package.json` should reflect new node and npm version.
- [x] Remove `$TRAVIS_NODE_VERSION` env var from travis config, it should use the default nvmrc file now.
- [x] Write a new npm-shrinkwrap file to ensure everything is correct.

msecret:
Background:

Every time we have to update our npm dependencies, including adding new ones, upgrading old ones, or deleting unnecessary ones, we usually have to delete our whole node_modules folder, install again due to the way shrinkwrap works. This process can take 5-10 minutes making maintaining dependencies a burden.

The newer version of npm (> 3.0) includes functionality to automatically change the npm shrinkwrap file eliminating the need to completely rebuild the node_modules folder. This could save 5-10 minutes for every time we have to modify a package.

This work should be taken on when we determine that we have to update packages enough times to justify the amount of time this work will take. So that would mean 5-10 minutes * times we have to update dependencies compared to time this story takes.

msecret:
This npm upgrade changes the structure of the `node_modules` folder so could break any code that directly uses the node modules folder. Here is a list of potential places that would have to be investigated:

```
package.json:11:    "build": "npm run clean && node_modules/webpack/bin/webpack.js",
package.json:12:    "codecov": "cat ./coverage/coverage-final.json | ./node_modules/codecov.io/bin/codecov.io.js",
webpack.config.js:39:          path.resolve(__dirname, 'node_modules/cloudgov-style'),
webpack.config.js:68:    fallback: path.resolve(__dirname, 'node_modules')
```

msecret:
I tried this and the build, the tests and using the site all seemed to work.

jeremiak:
Looks good to me!

PR #467, opened 2016-07-12
Update node/npm to new version

msecret:
Updates node to version 5.12 and npm to version whatever. Main point
was to update npm to version greater then 3 because of all the features
it adds at that point.

It would be nice to merge this before any package.json updates.

Ref #331


jeremiak:
I'm concerned about moving to a `5.x` branch because the odd numbered branches don't ever enter the NodeJS [long term support schedule](https://github.com/nodejs/LTS#lts_schedule). Is the goal to get `npm` anything above `3.0`? If so, I can try this with node `6.x` and still use `npm@3.8`?

jeremiak:
This is great, I'm really impressed with how little had to change to go up two full versions of node! Plus, the [way that npm3 handles dependency resolution](https://docs.npmjs.com/how-npm-works/npm3) seems to make a lot more sense and will hopefully lead to even faster install times since dependencies can be shared.

PR #475, opened 2016-07-14
Track Google Analytics only on production

jeremiak:
We don't want to pollute our user data with development data, so Google Analytics should only be used in production.

Check for the environment variable `NODE_ENV`, and if it is present, then track events in Google Analytics. Otherwise, we do not.

Idea from a [Dan Abramov tweet](https://twitter.com/dan_abramov/status/592692202335301636)

### DIFF
--- a/manifests/eastwest/manifest-prod.yml
+++ b/manifests/eastwest/manifest-prod.yml
@@ -5,3 +5,4 @@ applications:
   host: dashboard
   env:
     CONSOLE_HOSTNAME: https://dashboard.cloud.gov
+    NODE_ENV: production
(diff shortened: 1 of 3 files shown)


jcscottiii:
Are we counting the govcloud environment dashboard to be production to? I don't think so but I would to be explicit somewhere that we aren't considering govcloud as production. Of course, I could be wrong and we could want data from dashboard.fr.cloud.gov

jeremiak:
@jcscottiii I'm totally fine to update the manifest for govcloud as well to say that's production. Honestly, not quite sure what the distinction between the two is. If you consider GovCloud production let's update the manifest to reflect that!

msecret:
Like what are we doing with the GovCloud env?

msecret:
I think we're closing this in favor of #476

PR #476, opened 2016-07-14
Modify new relic config to include SPA

msecret:
SPA is a new nr feature and I think requires an update to the snippet
to work.


static/index.html
```diff
@@ -10,25 +10,37 @@
     <meta property="og:title" content="cloud.gov deck">
     <meta property="og:url" content="https://console.cloud.gov/">
 
+    <script>
+      if (window.location.host === 'dashboard.cloud.gov') {
+        window.PRODUCTION = 1;
```
jeremiak: I'm chill with this since as long as it is defined it will serve the purpose, but I would prefer this to be a boolean of `true`

PR #477, opened 2016-07-14
Move from Travis to CircleCI

jcscottiii:
Due to Travis being being very slow at times, it makes it hard to develop quickly.

This PR is the introduction of the configuration to run on CircleCI.

Characteristics:
`node_modules` is cached between builds. This helps out alot. I have the build run `npm install` beecause I'm assuming if something is updated in package.json, even though `node_modules` is cached, the new stuff will be then updated. Verify that this true? @msecret @jeremiak 
- With the cache, it speeds the build up from an empty cache (~9 mins) to < 4 mins

Weirdness:
- Every command is run into a sub shell in circleCI. This wouldn't matter but for the Go code and the GOPATH I had to move stuff around. As a result, every command I have to `cd` into the new place where the code is.

## TODO
- [x] Deploy scripts
- [x] Replace TRAVIS environment variables with CIRCLE equivalent
- [ ] Enable CircleCi for the project.
- [ ] Put CF login credentials into the UI (they can't be encrypted into the yml like travis)
  - [ ] Add `CF_PASSWORD` - @msecret @jeremiak this is the credentials for eastwest. Do y'all have it?
  - [ ] Add `CF_USERNAME` - This is just `cgdeck-deployer`
  - [ ] ADD `CF_PASSWORD_GC`- I have this... I think since I just set this up.
- [x] Setup deployment in general from circle by installing CF and the zero-downtime-push plugin
- [ ] Add the CircleCI service to codechech.sh for goveralls with token. https://github.com/mattn/goveralls#circleci (Maybe save for later)


circle.yml
```diff
@@ -0,0 +1,30 @@
...
+    - sudo tar -C /usr/local -xzf download/$GODIST
+
+dependencies:
+  cache_directories:
+    - "~/.go_workspace/src/github.com/18F/cg-deck/node_modules/"
+  pre:
+    - go version
+    - go get github.com/Masterminds/glide
+    - mkdir -p /home/ubuntu/.go_workspace/src/github.com/18F/cg-deck && cp -R /home/ubuntu/cg-deck/. /home/ubuntu/.go_workspace/src/github.com/18F/cg-deck
+    - cd /home/ubuntu/.go_workspace/src/github.com/18F/cg-deck && glide install
```
msecret: Would it help to make that long path an env var?

jcscottiii: great idea!!

jcscottiii:
@jeremiak @msecret This should be good to go! We still need to do the TODOs in the UI but the config is done. This could be merged  and then eventually we switch to using circleci (enable access to this repo for CircleCI) after we change the repo name but we can at least already have the config merged.

jeremiak:
@jcscottiii sweet! How do I see the builds working in CircleCI?

package.json
```diff
@@ -82,7 +82,7 @@
     "karma-jasmine": "^0.2.2",
     "karma-jasmine-matchers": "^0.1.3",
     "karma-phantomjs-shim": "^1.0.0",
-    "karma-phantomjs2-launcher": "^0.3.2",
+    "karma-phantomjs-launcher": "^1.0.1",
```
msecret: I think this also requires a change to the `npm-shrinkwrap.json` file. The easiest way to do this is ensure you're on a version of npm greater then 3 (`nvm use` in the repo) and then `npm update --save karma-phantomjs-launcher@^1.0.1`. I *think* newer versions of npm will update the shrinkwrap.

jcscottiii: Just ran this:
```bash
$ nvm install
Found '/Users/jamesscott/Development/Workspaces/mycgdashboard/src/github.com/jcscotiii/cg-deck/.nvmrc' with version <6>
Downloading https://nodejs.org/dist/v6.3.0/node-v6.3.0-darwin-x64.tar.xz...
######################################################################## 100.0%
Now using node v6.3.0 (npm v3.10.3)
$ nvm use
Found '/Users/jamesscott/Development/Workspaces/mycgdashboard/src/github.com/jcscotiii/cg-deck/.nvmrc' with version <6>
Now using node v6.3.0 (npm v3.10.3)
$ npm update --save karma-phantomjs-launcher@^1.0.1
```

Nothing changed.

jcscottiii: what is the `npm-shrinkwrap.json` btw?

jcscottiii:
@jeremiak @msecret you can see the public results here: https://circleci.com/gh/jcscottiii/cg-deck

If you want to mess with the settings yourself, I've sent you collaborator invites for github. Then if you log in to circleci with github, you should have access to the repo settings

package.json
jeremiak: It is kind of like a `Gemfile.lock`. It has the exact version of everything installed including nested dependencies. If it is present, `npm install` will use it instead of `package.json` so we use it to make sure we're all installing the same versions of everything.

It's pretty tedious but it ensures we don't have hard to track down bugs due to us using different versions of dependencies. I think Marco has some work related to improving this in #467 though

jeremiak:
@jcscottiii Cool, and it looks like I can even see that in an Incognito window so its public. Is there something we should do to make it available under 18F instead of a personal Github user?

msecret:
Wait, isn't that URL just temporary so we can see it, @jcscottiii

jcscottiii:
@jeremiak @msecret yeah it's temporary. we need to setup circle ci for `18F/cg-deck`.  if i activate it now before this PR is merged, it will run and spit all these bad errors on the other PRs because they won't have the circle.yml.

Issue #434, opened 2016-07-06
Stopped app usage info says "MB out of MB" instead of "0 MB out of 0 MB"


brittag:
When I look at a stopped app on dashboard-master.cloud.gov, the memory usage and disk usage lines say "MB out of MB":

![screen shot 2016-07-06 at 4 21 26 pm](https://cloud.githubusercontent.com/assets/391313/16637679/b431ee1a-4395-11e6-8920-0ca15549136a.png)

This is confusing to me as a user since it's missing numbers - I'm guessing it should probably say "0 MB out of 0 MB".

This seems helpful to fix because users likely want to know with certainty whether their stopped apps are using any memory (since that affects their quota usage, which affects billing).

msecret:
Acceptance critiera:
- [ ] The UI should say 0 MB.

jeremiak:
I wonder if we should say additional context. Something like:
"Stopped apps do not use any memory"

Thoughts @brittag and @msecret?

brittag:
That'd be nice as an enhancement if it's easy. Let's have it say:

* **Memory usage:** 0 MB. *Stopped apps don't use any memory.*
* **Disk usage:** 0 MB. *Stopped apps don't use any disk space.*

jeremiak:
This looks good on `master`

PR #478, opened 2016-07-14
Format when app has 0 mb of disk or mem

msecret:


jeremiak:
Refs #434

jeremiak:
Screenshots for b7a6010

App running:
screen shot 2016-07-15 at 1 12 32 pm

App stopped:
screen shot 2016-07-15 at 1 13 40 pm

PR #479, opened 2016-07-14
Header scope to applist table for accessibility

msecret:
In order to be accessible, when a table has headers diverging from the
usual vertical headers at the top of the table, a scope has to be added
to explain to a computer where the headers are. Additionally removes
need for strong tags with changes in cg-style

Requires https://github.com/18F/cg-style/pull/111


jeremiak:
🚢

Issue #468, opened 2016-07-13
App route information is missing "domain" information


brittag:
When I look at one of my apps on dashboard.cloud.gov and check the route information, there's no "domain" information available:

![screen shot 2016-07-12 at 5 56 18 pm](https://cloud.githubusercontent.com/assets/391313/16788440/01768f96-485a-11e6-8440-5477effb375b.png)

When I look at this app on console.cloud.gov, the domain information is available, and it says "apps.cloud.gov":

![screen shot 2016-07-12 at 5 56 42 pm](https://cloud.githubusercontent.com/assets/391313/16788441/028d6c56-485a-11e6-8b1d-17ed1e20fac0.png)

This page should display domain information.

This might be one of the issues that is currently blocking the Director of Infrastructure from being able to use dashboard.cloud.gov as effectively as console.cloud.gov (https://github.com/18F/cg-deck/issues/460).

msecret:
Technical implementation:

I'm pretty sure this requires making separate requests to find the domains in an org and match it to the domain of the app. This likely means adding new methods to cf_api and maybe have the app store keep the domains for now (it might not need it's own store yet)

brittag:
Ah Bret helped me figure out that I filed this bug wrong, since "Path" is separate from "Domain".

The bug here is that the "Domain" column is missing. The "Path" column is working fine (it just doesn't exist in the old Deck).

![screen shot 2016-07-14 at 3 57 38 pm](https://cloud.githubusercontent.com/assets/391313/16858449/c228ac74-49db-11e6-9c36-50119e714a16.png)

jeremiak:
Acceptance criteria:
- [ ] A user should be able to see the host, domain and path for all routes bound to an application

brittag:
Hmm, reopening this since it doesn't seem to be working for me. When I look at dashboard-master.cloud.gov, I see this, which seems to be missing domain info:

![screen shot 2016-07-20 at 4 31 39 pm](https://cloud.githubusercontent.com/assets/391313/17006683/757e7924-4e97-11e6-9533-7e6392099555.png)

On console.cloud.gov I see this:

![screen shot 2016-07-20 at 4 32 28 pm](https://cloud.githubusercontent.com/assets/391313/17006700/9a178474-4e97-11e6-883a-f5f7a50f5134.png)

msecret:
@jeremiak maybe not all the data is actually in the API? Or its in a different place? Or maybe there's just some default values when the data is empty?

[...]

PR #482, opened 2016-07-15
Show route host, domain and path in app view

jeremiak:
Current:
screen shot 2016-07-15 at 11 46 37 am

PR:
screen shot 2016-07-15 at 11 45 31 am

Refs #468

### DIFF
--- a/static_src/stores/base_store.js
+++ b/static_src/stores/base_store.js
@@ -112,7 +112,7 @@ export default class BaseStore extends EventEmitter {
    * @param {fn} cb - defaults to calling this.emitChange()
    *                  when the store's data has changed
    *
-   */
+  */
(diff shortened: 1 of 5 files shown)


static_src/stores/base_store.js
```diff
@@ -136,6 +136,48 @@ export default class BaseStore extends EventEmitter {
     return cb(true);
   }
 
+  /* mergeAll
+   *
+   * If the dataToMerge exists in the store (as found by
+   * the mergeKey), then merge in new data if there are
+   * changes. The merge algorithm differs from the one
```
msecret: Just so I understand, the difference between this and mergeMany is this will merge when set of data into mutliple different items in the store?

PR #483, opened 2016-07-15
Use sorting in render to sort orgs and spaces

msecret:
That was easy


jeremiak:
Nice. If we end up doing this a lot maybe we can implement a common method on BaseStore or something.

PR #492, opened 2016-07-19
Fix initial load of space page

msecret:
Initial load of space page was failing with currentOrg.guid undefined
because the org probably isn't fetched yet.

This PR ensures the space component checks to make sure a currentOrg
is around before doing shit with it.


jeremiak:
Looks good, merging.

PR #494, opened 2016-07-19
More fixes for initial space load

msecret:
- Moves service instance fetch to the main file, otherwise two
  dispathces happen at the same time.
- Makes an additional data check in user store.

Both these were reporting errors before.

### DIFF
--- a/static_src/main.js
+++ b/static_src/main.js
@@ -56,6 +56,9 @@ function space(orgGuid, spaceGuid, potentialPage) {
   cfApi.fetchOrg(orgGuid);
   spaceActions.fetch(spaceGuid);
   // TODO use constant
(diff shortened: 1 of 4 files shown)


static_src/main.js
```diff
@@ -56,6 +56,9 @@ function space(orgGuid, spaceGuid, potentialPage) {
   cfApi.fetchOrg(orgGuid);
   spaceActions.fetch(spaceGuid);
   // TODO use constant
+  if (potentialPage === 'services') {
+    serviceActions.fetchAllInstances(spaceGuid);
```
jeremiak: 👍

jeremiak:
Looks good to me!

PR #497, opened 2016-07-20
Rename to dashboard in deploy process

msecret:


jcscottiii:
It looks good but these applications need to be manually changed one time so that 1) zero downtime will pick it up and 2) CF will see these as two separate apps and the new cg-dashboard app will see cg-deck currently uses the same url and will fail. we can pair and do this transition.

jcscottiii:
Made one change. makes it easier to search between prod and master in logs.cloud.gov

jcscottiii:
Changed the current spaces and app names

PR #498, opened 2016-07-20
Ms change go names

msecret:
Build off https://github.com/18F/cg-deck/pull/497, that has to be merged first.


jcscottiii:
in slack:
> @jeremiak: looks green to me

> @msecret: Ship it

PR #509, opened 2016-07-22
Show domains and routes in app view

jeremiak:
When an `APP_RECEIVED` event is fired, the `app` object that comes along with it contains all the route and domain information we need. So on that event, we can update the route store instead of only relying on sending a subsequent network request. If the `app` object doesn’t contain a child `routes` array then we will just ignore the event.

Also changed the `appGuid` key in the route store to `app_guid` to be consistent with the kinds of data structures we keep in our stores.

Refs #468

### DIFF
--- a/static_src/components/route_list.jsx
+++ b/static_src/components/route_list.jsx
@@ -8,7 +8,7 @@ import createStyler from '../util/create_styler';
 
 function stateSetter(appGuid) {
   const routes = RouteStore.getAll();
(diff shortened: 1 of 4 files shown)


msecret:
LGTM

brittag:
I don't have a good way to test this locally with my own cloud.gov account, since my local copy is just the test demo version, but it at least doesn't seem to introduce any new problems in what I can see. :)

PR #511, opened 2016-07-25
Finish GET fixtures

msecret:
This adds the rest of the fixtures for getting data, doesn't add routes for when data
is modified.

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -7,7 +7,10 @@ var services = data.services;
 var serviceInstances = data.serviceInstances;
 var servicePlans = data.servicePlans;
 var spaces = data.spaces;
-var users = data.users;
+var currentUser = data.currentUser;
+var spaceUsers = data.spaceUsers;
+var orgUsers = data.orgUsers;
+var orgUserRoles = data.orgUserRoles;
 
 var BASE_URL = '/v2';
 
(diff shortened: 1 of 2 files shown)


jeremiak:
This is totally cool, and adds a whole bunch of data to the app!

At this point all of the views should be visible to the user, but they still won't be able to do any write actions such as creating apps, service instances, etc.

test/api.js
```diff
@@ -42,7 +42,11 @@ module.exports = function api(smocks) {
     label: 'App stats',
     path: `${BASE_URL}/apps/{guid}/stats`,
     handler: function (req, reply) {
-      reply(appStats);
+      var app = apps.filter((a) => a.guid === req.params.guid).pop();
+      var response = Object.assign({}, appStats, {
+        state: app.state
```
msecret: Should we also set current mem and disk to 0 or null so it's a more correct response?

msecret: Wait, doesn't `stats` not return anything for stopped apps?

jeremiak: Yup, you're right. Looks like it returns the following `400` error:

```
{
  "code": 200003,
  "description": "Could not fetch stats for stopped app: help",
  "error_code": "CF-AppStoppedStatsError"
}
```

I'll fix that, nice catch

PR #513, opened 2016-07-26
Move test folder so just one folder

msecret:
It deeply bothers me that there's two separate test folders so I
moved it all into one test folder under src. If we had other types of
JS tests that aren't unit, they'd go under that folder too.


jeremiak:
Not sure why these tests are failing after 4d3b817

static_src/tests.bundle.js
```diff
@@ -1,3 +1,3 @@
-var context = require.context('./test', true, /.+\.spec\.jsx?$/);
+var context = require.context('./test/unit', true, /.+\.spec\.jsx?$/);
```
jeremiak: I'm confused why this matters, since I would've hoped that the local command failed as well.

jeremiak:
Nice, this looks good. I'm pretty confused why that 8385cfc makes a difference but now that its green I'm going to merge

PR #526, opened 2016-07-29
Have to move non-prod apps to apps domain

msecret:


jeremiak:
This looks good, merging now.

PR #527, opened 2016-07-29
Refactor fetching to not emit a change event

msecret:
Refactors fetching code so it doesn't emit a change event.

This allows more precise control to the developer when a change event should happen. The exact problem that lead to this was when new service plans were coming in, they weren't being shown because a change event wasn't firing because fetching was already set to false. Might need to further expand this pattern of fetching.


jeremiak:
Nice, I like this much better. It is good to have to explicitly call `this.emitChange` so that each store has a bit more control over when it is "loading" and when it is done.

PR #529, opened 2016-07-30
Refactor fetching

msecret:
Refactors fetching to use a `fetched` property which separately tracks whether something has been fetched or not. This allows us to test if something is empty because it wasn't fetched yet or because it was fetched but theres nothing there.

The user page is definitely still broken, that needs to be fixed by including another route on the router for the two separate user pages.

### DIFF
--- a/static_src/components/app_container.jsx
+++ b/static_src/components/app_container.jsx
@@ -17,6 +17,7 @@ function stateSetter(current) {
     currentAppGuid: current.currentAppGuid,
     currentOrgName: OrgStore.currentOrgName,
     currentSpaceName: SpaceStore.currentSpaceName,
+    empty: AppStore.isEmpty(),
     loading: AppStore.fetching
   };
 }
@@ -120,11 +121,18 @@ export default class AppContainer extends React.Component {
   }
 
   render() {
-    let content = <h4 className="test-none_message">No app</h4>;
+    let loading = <div></div>;
+    let content = <div></div>;
 
     if (this.state.loading) {
-      content = <Loading text="Loading app now" />
-    } else if (this.state.app.name) {
+      loading = <Loading text="Loading app now" />;
+    }
+
+    if (this.state.isEmpty) {
+      content = <h4 className="test-none_message">No app</h4>;
+    }
+
+    if (this.state.app.name) {
       content = (
         <div>
           <h2>{ this.fullTitle }</h2>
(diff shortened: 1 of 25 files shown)


static_src/components/app_container.jsx
```diff
@@ -120,11 +127,18 @@ export default class AppContainer extends React.Component {
...
-      content = <Loading text="Loading app now" />
-    } else if (this.state.app.name) {
+      loading = <Loading text="Loading app now" />;
+    }
+
+    if (this.state.empty) {
+      content = <h4 className="test-none_message">No app</h4>;
+    }
+
+    if (appReady(this.state.app)) {
```
msecret: Should this always be else if?

static_src/components/app_list.jsx
```diff
@@ -71,11 +73,18 @@ export default class AppList extends React.Component {
...
-      content = <Loading text="Loading apps" />;
-    } else if (this.state.apps.length) {
+      loading = <Loading text="Loading apps" />;
+    }
+
+    if (this.state.empty) {
+      content = <h4 className="test-none_message">No apps</h4>;
+    }
+
+    if (this.state.apps.length) {
```
msecret: Should this always be else if?

static_src/stores/org_store.js
```diff
@@ -25,22 +25,24 @@ class OrgStore extends BaseStore {
       case orgActionTypes.ORG_FETCH: {
         cfApi.fetchOrg(action.orgGuid);
         this.fetching = true;
+        this.fetched = false;
```
msecret: For all stores, do we need to emit change here because these values changed?

msecret:
I wrote some comments to call out my concerns to the reviewer.

static_src/components/app_container.jsx
jeremiak: These are three mutually exclusive states right? I think it would be more clear if that was in `else if` blocks

static_src/stores/org_store.js
jeremiak: We don't need to, but perhaps we should? I feel like emitting changes should be considered a cheap operation and will just cause React to determine if diffs should be flushed to the DOM and act accordingly.

In this case, I think we should emit a change event. When we make a request, we're changing the state of the store and I imagine all views that subscribe to the store care about this particular value change since it likely means they will show a `<Loading /`> component.

jeremiak:
Nice work. While having to manually control `fetching` seems like more work on the face of it, having the flexibility for each store to determine how it will set its state makes sense. Also, having `fetched` and `fetching` is awesome sot hat we don't have to infer specific states from data not being there.

Issue #515, opened 2016-07-28
Activity log for cloud foundry events per application


jeremiak:
In order for developers to troubleshoot their applications, they need to understand what recent system events occurred.

Acceptance criteria:
- [ ] An activity list of [platform events](https://apidocs.cloudfoundry.org/226/events/list_all_events.html) with timestamps
- [ ] Timestamps should be in the format of "Mar 09 2017   05:17:24"
- [ ] Crash events should be prominent
- [ ] Any changes due to user activity (quota changes, services bound, ssh logins, etc) should be prominent but different from crash events

jeremiak:
@vz3 Can you add the design comps you showed me earlier to this issue so I can continue to build against them?

vz3:
## Pattern
![activity log](https://cloud.githubusercontent.com/assets/582918/17533889/c7a35b80-5e54-11e6-8a15-2ba037e5aefc.png)

## Example
![activity log - example](https://cloud.githubusercontent.com/assets/582918/17533888/c796136c-5e54-11e6-84c2-9624a754f73d.png)

vz3:
Quick iteration on this:
-Removes background color band.
-Removed glyph for more subtle visual element.
-Bolded text for priority events.
-Removes redundant dates: date should only be showed in first instance. (Need to validate usability on this)

![activity log](https://cloud.githubusercontent.com/assets/582918/17566451/6a05a450-5f09-11e6-8ccf-5bf8f8e764a2.png)

PR #533, opened 2016-08-05
Add basic activity log component to app view

jeremiak:
Adds a basic activity log component in the app view. The component is currently build from the CF `/spaces/:spaceGuid/events` endpoint and crudely handles a few different event types right now but could be expanded to handle events from different data sources.

Though the endpoint is called "events", I use "activity" in the application to help differentiate between the component and the data, and because we will likely include data from sources beyond the CF events API. The data is therefore stored in the `ActivityStore`, which responds to `activityActions`.

This depends on some new changes to `cloudgov-style` but it should still be intelligible without that.

A user can click on the timestamp to have it formatted as a date or as a relative time stamp, and while we're developing the features a user can click on the activity log item text and see the raw `event` JSON.

Refs #515


static_src/components/activity_log_item.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+  get content() {
+    let content = `${this.props.item.type} isn't handled`;
+    const item = this.props.item;
+    const metadata = item.metadata;
+    const route = RouteStore.get(metadata.route_guid);
+    const url = (route) ? `${route.host}${route.domain}/${route.path}` : 'a url';
+    const link = (route) ? (<a href={ `//${url}` }>{ url }</a>) : url;
+
+    // TODO: if route is not found, trigger fetch action to get it
```
msecret: I actually don't think actions should be called in JSX components because they lead to two dispatches happening at same time which leads to error.

static_src/components/activity_log_item.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+    if (item.type === 'app.crash') {
```
msecret: Can this be made into a map or a switch?

msecret: Now that I think about it, I'm not sure it matters, feel free to keep as is.

static_src/components/activity_log_item.jsx
jeremiak: Hmm, I think that makes sense. Where would we trigger this action then?

static_src/components/activity_log_item.jsx
jeremiak: I do like it better as a switch, so I'm going to convert it to that

static_src/components/activity_log_item.jsx
msecret: I dunno, most come from the route. Or from a store. we can think about it.

static_src/components/activity_log.jsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  _onChange() {
+    this.setState(stateSetter(this.props));
+  }
+
+  render() {
+    return (
+      <div>
+        <h1>{ this.props.title }</h1>
```
jeremiak: @msecret Do you think `<h1>` is the most semantic thing to use here? It is a component and therefore might not be the only `<h1>` on the page. Another thing we can do is wrap it in a `<section>` instead of a `<div>` since I think each section can have its own heading hierarchy.

static_src/components/activity_log.jsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  _onChange() {
+    this.setState(stateSetter(this.props));
+  }
+
+  render() {
+    return (
+      <div className={ this.styler('activity_log-container') }>
+        <h1>{ this.props.title }</h1>
```
jeremiak: @msecret Do you think this should be an `<h1>` semantically? If not, maybe this should just be a `<strong>`? It is a heading, and it is a component so it might not be the only heading on the page. I think we can also change the wrapper `<div>` to be a `<section>` since each section can have its own heading hierarchy.

jeremiak:
I still need to update the way each `<ActivityLogItem />` handles the text so that we don't get awkward copy such as below:

screen shot 2016-08-11 at 4 55 17 pm

jeremiak:
@msecret I think this is now ready, and I'd ask that you check your applications for weird copy on crash events.

static_src/components/activity_log_item.jsx
```diff
@@ -65,6 +65,27 @@ export default class ActivityLogItem extends React.Component {
...
+        content = `the app instance exited with ${metadata.exit_status} status`;
+        break;
+      case 'out of memory':
+        content = 'it ran out of memory';
+        break;
+      case 'failed to accept connections within health check timeout':
+        content = `it ${metadata.exit_description}`;
+        break;
+      default:
+        content = 'bad things happen';
```
msecret: Shouldn't it be "happened"?

jeremiak: Definitely. Updating...

static_src/components/activity_log.jsx
msecret: My first guess would it would be an `h2` or `h3`, but I don't always have the best understanding of how heading tags should work throughout the page.

static_src/actions/activity_actions.js
```diff
@@ -0,0 +1,42 @@
+
+
+import AppDispatcher from '../dispatcher.js';
+import { activityActionTypes } from '../constants';
+
+export default {
+  // fetch(appGuid) {
```
msecret: Should this be removed?

static_src/test/unit/util/cf_api.spec.js
```diff
@@ -510,6 +595,40 @@ describe('cfApi', function() {
...
+
+      stub.returns(testPromise);
+
+      cfApi.fetchAppLogs(expected).then(() => {
+        expect(spy).toHaveBeenCalledOnce();
+        done();
+      });
+
+
+      // let actual = spy.getCall(0).args[0];
```
msecret: Should these be enabled?

jeremiak: No, I think it should actually be removed. I'll do that and rebase master

static_src/actions/activity_actions.js
jeremiak: Ya, I think so. I'll do that in a bit.

static_src/test/unit/stores/activity_store.spec.js
```diff
@@ -0,0 +1,139 @@
...
+        logs
+      });
+
+      expect(ActivityStore.getAll().length).toEqual(1);
+      expect(spy).toHaveBeenCalledOnce();
+    });
+  });
+
+  describe('parseLogItem()', function () {
+    it('should parse log items', function () {
```
msecret: What does parse log items mean? What's it actually doing?

jeremiak: The logs API returns big strings that represent log items (such as the one below), so parsing is using RegExp to parse out the data into named variables

```
  {
    "message": "dashboard.cloud.gov - [19/08/2016:13:56:53.239 +0000] \"GET / HTTP/1.1\" 302 0 310 \"https://login.cloud.gov/\" \"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36\" 127.0.0.1:45574 x_forwarded_for:\"50.139.84.209\" x_forwarded_proto:\"https\" vcap_request_id:ed21da85-42eb-4107-bce0-6805e20b9900 response_time:0.166580506 app_id:ff813a1f-a455-404e-a283-c9addc60ae29\n"
  }
```

Issue #521, opened 2016-07-28
User page shows "no users" before users appear


msecret:

msecret:
Something that has to be done separately to fix the space store is a new route for org vs space needs to be added, so we only make the request for one of them.

Otherwise we'd have to distinguish fetching and fetched for org vs space, which seems like too much work.

adborden:
Looks good, seeing the loading animation before the users appear.

PR #535, opened 2016-08-09
Separate user page routes

msecret:
Fixes the loading state problems on the user page by having separate routes for org and space users and only loading the correct corresponding user types.

Also refactors the space router so each page has it's own function that all use a few central method for spaces.

### DIFF
--- a/static_src/components/main_container.jsx
+++ b/static_src/components/main_container.jsx
@@ -9,17 +9,23 @@ import Disclaimer from './disclaimer.jsx';
 import Header from './header.jsx';
 import Login from './login.jsx';
 import LoginStore from '../stores/login_store.js';
+import OrgStore from '../stores/org_store.js';
+import SpaceStore from '../stores/space_store.js';
 import { Nav } from './navbar.jsx';
 
-function getState() {
-  return { isLoggedIn: LoginStore.isLoggedIn() };
+function stateSetter() {
+  return {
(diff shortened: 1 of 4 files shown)


static_src/components/main_container.jsx
```diff
@@ -9,17 +9,23 @@ import Disclaimer from './disclaimer.jsx';
...
 import LoginStore from '../stores/login_store.js';
+import OrgStore from '../stores/org_store.js';
+import SpaceStore from '../stores/space_store.js';
 import { Nav } from './navbar.jsx';
 
-function getState() {
-  return { isLoggedIn: LoginStore.isLoggedIn() };
+function stateSetter() {
+  return {
+    currentOrgGuid: OrgStore.currentOrgGuid,
```
msecret: In another PR, I'll be moving to this pattern of always getting the guids from the stores rather then passing the initial one down for `main`. So components will consistently get these from the stores rather then from two places (props and stores). Is that an improvement?

jeremiak: Definitely an improvement. It also helps to keep the stores as the single source of truth about the state of the data. If we're going to have the stores keep track of what app/space/org is the "current" one, then the components should use the store to get that information.

jeremiak:
This looks good to me! I like having a URL for each view

msecret:
Ref #521

Issue #491, opened 2016-07-19
Make component state on initial load and after clicking around consistent


msecret:
I've noticed a good amount of bugs are due to differences in state between the initial page load and clicking through things. I was thinking we could refactor the code somehow so we could improve consistency, reducing bugs.

Acceptance criteria
- [ ] Non `initial[Org|Space]Guid` props should be passed around, they should all use the stores.
- [ ] There should be no more bugs related to data on initial load not being there but being there on going back to the page.

msecret:
A large part of this will be keeping the convention of getting the currnet space and current org guid from the stores rather then passing it from `main`. This will mean that components are getting the data from the same place always, leading to more consistency.

PR #537, opened 2016-08-10
Refactor initial load

msecret:
Refactors initial load by having all components use the stores to get the initial org/space guids. This helps developers and reduces bugs because where you get the data is consistent, leading to less mistakes around not passing around props properly. 

Builds off #529, merge that first

Def #491

### DIFF
--- a/static_src/main.js
+++ b/static_src/main.js
@@ -48,7 +48,6 @@ function org(orgGuid) {
 function space(orgGuid, spaceGuid) {
   orgActions.toggleSpaceMenu(orgGuid);
   spaceActions.changeCurrentSpace(spaceGuid);
-  // TODO what happens if the space arrives before the changelistener is added?
   cfApi.fetchOrg(orgGuid);
   spaceActions.fetch(spaceGuid);
 }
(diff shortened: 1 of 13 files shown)


static_src/main.js
```diff
@@ -93,7 +92,7 @@ function app(orgGuid, spaceGuid, appGuid) {
   appActions.fetch(appGuid);
   appActions.fetchStats(appGuid);
   ReactDOM.render(
-    <MainContainer initialSpaceGuid={ spaceGuid }>
+    <MainContainer>
       <AppContainer
         initialAppGuid={ appGuid }
```
jeremiak: Shouldn't AppStore also follow this pattern and maintain the current application guid with an action to change it?

jeremiak:
Looks good to me! Just noted that some of the stores don't follow this pattern though it might make sense if they did, such as the AppStore.

static_src/main.js
msecret: Yea I guess it should, should we do it now though or wait?

jeremiak: Let's merge now and then open a new PR just for the current app stuff. The reason for this is that there are a lot of upcoming changes in the app container and it will be easier to think about the conflicts with fewer changes.

PR #542, opened 2016-08-12
Configure karma to only report test failures

jeremiak:
There used to be a lot of things that would get printed to the console when the tests were run. It was capturing `console.log` output from the console and showing tests that passed. All of this was a lot of information, and made finding failing tests a bit more cumbersome.

I am now capturing all console.log output so nothing gets reported and have introduced a new karma test reporter to only show when a test fails.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -15,28 +15,33 @@ module.exports = function(config) {
 
       plugins: [
         'karma-chrome-launcher',
+        'karma-failed-reporter',
         'karma-jasmine',
         'karma-jasmine-matchers',
         'karma-phantomjs2-launcher',
         'karma-phantomjs-shim',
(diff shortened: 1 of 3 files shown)


karma.conf.js
```diff
@@ -15,28 +15,33 @@ module.exports = function(config) {
...
+        'static_src/tests.bundle.js': ['webpack']
       },
 
       webpack: webpackConfig,
 
-      reporters: ['progress'],
+      reporters: ['failed'],
+
+      client: {
+        captureConsole: false
```
msecret: There's time when we might need console output. Does it still show up when running tests in browser? Shoudl it be configurable?

jeremiak: Good point, it should be configurable. I'll make that change

Issue #543, opened 2016-08-16
Create loading indicator for loading state component


msecret:
In order for users to easily see when some content is loading, there should be a visual loading indicator.

Acceptance criteria:
- [ ] The indicator should only appear after 300ms to 500ms of data not loading.
- [ ] The loading state should hide all the content until it is all loaded, even on second page loads.

adborden:
Confirmed the loader exists on the user management page.

PR #544, opened 2016-08-17
Add loading component

msecret:
Adds a more proper loading indicator which appears after ~400ms and blanks out the whole component it's part of.

Builds on https://github.com/18F/cg-dashboard/pull/537 merge that first

Ref #543

### DIFF
--- a/static_src/components/app_container.jsx
+++ b/static_src/components/app_container.jsx
@@ -128,12 +128,9 @@ export default class AppContainer extends React.Component {
   }
 
   render() {
-    let loading = <div></div>;
     let content = <div></div>;
 
-    if (this.state.loading) {
-      loading = <Loading text="Loading app now" />;
-    }
+    let loading = <Loading text="Loading app now" active={ this.state.loading } />;
 
     if (this.state.empty) {
       content = <h4 className="test-none_message">No app</h4>;
@@ -142,6 +139,7 @@ export default class AppContainer extends React.Component {
         <div>
           <ActivityLog initialAppGuid={ this.state.app.guid } title="Recent activity" />
           <h2>{ this.fullTitle }</h2>
+          { loading }
           <section className={this.styler('section-card')}>
             <h3>About this application</h3>
             <table>
@@ -207,7 +205,6 @@ export default class AppContainer extends React.Component {
 
     return (
       <div>
-        { loading }
         { content }
       </div>
     );
--- a/static_src/components/app_container.jsx
+++ b/static_src/components/app_container.jsx
@@ -128,18 +128,16 @@ export default class AppContainer extends React.Component {
   }
 
   render() {
-    let content = <div></div>;
-
-    let loading = <Loading text="Loading app now" active={ this.state.loading } />;
+    let loading = <Loading text="Loading app" />;
+    let content = <div>{ loading }</div>;
 
     if (this.state.empty) {
       content = <h4 className="test-none_message">No app</h4>;
-    } else if (appReady(this.state.app)) {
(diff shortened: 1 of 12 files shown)


msecret:
cg-dash-loading

vz3:
@msecret In the case of the examples in which the whole screen is blank, is there any way to show some contextual info before load? (i.e. more like the "___ space in your ___ organization" example)

let me know if that doesn't make sense

msecret:
@vz3 so show the header for the page essentially? So "Spacename" in your "orgName" organziation?

vz3:
@brittag Would love to get your thoughts on this. Sometimes we can get headers, but sometimes information on the headers is part of what's loading. So what do we show users in the meanwhile?

@msecret once this is live, do you expect the delay to be about as short as it is in the GIF?

msecret:
I'll try and explain a better example:

On the app page, when it first loads the heading could be more like

### _ application in your _ space which is in your *cf* org

Notice the "_" where there's missing information. We could also code it to not show parts that aren't loaded yet but then end up with something like:

### application which is in your *cf* org

Something to keep in mind with that ^ approach, we should be careful to not make the logic of what displays too complicated to ensure we can use it as a shared coding pattern across the site.

When it's finally loaded, the pieces will fall into place to make the full heading

### *dashboard* application in your *staging* space which is in your *cf* org

brittag:
I tried checking out this branch to try to use it and see how it feels, but something wasn't right:

```
ERROR in ./static_src/components/loading.jsx
Module not found: Error: Cannot resolve module 'cloudgov-style/img/loading.gif' in /Users/brittagustafson/git-projects/cg-dashboard/static_src/components
 @ ./static_src/components/loading.jsx 35:15-56
```

So, a couple speculative thoughts since I haven't actually tried it:

I like the concept of showing *_* application in your *_* space which is in your *cf* org, but I suspect that in practice, this would look weird since I suspect the words would distractingly jump as the info loads.

Also, if we expect the loading delay to be brief for most users, showing something very briefly might be a mildly frustrating experience for the reader ("what did I just miss?").

msecret:
Based on what I've seen so far:
- The text does get jumpy when new parts of it load. Less so on pages with a smaller heading, more so on pages like the app page which have 3 parts of information
- The loading delay usually is very brief, often less then a second.

brittag:
Interesting! Brief + jumpy = I'd be gently inclined toward no contextual info before load.

vz3:
@brittag Sorry I'm struggling a bit at communicating this clearly:
None of the contextual info I'd propose would disappear upon load. At least that's not how I was thinking about it.

@msecret Though now that I'm typing this out, I wonder if in _any_ case we'd have the appropriate to build any sort of (useful) header before load?

If that's the case, I agree with Britta, let's just do a blank screen and a loader. That still feels really not ideal to me, but I'll need to see it on a test server. But I think in an ideal state we have a header which at least provides feedback to the user immediately about where they are at.

brittag:
@vz3 My thinking is that since our headers aren't static (as in our current style with `_ application in your _ space which is in your _ org`), if we fleshed them out after load, users could have the impression that something had disappeared, since there would be a quick flash of something and then the loaded info. I'm still just speculating though since I haven't seen this in practice. :)

msecret:
I propose we continue with what we have for now, and then maybe the 3 of us could pair to together for half and hour so I could show you guys what the other option of having a header would look like on various pages under various network speeds.

jeremiak:
Nice, this looks good per the discussion above. I added (per @msecret's recommendation) a timeout to some of the route handlers `/static_src/test/server/api.js`

PR #545, opened 2016-08-19
Refactor current app guid

msecret:
Rather then passing the current app guid to the app container from main, this makes the container look for it in the store. The main function then changes it with a new change current action.

This is done to keep consistent with how space and org stores/containers are doing it. Sub components of app container will not be changed.

Built off #544


jeremiak:
This looks good to me!

PR #546, opened 2016-08-22
Update cg-style to 1.1.0

msecret:


msecret:
This should probably wait on https://github.com/18F/cg-style/pull/131 so we get all changes in

PR #547, opened 2016-08-23
Show empty state for activity component

msecret:


static_src/test/unit/stores/service_plan_store.spec.js
```diff
@@ -137,6 +137,7 @@ describe('ServicePlanStore', function() {
 
     it('should set fetching to false, fetched to true', function() {
       ServicePlanStore.fetching = true;
+      ServicePlanStore.waitingOnRequests = false;
```
jeremiak: What is this variable used for?

static_src/stores/activity_store.js
```diff
@@ -66,35 +66,43 @@ class ActivityStore extends BaseStore {
...
       case activityActionTypes.EVENTS_FETCH:
+        this.fetching = true;
+        this.fetched = false;
         cfApi.fetchSpaceEvents(action.spaceGuid);
+        this.emitChange();
         break;
 
       case activityActionTypes.EVENTS_RECEIVED:
+        this.fetching = false;
+        this.fetched = true;
```
jeremiak: Do you think there will be a potential state where the activity log will be wrong if `.fetched` gets set to true here? The activity log currently checks to see if `.fetched` is true AND the length of the data array is 0. But the events fetch request is initiated before the logs fetch request, so there can be a time when the "No recent activity" condition can be met in the component, but one of the requests is still in flight/hasn't been handled.

I think this is probably something we can live with for a while and see if there are problems with the usability before doing more work, or a lightweight check could be to have `_` variables on the store for each request and the `.fetched` can be a derived property from both?

static_src/test/unit/stores/service_plan_store.spec.js
msecret: It's so the service plan store can wait for all service plan requests to come through before setting fetching and fetched.

static_src/stores/activity_store.js
msecret: Yea I did think of this. Most other stores that are listening for more then one set of data, I decide what data is most important, but in this case, they're on equal footing. I think I'd prefer to get something out for now, but will try testing it out.

jeremiak: Onboard with that

static_src/test/unit/stores/service_plan_store.spec.js
jeremiak: Cool, that makes sense. Maybe its something we can refactor into the base store since there are a few stores that have to be aware of multiple requests.

jeremiak:
🚀

PR #548, opened 2016-08-23
Use separate fetched and fetching on activity

msecret:
As discussed in #547, in order for the activity store to be fetched
both events and logs should be fetched (and fetching same thing).

This changes the store to track each separately and provide the same
fetched and fetching API but based on the two separate things tracked.

### DIFF
--- a/static_src/stores/activity_store.js
+++ b/static_src/stores/activity_store.js
@@ -60,33 +60,33 @@ class ActivityStore extends BaseStore {
     super();
     this._data = new Immutable.List();
     this.subscribe(() => this._registerToActions.bind(this));
-    this.eventsFetched = false;
-    this.eventsFetching = false;
-    this.logsFetched = false;
-    this.logsFetching = false;
+    this._eventsFetched = false;
+    this._eventsFetching = false;
+    this._logsFetched = false;
+    this._logsFetching = false;
   }
 
   get fetched() {
-    return this.eventsFetched && this.logsFetched;
+    return this._eventsFetched && this._logsFetched;
   }
 
   get fetching() {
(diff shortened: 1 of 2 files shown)


static_src/stores/activity_store.js
```diff
@@ -60,21 +60,33 @@ class ActivityStore extends BaseStore {
     super();
     this._data = new Immutable.List();
     this.subscribe(() => this._registerToActions.bind(this));
+    this.eventsFetched = false;
+    this.eventsFetching = false;
+    this.logsFetched = false;
+    this.logsFetching = false;
+  }
+
+  get fetched() {
```
jeremiak: I really like this! By making the store interface include these values, we can enable any of the stores to calculate them on their own.

static_src/stores/activity_store.js
```diff
@@ -60,21 +60,33 @@ class ActivityStore extends BaseStore {
     super();
     this._data = new Immutable.List();
     this.subscribe(() => this._registerToActions.bind(this));
+    this.eventsFetched = false;
```
jeremiak: Minor quibble, but it seems that we've been using `_` prefixed variable names for things that only the store should care about. Should these variables follow that convention or are they intended for public use?

jeremiak:
This looks good, just made the point above about private variables. If that's a moot point, let's merge!

static_src/stores/activity_store.js
msecret: Agreed they should be private

jeremiak: Cool, I just made that change

jeremiak:
I made the changes for prefixing private variables with an `_` so merging now

PR #558, opened 2016-08-25
Services panel data

msecret:
This PR addresses the data changes required to get the service panel UI working.

### DIFF
--- a/static_src/stores/service_binding_store.js
+++ b/static_src/stores/service_binding_store.js
@@ -1,0 +1,38 @@
+/*
+ * Store for service bindings data. Bindings are what link apps to service
+ * instances.
+ */
+
+import Immutable from 'immutable';
+
+import AppDispatcher from '../dispatcher';
+import BaseStore from './base_store.js';
+import cfApi from '../util/cf_api.js';
+import { serviceActionTypes } from '../constants.js';
+
+class ServiceBindingStore extends BaseStore {
+  constructor() {
+    super();
+    this.subscribe(() => this._registerToActions.bind(this));
+    this._data = new Immutable.List();
+  }
+
+  _registerToActions(action) {
+    switch (action.type) {
+      case serviceActionTypes.SERVICE_BINDINGS_FETCH: {
+        cfApi.fetchServiceBindings(action.appGuid);
+        this.fetching = true;
+        this.fetched = false;
+        this.emitChange();
+        break;
+      }
+
+      default:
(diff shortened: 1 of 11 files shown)


static_src/stores/service_plan_store.js
```diff
@@ -62,6 +62,27 @@ class ServicePlanStore extends BaseStore {
...
       }
 
+      case serviceActionTypes.SERVICE_INSTANCES_RECEIVED: {
+        const instances = this.formatSplitResponse(action.serviceInstances);
+        this.fetched = false;
+        this.fetching = true;
+        this.emitChange();
+        const planRequests = [];
+        for (const instance of instances) {
+          planRequests.push(cfApi.fetchServicePlan(instance.service_plan_guid));
```
jeremiak: Should we check to see if we have that service plan before making network requests? Or is there a risk that the data might have changed and therefore we always want to fetch the data?

msecret: That's a good point. Most of our logic now just fetches the entities and doesn't check that they're there. But in reality, these plans aren't changing so it's pretty safe to check.

Should we set a pattern to not fetch by default and apply it here?

jeremiak: I like the strategy of not fetching by default, as long as it is easy to override.

We could do something simple like:
``` 
planRequests = instances.filter((instance) => {
  return !this.get(instance.service_plan_guid);
}).map((instance) => {
  return cfApi.fetchServicePlan(instance.service_plan_guid);
});
```

static_src/stores/service_binding_store.js
```diff
@@ -0,0 +1,52 @@
...
+        cfApi.fetchServiceBindings(action.appGuid);
+        this.fetching = true;
+        this.fetched = false;
+        this.emitChange();
+        break;
+      }
+
+      case serviceActionTypes.SERVICE_BINDINGS_RECEIVED: {
+        const bindings = this.formatSplitResponse(action.serviceBindings);
+        this.mergeMany('guid', bindings, () => { });
```
jeremiak: This empty callback is kind of weird. What about calling lines 38-40 from within the callback to ensure its after all the bindings have been merged?

jeremiak:
@msecret This looks good, though I made two small changes that you should take a look at. I think this is otherwise ready to merge and start on the UI

jeremiak:
Rebased with `master`

Issue #559, opened 2016-08-25
A user should be able to see all of their bound routes in a panel interface


jeremiak:
Refs #517

suprenant:
In order for developers to view which public URLs where their apps are available, they need a list of routes that are currently bound to the application.

Acceptance criteria:
- [ ] The user can see the number of routes bound
- [ ] The user can see the individual public URLs that are currently bound
- [ ] The user can see how many routes that are available given the user's quota

Refs 18f/cg-design#5

PR #564, opened 2016-08-29
Add view for app routes

jeremiak:
Adds a few components for #559 so that the user can see all of the application's routes.

It also adds a bunch of `<Panel*>` components that form the UI elements in which we display the routes, and soon to be used to display services and memory related stats of the application.

Adds a QuotaStore (and actions, constants, and cfApi calls) to get all quotas for a user so that we can know how many routes a given org/space is allowed to have.

Right now, all the styles are in `panel.css` in this project, but that should be refactored into a cg-style component soon.

Refs #559
Refs 18F/cg-design#5


static_src/components/app_settings_panel.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+import OrgStore from '../stores/org_store.js';
+import QuotaStore from '../stores/quota_store.js';
+import SpaceStore from '../stores/space_store.js';
+
+import createStyler from '../util/create_styler';
+
+const propTypes = {
+  title: React.PropTypes.string,
+  initialAppGuid: React.PropTypes.string.isRequired
```
msecret: I think we shoudl technically move this to use the app store at some point?

static_src/components/panel_action.jsx
```diff
@@ -0,0 +1,48 @@
...
+  handleClick: React.PropTypes.func
+};
+
+const defaultProps = {
+  disabled: false,
+  handleClick: (e) => e,
+  type: 'link'
+};
+
+export default class PanelAction extends React.Component {
```
msecret: I can take a look at moving this to the new `actions.jsx`

static_src/components/panel_header.jsx
```diff
@@ -0,0 +1,23 @@
...
+export default class PanelHeader extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+    this.styler = createStyler(style, panelCss);
+  }
+
+  render() {
+    return (
+      <div className={ this.styler('panel-header') }>
```
msecret: Should this use the `<header>` tag?

static_src/components/route_form.jsx
```diff
@@ -0,0 +1,129 @@
...
+
+    return changed;
+  }
+
+  render() {
+    const route = this.props.route;
+    const domains = this.props.domains;
+
+    return (
+      <form className={ this.styler('route-form') }>
```
msecret: Forgot about this, but did you see https://github.com/18F/cg-dashboard/blob/master/static_src/components/form.jsx? Maybe this form is too hard to fit into that framework though.

static_src/components/route_list.jsx
```diff
@@ -37,57 +45,65 @@ export default class RouteList extends React.Component {
...
+          <PanelAction text="Add route" />
+        </PanelHeader>
+        { this.state.routes.map((route) => {
+          const fullRoute = (route.path) ?
+            `${route.host}.${route.domain}/${route.path}` :
+            `${route.host}.${route.domain}`;
+          const handler = this._handleRouteAction.bind(this, route.guid);
+          let rowContent;
+          if (route.editing) {
+          // if (true) {
```
msecret: Can this ben taken out?

static_src/components/app_settings_panel.jsx
jeremiak: Agreed

static_src/components/route_list.jsx
jeremiak: Yup, that was some debugging code

static_src/components/route_form.jsx
msecret: I have a feeling this form is very specific, like it even has its own CSS right?

jeremiak: Right now it definitely does. For example, the big "." and "/" in the UI are with CSS and pseudo elements. If we decide we want to make it into a generic component that could just be text in a component, but even then it would need some custom CSS.

msecret:
I'm good to merge this, are you good to merge considering my changes?

jeremiak:
From 5f7da08cb7106d7478e781003163175ae02c2b17:
> Question, for the styles, should we be defining them so precise to
> the actual styles in cg-style? For instance, rather then making a button
> of stle "outline", maybe we should make a button of style or function
> cancel, or confirm. So the vars are more based on functionality then
> their styleing?

Definitely! I think once we have names for them as molecules in the pattern library we should refactor this to use those names!

jeremiak:
@msecret This looks good to me, merge freely! I like this pattern of building components that take most of their values via children so that the parent views make more sense.

Issue #518, opened 2016-07-28
View services in a panel


jeremiak:
In order for developers to keep their apps running, they need to know which services are currently bound, which are available to be bound, and the size of each service.

Acceptance criteria:
- [ ] The panel should show the service limit per the org/space quota
- [ ] The panel should clearly show the number of currently bound services
- [ ] The panel should clearly show bound service instances with an icon of that service for ease of recognition
- [ ] The panel should show currently unbound service instances that can be added to the application

msecret:
Based on mockups here: https://github.com/18F/cg-design/issues/5#issuecomment-241795547 both view and edit functionality is a little too much work for 2-3 days of work, so I'm making this story about the view aspect.

msecret:
How to get the data:

Get all bound services to an app:
`GET /v2/apps/$app_guid/service_bindings`

Which gives `service_instance_guid`.

Get the full service instance from the service bindings
`GET /v2/service_instances/$service_instance_guid`

Get all unbound services for the space
`GET /v2/spaces/$space_guid/service_instances`

For each bound and unbound service:

Get the service plan with the service instance guid
` GET /v2/service_plans/$service_instance_guid`

From the service plan, get the `name`, and the cost with `extra: costs:[{"amount":{"usd":0},"unit":"MONTHLY"}]`.

With the service instance, use the `last_operation.type|state` to know what's going on with the instance.

msecret:
### Store design


```js
// main.js
function app(spaceGuid, appGuid) {
  cfApi.fetchServiceBindings(appGuid);
  cfApi.fetchServiceInstances(spaceGuid);
}
```

```js
// service_instance_store
on serviceActionTypes.SERVICE_BINDINGS_RESV:
  each action.bindings
    serviceReqs.push(cfApi.fetchService(binding.service_instance_guid));
  if (serviceReqs.length) this.waitingOnRequests = true;
  Promise.all(serviceReqs).then((res) => {
    this.waitingOnRequests = false;
    fetching = false;
    fetched = true;
    emitChange();
  });

  on serviceActions.SERVICE_INSTANCE_RESV:
     if (!this.waitingOnRequests) {
       this.fetching = false;
       this.fetched = true;
     }
     this.merge('guid', instance, () => { });
     ...
```

msecret:
### Component design

```js
// service_instance_list_panel.jsx
propTypes = {
  instances: array,
  bound: boolean
}

render() {
  <Panel>
[...]

PR #565, opened 2016-08-31
Services panel ui

msecret:
Provides a UI to see what services instances you have on a space and what instances you have currently bound to an app.

Uses newest versions of visual designs and design here: https://github.com/18F/cg-design/issues/6

Ref #518


msecret:
Start on visual design:

screen shot 2016-08-31 at 5 17 17 pm

msecret:
With error state

screen shot 2016-08-31 at 5 32 23 pm

msecret:
Formatting changes and adding red failed text

screen shot 2016-08-31 at 5 48 04 pm

vz3:
Could you change $0 monthly to "Free" (assuming that's accurate)

static_src/components/service_instance.jsx
```diff
@@ -0,0 +1,77 @@
...
+  serviceInstance: React.PropTypes.object,
+  bound: React.PropTypes.bool
+};
+
+const defaultProps = {
+  serviceInstance: {},
+  bound: false
+};
+
+export default class ServiceInstanceListPanel extends React.Component {
```
msecret: Rename to `ServiceInstance`

static_src/components/service_instance.jsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  get cost() {
+    const cost = ServicePlanStore.getCost(this.props.serviceInstance.servicePlan);
+    if (cost === 0) return 'Free';
+    return `$${cost.toFixed(2)} monthly`;
+  }
+
+  render() {
+    let content = <div></div>;
```
msecret: What should happen when content is missing? Shoudl we have separate loading?

static_src/components/service_instance.jsx
```diff
@@ -0,0 +1,77 @@
...
+        <div>
+          <h5 className={ this.styler('panel-column') }>
+            { serviceInstance.name }
+          </h5>
+          <span className={ this.styler('panel-column') }>
+            { serviceInstance.servicePlan &&
+              <span>{ serviceInstance.servicePlan.name }</span>
+            }
+            <br />
+            { serviceInstance.servicePlan &&
```
msecret: What should the cost be if service plan not there yet? probably dont' display?

msecret: Move check to cost method?

static_src/components/service_instance_list_panel.jsx
```diff
@@ -0,0 +1,48 @@
...
+const defaultProps = {
+  serviceInstances: [],
+  bound: false
+};
+
+export default class ServiceInstanceListPanel extends React.Component {
+  constructor(props) {
+    super(props);
+    this.props = props;
+    this.state = {};
```
msecret: Not needed.

static_src/components/service_instance_panel.jsx
```diff
@@ -0,0 +1,104 @@
...
+};
+
+function stateSetter() {
+  const currentSpaceGuid = SpaceStore.currentSpaceGuid;
+  const currentAppGuid = AppStore.currentAppGuid;
+  let serviceInstances = ServiceInstanceStore.getAllBySpaceGuid(
+    currentSpaceGuid);
+  serviceInstances = serviceInstances.map((serviceInstance) => {
+    const servicePlan = ServicePlanStore.get(serviceInstance.service_plan_guid);
+    if (ServicePlanStore.getAll().length) {
```
msecret: debug, remove

static_src/stores/service_instance_store.js
```diff
@@ -36,6 +44,28 @@ class ServiceInstanceStore extends BaseStore {
       serviceInstance.space_guid === spaceGuid);
   }
 
+  getInstanceState(serviceInstance) {
+    const lastOp = serviceInstance.last_operation;
+    if (!lastOp) return 'running';
+
+    if (lastOp.state === 'failed') {
+      return 'failed';
```
jeremiak: Would it be better to have these as constants instead of just strings?

static_src/stores/service_instance_store.js
```diff
@@ -69,6 +99,7 @@ class ServiceInstanceStore extends BaseStore {
         break;
       }
 
+      /*
```
jeremiak: Should this comment be removed?

jeremiak:
I think we should always be setting store variables and such in the callback to the `merge*` functions so that when the change is emitted, all the data is ready for the view. For example, [this code from the service binding store](https://github.com/18F/cg-dashboard/blob/ms-services_ui/static_src/stores/service_binding_store.js#L38-L40) should be in the callback:

```
      case serviceActionTypes.SERVICE_BINDINGS_RECEIVED: {
        const bindings = this.formatSplitResponse(action.serviceBindings);
        this.mergeMany('guid', bindings, () => { });
        this.fetching = false;
        this.fetched = true;
        this.emitChange();
        break;
      }
```

static_src/components/service_instance_list_panel.jsx
```diff
@@ -0,0 +1,54 @@
...
+import style from 'cloudgov-style';
+import React from 'react';
+
+import PanelRow from './panel_row.jsx';
+import ServiceInstance from './service_instance.jsx';
+
+import createStyler from '../util/create_styler';
+
+const propTypes = {
+  serviceInstances: React.PropTypes.array,
```
jeremiak: Should this sort of information come straight from the stores? Right now [`<RouteList />`](https://github.com/18F/cg-dashboard/blob/master/static_src/components/route_list.jsx#L5-L15) uses a bunch of stores so so that the component is isolated and doesn't need many props.

I think this is nice because that keeps the app container component a bit cleaner and doesn't have a bunch of logic about data filtering/processing to hand off to child components.

Of course, the benefit of having a prop is that the API is super clear and its obvious how to use the the component or have it display different data.

Maybe we can have these two methods for now and see how most of the other panel components shake out? I'd think that we'd want them all to function in pretty similar manners for ease of comprehension.

jeremiak:
This looks super slick! I left a few comments, but then I think it is ready to merge

static_src/components/service_instance_list_panel.jsx
msecret: So this doesn't end up in app container (which I'd agree shouldn't take care of everything), it ends up in `service_instance_panel`. The reason I did it like this was so each instance list could get a bunch of instances and  wouldn't worry if they're bound or not. But I ended up not adding "bound" as a property on the instance so I still have to pass whether they're bound or not. 

One benefit of it being here rather then each instance list; I think they'd both have to do the fetching of the same elements, where here it's just done once.

msecret:
I rolled up all things into #568, closing this and that should be merged.

Issue #560, opened 2016-08-25
A user should be able to edit any existing routes and add a new one


jeremiak:
Refs #517

suprenant:
In order for developers to managing the public URLs where their apps are available, they need features such as the ability to:

- Create new routes
- Update existing routes
- Delete existing routes
- Have their delete action confirmed using https://github.com/18F/cg-style/blob/master/src/css/components/actions.scss#L7
- Prevent updating or creating routes such that they collide with an existing route

Acceptance criteria:
- [ ] The user can click a button that Add Route granting the ability to add a new route
- [ ] The user can click an Edit Route link to update a route
- [ ] The user can click a Delete Route link to delete a route
- [ ] The user can see a Delete Action confirmation when deleting a route
- [ ] The user can see a notification if they attempt to update or create routes that collide with an exiting route
- [ ] The user can see buttons to Cancel or Apply their changes 
- [ ] The user can only create a route that has a domain that has been assigned within the organization

Refs 18f/cg-design#5

PR #566, opened 2016-09-01
Create, update, or delete routes

jeremiak:
Adds some UI so that users can create routes (which are subsequently bound to the application), update routes, or delete routes.

There are some weird aspects of the UI right now. For example, there is a count of routes in the panel header. The "of" number is derived from the space quota (the org quota if there is no space quota) `total_routes`. However, it is weird to provide this in the application context because that number isn't necessarily reserved for the application, but for all the apps in the space. Indeed, if the user has a space quota route limit of 10, and has 9 unbound routes created in the space, the user can actually only create one more route for the app even though the UI makes it look like they can create 10.

To get around this we should figure out how to show all the routes for the space (those that are bound to other apps and those that aren't bound at all)

The tests are also breaking for me locally but I'm unclear why at the moment.

Refs #560

### DIFF
--- a/static_src/components/route_form.jsx
+++ b/static_src/components/route_form.jsx
@@ -14,40 +14,56 @@ import routeFormCss from '../css/route_form.css';
 const propTypes = {
   domains: React.PropTypes.array.isRequired,
   route: React.PropTypes.object,
-  handleSubmit: React.PropTypes.func,
-  handleCancel: React.PropTypes.func
+  submitHandler: React.PropTypes.func,
+  cancelHandler: React.PropTypes.func,
+  deleteHandler: React.PropTypes.func
 };
 
 const defaultProps = {
   route: {},
-  handleSubmit: () => {},
-  handleCancel: () => {}
+  submitHandler: () => {},
+  cancelHandler: () => {}
 };
 
 export default class RouteForm extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      domain: props.route.domain,
+      domain_name: props.route.domain_name,
+      domain_guid: props.route.domain_guid, // snake_case since it's an api arg
+      guid: props.route.guid,
       host: props.route.host,
       path: props.route.path
     };
     this._onChange = this._onChange.bind(this);
+    this._onSubmit = this._onSubmit.bind(this);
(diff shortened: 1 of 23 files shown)


jeremiak:
Adding and deleting routes interactions:
adding-and-deleting-routes

static_src/components/route_form.jsx
```diff
@@ -6,56 +6,81 @@ import Action from './action.jsx';
...
+  cancelHandler: React.PropTypes.func,
+  deleteHandler: React.PropTypes.func
 };
 
 const defaultProps = {
   route: {},
-  handleSubmit: () => {},
-  handleCancel: () => {}
+  submitHandler: () => {},
+  cancelHandler: () => {}
```
msecret: Why is there no default for `deleteHandler`?

static_src/components/route_list.jsx
```diff
@@ -2,92 +2,177 @@
...
+
+  _removeCreateRouteForm() {
+    routeActions.hideCreateForm();
+  }
+
+  _createRoute(route) {
+    const { spaceGuid } = this.state;
+    const domainGuid = route.domain_guid;
+    routeActions.createRoute(domainGuid, spaceGuid, route);
+    // routeActions.createRouteAndBindForApp(appGuid, domainGuid, spaceGuid);
```
msecret: Should be removed?

static_src/components/route_form.jsx
jeremiak: The form doesn't always have a delete handler. For example, when creating a route there is nothing to delete

static_src/components/route_list.jsx
jeremiak: Yup, removed. Actually that whole function is no longer relevant so I pulled it out.

msecret:
When creating a new route, and you haven't selected the domain yet, it says "undefined".

image

Is there something else it's supposed to do based on design?

msecret:
@jeremiak this is looking pretty good besides that minor issue. Do you want me to fix that? Or do you have time?

jeremiak:
@msecret If you could take a stab at it that would be great. This part of my afternoon is chock full of meetings

PR #568, opened 2016-09-02
Design app page

msecret:
Built off #565


msecret:
In implementing the visual design for this, I realized we're making some styling changes that should technically be global and effect the whole cloud.gov brand. Of that list, are there any that we're not ready to commit to right now?

I'm of the opinion that we should go for it, as I don't think any will break design.

Global style changes to consider:

- [ ] Primary color to #157DE6
- [ ] New or replace gray color #8E8E8E
- [ ] Background gray color #F3F4F8
- [ ] Font weight of 500 on at least content sections
- [ ] Links changed to bold? font weight and underlined
- [ ] Different sizes for headings? <-- this could be kept panel specific
- [ ] Outline buttons to be black outline rather then primary
- [ ] New "success" color

msecret:
screen shot 2016-09-01 at 6 25 21 pm

msecret:
image

static_src/components/action.jsx
```diff
@@ -1,11 +1,12 @@
 
 import React from 'react';
 import style from 'cloudgov-style/css/cloudgov-style.css';
+import panelCss from '../css/panel.css';
 
 import createStyler from '../util/create_styler';
 
 const BUTTON_STYLES = [
-  'gray',
+  'cautious',
```
jeremiak: 👍

static_src/components/app_container.jsx
```diff
@@ -56,32 +56,6 @@ export default class AppContainer extends React.Component {
     this.setState(stateSetter());
   }
 
-  logsLink(appName) {
```
jeremiak: Woot! 🎉 

Glad to get ride of this link for more useful UI

jeremiak:
This looks pretty good. In my quick review, I noticed just a few things we should fix before merging:

- [ ] Cannot add a route for some reason
- [ ] Deleting a route works, but redirects you back to the initial dashboard route
- [ ] We should show the app state (RUNNING, STOPPED, etc) to provide context as to why it is using no memory

PR #574, opened 2016-09-07
Fetch shared and private domains separately

msecret:
The main domain action defaults to "private" as this isn't used yet.
The route store will check if the route uses a shared domain route or
a private domain route in the domain url and then uses the appropriate
cf api method. This fixes the bug where shared domains would never be
fetched, such as `apps.cloud.gov` domains.


static_src/test/unit/stores/route_store.spec.js
```diff
@@ -6,6 +6,7 @@ import '../../global_setup.js';
 import AppDispatcher from '../../../dispatcher.js';
 import cfApi from '../../../util/cf_api.js';
 import { wrapInRes, unwrapOfRes } from '../helpers.js';
+import routeActions '../../../actions/route_actions.js';
```
juliaelman: I think this might be missing a `from` to complete the import?

juliaelman:
Looks good and made one comment! 👍

PR #575, opened 2016-09-07
Fix loading state in app service instances

msecret:
It wasn't turning off when there were no service instances.


juliaelman:
LGTM

PR #588, opened 2016-09-10
Error handling

msecret:
Changing error handling so errors not related to request failures throw (making it to new relic and local devs) and request failures just go to new relic.

What was going on before these changes:
Due to how the promise code was working in `cf_api`, when an error happened almost anywhere in the application, such as in the jsx components or stores, those errors would go through the `catch` method in the promises returned in cf_api. These would almost all go to `errorActions.errorFetch` where they would be `console.log`ed. This was bad because:

- during local development, developer wouldn't get a full stack trace for their error, just the message from `console.log`.
- none of these errors would be "bubbled" up to the browser, so we had no errors in new relic.

What's going on now:
We're determining when an error in `catch` is a request error (like a fetch request to the cloud foundry api failed) or whether it's a general application error. If it's a request error, we'll send it just to New Relic and console.log it. If it's an actual error, we'll throw it because it should be caught.

I also changed a bunch of code in the cf_api tests so they correctly were async.

### DIFF
--- a/static_src/util/cf_api.js
+++ b/static_src/util/cf_api.js
@@ -24,6 +24,7 @@ function handleError(err, errHandler = errorActions.errorFetch) {
     } else {
       errHandler(err);
     }
+    analytics.noticeError(err);
   // Other exceptions should be thrown so they surface.
   } else {
     throw err;
--- a/static_src/util/cf_api.js
+++ b/static_src/util/cf_api.js
@@ -2,6 +2,7 @@
 import http from 'axios';
 
 import activityActions from '../actions/activity_actions.js';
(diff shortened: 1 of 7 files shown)


static_src/util/cf_api.js
```diff
@@ -15,6 +16,25 @@ import userActions from '../actions/user_actions.js';
 
 const APIV = '/v2';
 
+function handleError(err, errHandler) {
+  let handler = errHandler;
```
jcscottiii: you should initialize this to `errorActions.errorFetch` and the following conditional should check for the existence (instead of the absence) of `errHandler` and assign `errHandler` to `handler` if it exists

PR #590, opened 2016-09-12
Route errors handling data portion

msecret:
This addresses all the data that's needed to get error handling working for route errors. The UI should be able to use this data to show the errors to the user.

PR #591, opened 2016-09-12
Route errors ui

msecret:
See https://github.com/18F/cg-dashboard/commit/31fb668b70f658de268ee93bce4715d44c46da12

Builds off https://github.com/18F/cg-dashboard/pull/590

### DIFF
--- a/static_src/components/route_form.jsx
+++ b/static_src/components/route_form.jsx
@@ -164,8 +164,6 @@ export default class RouteForm extends React.Component {
                 return <FormError message={this.props.error.description} />;
               } else if (route.error) {
                 return <FormError message={route.error.description} />;
-              } else {
-                return <div></div>;
               }
             })()}
           </div>
--- a/static_src/components/route_form.jsx
+++ b/static_src/components/route_form.jsx
@@ -3,6 +3,7 @@ import style from 'cloudgov-style/css/cloudgov-style.css';
 import React from 'react';
 
 import Action from './action.jsx';
+import { FormError } from './form.jsx';
 import PanelActions from './panel_actions.jsx';
 
 import createStyler from '../util/create_styler';
@@ -13,6 +14,7 @@ import routeFormCss from '../css/route_form.css';
 const propTypes = {
   domains: React.PropTypes.array.isRequired,
   route: React.PropTypes.object,
+  error: React.PropTypes.object,
   submitHandler: React.PropTypes.func,
   cancelHandler: React.PropTypes.func,
   deleteHandler: React.PropTypes.func
@@ -20,6 +22,7 @@ const propTypes = {
 
 const defaultProps = {
   route: {},
+  error: null,
   submitHandler: () => {},
   cancelHandler: () => {}
 };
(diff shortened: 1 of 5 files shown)


msecret:
screen shot 2016-09-12 at 4 45 37 pm

screen shot 2016-09-12 at 4 45 56 pm

static_src/components/route_form.jsx
```diff
@@ -155,6 +158,17 @@ export default class RouteForm extends React.Component {
...
             value={ this.fullUrl }
           ></input>
+          <div style={{ paddingLeft: '.75rem' }}>
+            {(() => {
+              if (this.props.error) {
+                return <FormError message={this.props.error.description} />;
+              } else if (route.error) {
+                return <FormError message={route.error.description} />;
+              } else {
+                return <div></div>;
```
jcscottiii: is it possible to return nothing instead of a an empty div tag?

static_src/components/route_form.jsx
```diff
@@ -155,6 +158,17 @@ export default class RouteForm extends React.Component {
             className={ this.styler('route-form-preview') }
             value={ this.fullUrl }
           ></input>
+          <div style={{ paddingLeft: '.75rem' }}>
+            {(() => {
+              if (this.props.error) {
+                return <FormError message={this.props.error.description} />;
```
jcscottiii: just out of curiosity, what are the two sources of error for? i know where the route one comes from but what populates the `this.props.error`. 

and is it possible that there is an error in both `this.props.error` and `route.error` wondering if we should surface both of them in that case??

static_src/stores/route_store.js
```diff
@@ -76,6 +76,8 @@ class RouteStore extends BaseStore {
         cfApi.createRoute(domainGuid, spaceGuid, host, path).then((res) => {
           const routeGuid = res.metadata.guid;
           cfApi.putAppRouteAssociation(appGuid, routeGuid);
+        }).catch(() => {
+          // Do nothing
```
jcscottiii: why do nothing? a comment about why would be helpful.

static_src/components/route_form.jsx
msecret: I can try

static_src/components/route_form.jsx
msecret: Yea I think I might amend the naming.

One source of error is when creating a new route. That's very different because it's not attached to a specific route, because the route wasn't created yet, hence doesn't have a guid. This one comes from `props.error`.

The other error is when you try and update/delete an existing route and something goes wrong. This error is associated with a route so we place it on the actual route object.

When both have an error, we give precedence to the specific route error. This is an actual case that can happen where you have both the create new route and edit route forms open and two separate error messages.

Any suggestions on naming that would make it more clear?

static_src/stores/route_store.js
msecret: I found that if I didn't put that catch there then it would continue in the then callback which would break things. I'll just say actual error handling is done in the api code.

static_src/components/route_form.jsx
jcscottiii: just a comment here and two lines below to indicate that this is for creation error display vs the other is for _RUD

jcscottiii:
after you address the comments, is it still WIP (wondering if i should merge at that time)

msecret:
@jcscottiii maybe we should merge the cg-style PR first as this requires it: https://github.com/18F/cg-style/pull/145/files

PR #593, opened 2016-09-13
Exit tests when webpack build fails

msecret:
Otherwise all tests in the file that has a syntax error will just
not be run, leading to incomplete tests.

This was a problem for our dev team as we had 60 tests not
running without us even knowing.


jcscottiii:
@msecret oh wow. is this something that we should bring up to the rest of 18F frontend to use?

msecret:
@jcscottiii I don't think a large amount of our projects use the webpack - karma combo. Which is unfortunate in a way, we have about 50 combinations of tools to do the same thing.

PR #594, opened 2016-09-13
Add more app page fake data

msecret:
To keep the node testing server up to date with fake data so it's usable by designers and developers.

### DIFF
--- a/static_src/test/server/fixtures.js
+++ b/static_src/test/server/fixtures.js
@@ -277,6 +277,7 @@ var services = serviceGuids.map(function(guid, i) {
 module.exports.services = services;
 
 var serviceInstances = serviceInstanceGuids.map(function(guid, i) {
+  var servicePlanGuid = servicePlanGuids[i];
   return {
     metadata: {
       guid: guid,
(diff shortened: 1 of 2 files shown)


static_src/test/server/fixtures.js
```diff
@@ -376,3 +420,31 @@ var spaceUsers = users.map(function(user, i) {
 spaceUsers.pop();
 
 module.exports.spaceUsers = spaceUsers;
+
+var serviceInstanceBindings = appGuids.map(function(appGuid, i) {
+  var serviceInstanceGuid = serviceInstanceGuids[i];
+  return {
+    metadata: {
+      guid: serviceInstanceBindingGuids[i],
+      url: `${URL_BASE}/service_bindings/${serviceInstanceBindingGuids[i]}`,
```
jcscottiii: replace `serviceInstanceBindingGuids[i]` with `serviceInstanceGuid`

static_src/test/server/fixtures.js
```diff
@@ -165,9 +183,30 @@ var organizations = organizationGuids.map(function(guid) {
 
 module.exports.organizations  = organizations;
 
+var domains = domainGuids.map(function(guid, i) {
+  var domainType = (i % 2 === 0) ? 'private_domains' : 'shared_domains';
```
jcscottiii: you have this logic here and on line 207. maybe make a helper function like `getDomainType(i)`

jcscottiii:
what's the process of doing this? is it automated? i'm wondering is there a way to document where things should go in these files. thinking how to ease the curve for future devs. this is something for another PR though.

msecret:
Yea I just thought about how it would be cool if the initial data was automated, so you could pull in in somehow and then change a few things by hand.

PR #600, opened 2016-09-19
Route associating

msecret:
To associate a route has a single action, not with creation first.

### DIFF
--- a/static_src/components/route.jsx
+++ b/static_src/components/route.jsx
@@ -2,7 +2,9 @@
 import React from 'react';
 
 import Action from './action.jsx';
+import DomainStore from '../stores/domain_store.js';
 import PanelActions from './panel_actions.jsx';
+import RouteForm from './route_form.jsx';
 import createStyler from '../util/create_styler';
 import formatRoute from '../util/format_route';
 import routeActions from '../actions/route_actions.js';
@@ -28,6 +30,7 @@ export default class Route extends React.Component {
     this._unbindHandler = this._unbindHandler.bind(this);
     this._bindHandler = this._bindHandler.bind(this);
     this._editHandler = this._editHandler.bind(this);
+    this._updateRoute = this._updateRoute.bind(this);
   }
 
   componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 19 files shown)


msecret:
Requires styling here: https://github.com/18F/cg-style/pull/153

static_src/components/route.jsx
```diff
@@ -0,0 +1,178 @@
...
+    );
+  }
+
+  get actions() {
+    const actions = [];
+    const route = this.props.route;
+    if (!route) return actions;
+
+    if (!route.app_guid) actions.push(this.deleteAction);
+    if (route.app_guid) actions.push(this.editAction);
```
jcscottiii: use `else` / `else if` instead

static_src/components/route.jsx
```diff
@@ -0,0 +1,178 @@
...
+    actions.push(this.bindAction(!!route.app_guid));
+
+    return actions;
+  }
+
+  get confirmationMsg() {
+    return (this.props.route.app_guid) ? 'Unbind this route from this app?' :
```
jcscottiii: here, line 117, line 118 and line 119 all do this logic for checking the existence of `app_guid`.  you should create a helper function that does something like `isRouteBoundToApp(route)

static_src/components/route_form.jsx
```diff
@@ -93,31 +94,29 @@ export default class RouteForm extends React.Component {
...
-    );
-  }
-
-  // If there is a delete handler, this form is for editing a route
-  // Otherwise, it is for a new route
-  // The submit action label should be adjusted accordingly
   get submitActionText() {
-    if (this.props.deleteHandler) return 'Apply';
-    return 'Add route';
+    if (this.props.route.guid) return 'OK';
```
jcscottiii: also here could use the helper.

jcscottiii:
also for some of the codeclimate issues that are in tests files, you might want to add an exclude. we do it similarly here: https://github.com/opencontrol/compliance-masonry/blob/master/.codeclimate.yml#L10

PR #636, opened 2016-09-28
Data for instance binding error

msecret:
Getting the data flow done for showing errors when a service instance is bound on the app page.

### DIFF
--- a/static_src/stores/service_instance_store.js
+++ b/static_src/stores/service_instance_store.js
@@ -247,7 +247,7 @@ class ServiceInstanceStore extends BaseStore {
       case serviceActionTypes.SERVICE_INSTANCE_ERROR: {
         const instance = this.get(action.serviceInstanceGuid);
         if (!instance) break;
-        const newInstance = Object.assign({}, instance, {error: action.error});
+        const newInstance = Object.assign({}, instance, { error: action.error });
         this.merge('guid', newInstance, (changed) => {
           if (changed) this.emitChange();
         });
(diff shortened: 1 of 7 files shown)


static_src/test/unit/stores/service_instance_store.spec.js
```diff
@@ -603,4 +604,45 @@ describe('ServiceInstanceStore', function() {
...
+    it('should emit change if errored instance found', function() {
+      const instanceGuid = 'zvxcsdf34sint';
+      const instance = { guid: instanceGuid };
+      const err = testCFError;
+
+      ServiceInstanceStore.push(instance);
+
+      const spy = sandbox.spy(ServiceInstanceStore, 'emitChange');
+
+      serviceActions.instanceError('fake-guid', err);
```
msecret: It's to check that emitChange was only called on one of them, the first valid one, so 2nd call. So `expect(spy).toHaveBeenCalledOnce();`  is the only assertion.

static_src/stores/service_instance_store.js
```diff
@@ -217,7 +217,8 @@ class ServiceInstanceStore extends BaseStore {
         const instance = this.get(binding.service_instance_guid);
         if (!instance) break; // TODO throw error
         const updatedInstance = Object.assign({}, instance, {
-          changing: false
+          changing: false,
+          error: false
```
jcscottiii: hmm when i think of `error` i expect like a string or an object not a `boolean`. maybe change this to `hasError` or something similar.

msecret: I don't like hasError because then you have to check `hasError` and then look for a separate error object, which is two separate places data is kept and two separate checks.

Issue #601, opened 2016-09-19
Refine service states


suprenant:
As a user, I want a clear description of the current state of my services. 

Notes: 
 - Cloud Foundry API doesn't give us official state of stopped
 - API gives us the last operation (create, update, or delete) and whether or not the operation was successful and we determine a state from this
 - From this, we are currently displaying these states: Failed, deleting, processing, running, and stopped
 - Ref: https://apidocs.cloudfoundry.org/242/service_instances/retrieve_a_particular_service_instance.html
last_operation

suprenant:
@msecret @mogul – do you have anything to add/clarify here at the moment? This came up during our last sprint review, and Marco and I talked this over some.

mogul:
I'm having a hard time mentally mapping the combination of `last_operation.type` and `last_operation.state` to `failed`,`deleting`,`processing`,`running`,and `stopped`. Can you put in a table showing how those values are being mapped?

suprenant:
@msecret can you please add the table per Bret's request?

msecret:
|             | create  | update  | delete   | None    |
|-------------|---------|---------|----------|---------|
| **in progress** | Running | Running | Deleting | Stopped |
| **succeeded**   | Running | Running | Deleting | Stopped |
| **failed**      | Failed  | Failed  | Failed   | Stopped |

mogul:
Thanks! That's just what I was after. 

One confusing thing is `create+in-progress`... It says "Running" but the service not actually available to the app until it is `create+succeeded`! This _could_ be *very slow*... The API is asynchronous in case a service even actually has to be hand-provisioned! The same goes for `update`. `Running` can be misleading while things are in a waiting state.

Can I suggest:

|             | create  | update  | delete   | None    |
|-------------|---------|---------|----------|---------|
| **in progress** | Starting | Reconfiguring | Deleting | Stopped |
| **succeeded**   | Available | Available | - | Stopped |
| **failed**      | Failed  | Failed  | Failed   | Stopped |

msecret:
That sounds good Bret! Thanks for taking a look

msecret:
In the new UI, we're no longer showing all the states, we're only showing when something fails. Is that correct @thisisdano

thisisdano:
It probably makes sense to include the status for all of them except available.

Here are a couple options — a simple change to the status text, and a state with an activity animation:

[...]

PR #638, opened 2016-09-29
Rename process states based on discussion

msecret:
To make them clearer. Just expanded on if statement as it's easiest
to understand and shouldn't really get any more complicated then this.
Because logic is based on both type and state, using a map seems more
complicated to me.

Refs #601


msecret:
@suprenant can you take a look at this one?

suprenant:
@msecret will do!

Issue #637, opened 2016-09-29
Display action loading state for app page panels


msecret:
In order for a developer to know when a route or service is still in the process of being deleted, (un)bound, there should be a loading UI for actions.

Acceptance criteria:
- [ ] The loading UI appears for any action taking on an existing route or service on the app page panels.
- [ ] The UI should state what action is being taken, ie "Deleting" or "Binding".
- [ ] The UI should have an animated loading indicator.

msecret:
https://cloud.githubusercontent.com/assets/11464021/18878027/02768a02-8484-11e6-9634-484dbe3e7954.png

[design issue](https://github.com/18F/cg-design/issues/14)
[invision prototype](https://gsa.invisionapp.com/share/9T8GMJ8B8#/screens/192025036).

msecret:
Things required before development:

- @thisisdano can you provide me with a gif of the animated loading UI? The 3 alternating blue dots.
- @thisisdano and @vz3 is the design in https://github.com/18F/cg-design/issues/14 ready for development?

thisisdano:
@msecret You could do something like this in css for the loading animation: http://codepen.io/thisisdano/pen/jrLywZ

As for the second point, this was in the original issue:
> What if we don't know what the request is? Like if we don't know if it's "Starting", "Deleting", etc? We'll just show the loader? Trying to think of all possible scenarios now.

In what situations would we not know the kind of request? AFAIK this loader triggers only on a specific action, like `delete` or `bind`.

Other than that Q, I think it's ready to try.

vz3:
@thisisdano animation looks great @msecret thumbs up here

PR #642, opened 2016-10-04
Loading actions for routes and services

msecret:
Uses the `loading` property on the route and instances to display the new inline loader when an action for them is currently happening. So when binding or editing is being requested and waiting for response on server, will show loading like the UI.

Requires https://github.com/18F/cg-style/pull/161
Ref https://github.com/18F/cg-dashboard/issues/637


jcscottiii:
LGTM. feel free to merge once https://github.com/18F/cg-style/pull/161 is taken care of.

PR #663, opened 2016-10-07
Use inert to serve static assets

### LINKED ISSUE #633: Fix broken testing server
The testing server that uses the `smocks` library to create a cloud foundry API with fake data we specify is broken. This is likely due to the [code climate fixes pr](https://github.com/18F/cg-dashboard/pull/607). My guess is that I updated the `npm-shrinkwrap` file so that one of smocks dependencies was updated so it no longer works with the API we're using.

Potential remedies:
- [ ] Attempt to get `npm-shrinkwrap` back into a version state where `smocks` was working.
- [x] Fix our use of smocks so it works with the newer version. This would have the added benefit of fixing a few security problems which is blocking us from deploying a an actual server using this test data #577. Fixing smocks means adding a separate file handler as a plugin to `hapi`, the server module `smocks` uses.

adborden:
Fixes #633

smocks provides a [respondWithFile and respondWithFileHandler](http://jhudson8.github.io/fancydocs/index.html#project/jhudson8/smocks/method/Variant/respondWithFile) but it's not as flexible as we need it to be like handling `index.html` or figuring out mime types. Since we're really just trying to serve static assets, we should just use hapi's `inert` to do that.

It does add a bit of boilerplate to configure hapi, but overall feels like the right solution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "lint": "eslint -c .eslintrc ./static_src",
     "test": "npm run lint && karma start --single-run",
     "testing-server": "node ./static_src/test/server/server.js",
+    "watch-server": "npm run testing-server & npm run watch",
     "watch-test": "karma start --browsers Chrome",
     "watch": "webpack --watch"
   },
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "flux": "^2.0.3",
     "immutable": "^3.8.1",
     "imports-loader": "^0.6.4",
+    "inert": "^4.0.2",
     "jasmine-core": "^2.3.4",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -95,5 +96,8 @@
     "url-loader": "^0.5.7",
     "watch": "^0.16.0",
     "webpack": "^1.13.2"
+  },
+  "devDependencies": {
```
msecret: We;re currently keeping everything as regular dependencies, no dev deps (although I don't think I agree with that logic). Regardless, if we put this up to run on a server, shouldn't it be a regular dep?

adborden: Yep, good point, fixed.

Issue #655, opened 2016-10-06
Redefine page grid and max-width


msecret:
In order for dashboard users to have a unified experience with various cloud.gov sites, the grid and max-width should be changed to new design.

Acceptance criteria:
- [x] At max-width, the gutter should be 32px.
- [x] At max-width, the column should be 48px.
- [x] The max-width should be 928px
- [x] The gutters and columns should be responsive, therefore should be smaller px sizes when the browser window is smaller then the max width.

PR #665, opened 2016-10-12
Use new cg-grid

adborden:
Fixes https://github.com/18F/cg-dashboard/issues/655

- [x] Depends on https://github.com/18F/cg-style/pull/165
- [x] Depends on https://github.com/18F/cg-style/pull/167

@msecret have some questions, would like to pair with you on this.

screenshot from 2016-10-12 13-25-18

### DIFF
--- a/static_src/components/disclaimer.jsx
+++ b/static_src/components/disclaimer.jsx
@@ -17,7 +17,7 @@ export default class Disclaimer extends React.Component {
     const flagAlt = 'US flag signifying that this is a United States Federal Government website';
     return (
       <div className={ this.styler('usa-disclaimer') }>
-        <div className={ this.styler('usa-grid') }>
+        <div className={ this.styler('cg-grid') }>
           <span className={ this.styler('usa-disclaimer-official') }>
             An official website of the United States Government
             <img alt={ flagAlt } src={ flag }>
--- a/static_src/components/disclaimer.jsx
+++ b/static_src/components/disclaimer.jsx
@@ -17,7 +17,7 @@ export default class Disclaimer extends React.Component {
     const flagAlt = 'US flag signifying that this is a United States Federal Government website';
     return (
       <div className={ this.styler('usa-disclaimer') }>
-        <div className={ this.styler('cg-grid') }>
+        <div className={ this.styler('grid') }>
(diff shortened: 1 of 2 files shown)


adborden:
@thisisdano can you take a look?

**< 600px**
screenshot from 2016-10-13 16-36-50

**1600px**
screenshot from 2016-10-13 16-36-11

thisisdano:
@adborden That thick dark line should be gone after the changes in https://github.com/18F/cg-style/pull/169

THAT SAID — the dashboard is not really designed to be responsive. (And I don't think it's a realistic goal for this PI, but I'd be happy to be proven wrong...) I'm inclined to remove the viewport meta info from the dashboard header. It'd suck more to think it's responsive and find out it actually isn't.

thisisdano:
@adborden FYI, thinking this through now. More to come.

adborden:
Yeah, I think that makes sense. We can make the column a 15 column grid, and assume the main content will have 12 columns.

adborden:
@thisisdano I spent a little bit of time trying to expand the grid from 12-15 columns, but it turns out that it gets bit tricky. Even though we don't explicitly support a responsive design, the current implementation hides the sidenav on small screens at a non-standard breakpoint. This makes things kind of annoying, because the grid switches from 15 to 12 column grid at some breakpoint between medium and large and we need to account for that.

In the interim, while we still have the sidenav, I suggest we just treat it as being outside the grid and design/build toward a 12-column grid. This does mean that things outside the main content, like disclaimer and header will not align to the same grid as the main content when the sidenav is present. On smaller screens, it _does_ align since the sidenav isn't present. I think this is an okay compromise for now. What do you think?

thisisdano:
@adborden @msecret The 12-column main content area is where we'll be focussing 99% of our design energy for the foreseeable future, so I'm comfortable with this decision.

If the misalignment in disclaimer and header gets too too irritating, we can always override at some point.

That said, we will want to address the broken responsive crappiness at some point as well!

adborden:
Thanks @thisisdano  where does that leave this PR, are there any changes you're requesting?

thisisdano:
@adborden Sorry. I wasn't clear! I think this PR is good to go as is — we will adjust any grid/alignment/responsive stuff later when and if we see issues.

PR #674, opened 2016-10-15
Suggestion for updating navigation link names and links

brittag:
This is a suggestion for a few navigation link changes, to help the Dashboard sync up with the other sections of cloud.gov. It'd be ok for the Dashboard crew to decide you all don't want to integrate some of these (or do this a different way) - just putting this up as a proposal for one way to do this.
- I updated the header navigation "About" link (https://cloud.gov/#about) to "Overview" (https://cloud.gov/overview/), to match the current homepage style. (Also filed a PR for cg-docs for this - https://github.com/18F/cg-docs/pull/466)
- That means this page no longer has an easy link to the cloud.gov homepage, so I updated the cloud.gov logo to link to the cloud.gov homepage. This is a bit of a tricky decision. I remember @thisisdano has insight into this kind of decision - tagging him here for his thoughts. :)
- Now we have two links on this page called "Overview", since the Dashboard-local homepage is called "Overview" in the Dashboard sidebar. It's best to not have two things named the same thing, so I renamed the Dashboard-local homepage to "Guide". I'm not sure if that's a better name. As a team we could also discuss the name of the new "Overview" section some more - see https://github.com/18F/cg-docs/issues/332 where we had that discussion. (We also had the option of "Introduction", for example.)
- Since I renamed the "Overview" link, I also renamed it in the testing steps.

cc @jameshupp @berndverst for copy + customer-help thoughts as well.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -87,7 +87,7 @@ and the `testSpace01` space.
   - Click edit again.
   - Click "Delete route".
   - Ensure the route is deleted.
-1. Click on the `Overview` link in the sidenav and ensure the homepage appears.
+1. Click on the `Guide` link in the sidenav and ensure the homepage appears.
 
 ## Tagging
 
--- a/static_src/components/header.jsx
+++ b/static_src/components/header.jsx
@@ -25,7 +25,7 @@ export default class Header extends React.Component {
     <header className={ this.styler('header') }>
       <div className={ this.styler('header-wrap') }>
         <div className={ this.styler('header-title') }>
-          <a href="/" className={ this.styler('logo') } title="Home">
+          <a href="https://cloud.gov/" className={ this.styler('logo') } title="Home">
             <svg className={ this.styler('logo') }>
               <use
                 xlinkHref={ this.getImagePath('logo') }>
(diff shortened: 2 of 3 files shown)


thisisdano:
So, we have time scheduled in Sprint 2 to give this its proper consideration, but I'd like to toss my 2 cents in early. Here're a few of my general beliefs:

- The dashboard need not (and perhaps _should_ not) share the same navigation as the marketing pages. This is a common web app convention — signing in to the app reveals a new navigation structure — since user needs and expectations have a state change at that threshold.
- The expectation state-change threshold is signin.
- Documentation pages straddle the threshold (important both to signed in and non–signed in)
 - Implementation of docs nav varies in the market
 - Heroku and Harvest have a subdomain for docs/help with its own nav/UI
 - Dropbox has docs ('Help') as a path with its own UI
 - Each of these use its top-left link for docs/help home
- The top-left link (typically the logo) in the web app should link to the web app home, not the marketing home.
- Once signed in, the link to the marketing home could appear in a spot other than the primary (top-left) home link — like in the footer, for instance, as it does in Heroku.
- `Overview` is a great word/concept — which is why we have also been using it to describe the web app home. It should not describe both this view and the section of the marketing site that gives an overview of the platform. I have opinions, but I won't make any further suggestions at this point!

brittag:
I also just filed a PR at https://github.com/18F/cg-site/pull/549 to switch the "Contact" link to lead to the Help page. If that gets merged, this would need to be updated in the Dashboard nav as well.

brittag:
Also related: https://github.com/18F/cg-site/pull/552 - this adds a search box to the nav for the landing and updates pages.

suprenant:
Noted that we need to do some review on the squad here, which will inform the next steps for this PR: https://github.com/18F/cg-dashboard/issues/712

brittag:
Hooray, this PR is no longer helpful since the header is different now! The header might benefit from a "contact" link, to make it easier for dashboard users to find help if they need it, but I can discuss/file that elsewhere. :)

PR #685, opened 2016-10-20
Updates to style and sidenav

thisisdano:
The template companion to 18F/cg-style#169

### DIFF
--- a/static_src/components/navbar.jsx
+++ b/static_src/components/navbar.jsx
@@ -82,8 +82,8 @@ export class Nav extends React.Component {
 
   render() {
     const mainList = this.styler('usa-sidenav-list', 'sidenav-list', 'sidenav-level-one');
-    const secondList = this.styler('usa-sidenav-list', 'sidenav-list', 'sidenav-level-two');
-    const thirdList = this.styler('sidenav-list', 'usa-sidenav-sub_list', 'sidenav-level-three');
+    const secondList = this.styler('usa-sidenav-sub_list', 'sidenav-list', 'sidenav-level-two');
+    const thirdList = this.styler('sidenav-list', 'sidenav-level-three');
     const downArrow = this.styler('menu-arrow', 'sidenav-arrow', 'sidenav-arrow-down');
     const rightArrow = this.styler('menu-arrow', 'sidenav-arrow', 'sidenav-arrow-right');
     const header = this.styler('sidenav-header');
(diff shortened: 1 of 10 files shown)


static_src/components/navbar.jsx
```diff
@@ -112,7 +112,7 @@ export class Nav extends React.Component {
             subList = (
               <ul className={ secondList }>
                 <li>
-                  <a href={ this.orgHref(org) }>{ org.name }</a>
+                  <a href={ this.orgHref(org) }>{ org.name } overview</a>
```
thisisdano: previously "spaces" — but it's actually our rudimentary org page

**note:** this element should get the active class `sidenav-active-dashboard` any time its page is visible. I didn't know how to do this.

static_src/components/navbar.jsx
```diff
@@ -129,7 +128,7 @@ export class Nav extends React.Component {
                   </ul>
                 </li>
                 <li className={ this.styler('marketplace') }>
-                  <a href={ this.marketplaceHref(org) }>Marketplace</a>
+                  <a href={ this.marketplaceHref(org) }>{ org.name } marketplace</a>
```
thisisdano: **note:** this element should get the active class `sidenav-active-dashboard` any time its page is visible

static_src/components/navbar.jsx
```diff
@@ -82,17 +82,19 @@ export class Nav extends React.Component {
...
   render() {
     const mainList = this.styler('usa-sidenav-list', 'sidenav-list', 'sidenav-level-one');
-    const secondList = this.styler('usa-sidenav-list', 'sidenav-list', 'sidenav-level-two');
-    const thirdList = this.styler('sidenav-list', 'usa-sidenav-sub_list', 'sidenav-level-three');
+    const secondList = this.styler('usa-sidenav-sub_list', 'sidenav-list', 'sidenav-level-two');
+    const thirdList = this.styler('sidenav-list', 'sidenav-level-three');
     const downArrow = this.styler('menu-arrow', 'sidenav-arrow', 'sidenav-arrow-down');
     const rightArrow = this.styler('menu-arrow', 'sidenav-arrow', 'sidenav-arrow-right');
     const header = this.styler('sidenav-header');
+    const faHome = this.styler('fa', 'fa-home');
```
msecret: I really don't want to bring font awesome into the dashboard. We're only using it on docs right now and would prefer to remove if from there rather then add it to more places. Is that what we're attempting to do here with the `fa` class?

PR #697, opened 2016-11-01
Update cf-deck to cg-dashboard

adborden:


adborden:
Orgs? No, I don't think there's any action on the cloud foundry side. The apps running today are already running with these names (although I'm not 100% sure why that is).

Issue #630, opened 2016-09-27
Move common store functionality to shared base store


msecret:
In order to speed up developers when they're creating similar store interactions to what's already there, there should be reusable code and patterns for common store functionality.

msecret:
This could include:
- Adding/removing a bool property on store object (like route store removing)
- Toggle edit/delete actions
- Waiting on multiple entities (waitingOnRequests)

msecret:
@adborden

msecret:
```js
// route store
case routeActionTypes.ROUTE_TOGGLE_REMOVE: {
  this.toggleOption(action.route, 'remove');
  break
}

// base store
toggleRemove(entity, option) {
  const newEntity = Object.assign({}, entity, {
    [option]: !entity[option]
  });
  this.merge('guid', newEntity, (changed) => {
    if (changed) this.emitChange();
  });
}
```

PR #720, opened 2016-11-08
Refactor for loading status

adborden:
Resolves https://github.com/18F/cg-dashboard/issues/630

This introduces a semaphore for tracking in-flight requests built into the base store. If you want the requests tracked, use `this.load()` in your store. As requests are made, we increment a counter. When those requests resolve (or error), we decrement the count.

We combine `fetched` and `fetching` by keeping track of initialization. The loading status is considered loading until the first batch of requests are made _and_ completed.

### Future improvements

There's still some room for improvement. The `ActivityStore` is tracking two datasets to determine loading status. It can be moved to this load strategy, but requires some extra work.

This also introduces some extra change events but I believe react/redux is smart enough to handle that without much performance impact? The reason for this is that `LoadingStatus` hooks into the requests via promises and these happen async outside of the receive action, within the promise resolve.

### DIFF
--- a/static_src/test/unit/stores/base_store.spec.js
+++ b/static_src/test/unit/stores/base_store.spec.js
@@ -199,60 +199,6 @@ describe('BaseStore', () => {
     });
   });
 
-  describe('fetching', function() {
-    it('should initially be false', function() {
-      expect(store.fetching).toEqual(false);
-    });
-
-    it('should be able to be set to true', function () {
-      store.fetching = true;
-      expect(store.fetching).toEqual(true);
-    });
-
-    it('should cast all values to booleans', function () {
-      store.fetching = 'true';
-      expect(store.fetching).toEqual(true);
-
-      store.fetching = undefined;
-      expect(store.fetching).toEqual(false);
-    });
-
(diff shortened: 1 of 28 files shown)


static_src/test/unit/stores/base_store.spec.js
```diff
@@ -387,4 +333,54 @@ describe('BaseStore', () => {
       });
     });
   });
+
+  describe('load()', function () {
+    describe('when called', function () {
+      let spy;
+
+      beforeEach(function (done) {
+        spy = sinon.spy();
```
msecret: Is there a reason you're not using the sinon sandbox?

static_src/test/unit/stores/base_store.spec.js
```diff
@@ -387,4 +333,54 @@ describe('BaseStore', () => {
...
+        store.on('CHANGE', spy);
+
+        // unresolved promise
+        const promise = new Promise(function () { });
+
+        store.load([promise]);
+        setImmediate(done);
```
msecret: This is cool, I'm sad I didn't know about `setImmediate`

PR #721, opened 2016-11-09
Usage and limits readonly

### LINKED ISSUE #708: Adjust hierarchy of App page elements based on wireframe
What we're after:
 - Does the hierarchy of the page reflect the hierarchy of the wireframe
 - Does this match?
 - Don’t want to imply that implementing the wireframe is not implementing all of the functionality

msecret:
Implements readonly view of usage and limits, as well as implementing new sidenav still exists grid system which is defined in the screenshot posted that Dan sent me.

Relies on [cg-style#ms-statts](https://github.com/18F/cg-style/tree/ms-statts)

Also attempts to resolve #708, as the PR required a lot of layout work on the app page.

- Renames app_settings_panel to following naming conventions in design file.
- Creates new panel_block component which would be used similarly to panel_row expect it doesn't flexbox the children into columns (used for the strange scale UI).
- Added ability to define columns for panel group. Decided this way as it worked easiest for this implementation. I think we should change this when we use the grid more, but it's hard to know how we want to use it right now.
- The resource usage component will take care of state for the stat, and will also take care of forms when we build modifying the limits in.
- Most grid stuff is done though overrides in dashboard, as we probably want to wait till we do a more full nav change to make big grid updates.

TODO:
- Scale needs gray box around it.

@thisisdano what do you want to do about the "To start or stop an app, follow..." messaging thats there?

How it looks
screen shot 2016-11-08 at 4 43 39 pm

The grid system (using 2nd one as still have a sidebar)
-column-layout


static_src/components/stat.jsx
```diff
@@ -0,0 +1,52 @@
...
+const STATES = [
+  'error',
+  'warning',
+  'info',
+  'success',
+  'none'
+];
+
+const propTypes = {
+  primaryStat: React.PropTypes.string.isRequired,
```
msecret: Made these strings, so it's a pretty dumb layout component. This means ResourceUsage should take care of the state and such.

static_src/components/stat.jsx
```diff
@@ -0,0 +1,52 @@
...
+  statState: React.PropTypes.oneOf(STATES),
+  secondaryInfo: React.PropTypes.element
```
msecret: Does this need the customizability to be an element? Or can it just be a string?

I noticed in the UI, it's color should change along with that statState, but it should only change when the stat is red or is being edited. Should that more complex logic be held in the ResourceUsage component or here? That's kinda the decision to be an element or text.

static_src/components/resource_usage.jsx
```diff
@@ -0,0 +1,71 @@
...
+};
+
+export default class ResourceUsage extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+    this.styler = createStyler(style);
+  }
+
+  // TODO move to util
```
msecret: Does it really need to be though?

static_src/components/panel_group.jsx
```diff
@@ -13,10 +20,14 @@ export default class PanelGroup extends React.Component {
   }
 
   render() {
+    const gridClass = `grid-width-${this.props.columns}`;
```
msecret: Should this be changed like panel_block and default to 0 when columns not passed in, so you don't set any grid width class when you don't pass in columns number?

static_src/components/panel_block.jsx
```diff
@@ -0,0 +1,38 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {};
+    this.styler = createStyler(style);
+  }
+
+  render() {
+    let gridClass = '';
+    if (this.props.columns !== 0) {
+      gridClass = `grid-width-${this.props.columns}`;
```
msecret: Not sure this is needed

static_src/components/usage_and_limits.jsx
```diff
@@ -20,77 +22,86 @@ export default class UsageAndLimits extends React.Component {
...
+    </div>
+    );
+  }
+
+  get totalMemory() {
+    return (
+    <div>
+      <ResourceUsage title="Total memory"
+        amountUsed={ this.getStat('mem') * this.props.app.running_instances }
+        amountTotal={ this.props.quota.memory_limit * 100000 }
```
adborden: Hmm, well that number is wrong, `100000 = 100 * 1000`, that's 100KB units, not 1MB units. Instead of a constant or util, I usually just write these out as factors, like `1000 * 1000`. But if it's used in enough places, a constant seems more appropriate than a util function.

static_src/components/usage_and_limits.jsx
```diff
@@ -20,77 +22,86 @@ export default class UsageAndLimits extends React.Component {
...
-        </h5>
-        <span className={ this.styler('panel-column') }>
-          { this.getStat('disk_quota') } MB
-        </span>
-        <span className={ this.styler('panel-column') }
-            style={{ textAlign: 'left' }}>
-          Using { this.getStat('disk') }  MB
-          { helpText }
-        </span>
+    <div style={{ marginTop: '1rem' }}>
```
adborden: What's the thinking behind one-off styling vs baking it into a class?

static_src/components/usage_and_limits.jsx
```diff
@@ -20,77 +22,86 @@ export default class UsageAndLimits extends React.Component {
...
+  get scale() {
+    return (
+      <div style={{ textAlign: 'center' }}>
+        <h5>App scale</h5>
+        <span className={ this.styler('stat-primary')}>
+          { this.props.app.running_instances }X
         </span>
-        <span className={ this.styler('panel-column') }></span>
+        <br />
+        <h5 style={{ fontSize: '0.75rem', maxWidth: '5rem', display: 'inline-block', marginTop: '3rem' }}>
```
adborden: Even if it's not being used anywhere else, putting it in `cg-style` allows things like linting, use of constant variables (like padding or common font sizes), makes search/replace easier.

static_src/components/usage_and_limits.jsx
msecret: It should probably be a class, just had trouble figuring out where it belonged

PR #725, opened 2016-11-15
Restart app, data part

msecret:

PR #729, opened 2016-11-15
Restart app, ui part

msecret:
Built on top of https://github.com/18F/cg-dashboard/pull/725

### DIFF
--- a/static_src/test/unit/stores/app_store.spec.js
+++ b/static_src/test/unit/stores/app_store.spec.js
@@ -262,12 +262,19 @@ describe('AppStore', function() {
   });
 
   describe('on app restarted', function() {
-    it('should emit a change', function() {
-      const spy = sandbox.spy(AppStore, 'emitChange');
+    it('should poll until running instances is greater then 0', function() {
+      const spy = sandbox.spy(AppStore, 'poll');
+      const expectedRes = {
+        data: { running_instances: 1 }
+      };
 
       appActions.restarted('zxlcvkjklv');
 
       expect(spy).toHaveBeenCalledOnce();
+      let args = spy.getCall(0).args;
+      expect(args[0]).toBeFunction();
+      expect(args[0](expectedRes)).toBeTruthy();
+      expect(args[1]).toBeFunction();
     });
   });
 });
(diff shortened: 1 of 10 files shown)


static_src/util/cf_api.js
```diff
@@ -237,6 +239,10 @@ export default {
                           appActions.receivedApp);
   },
 
+  fetchAppStatus(appGuid) {
```
msecret: I created a new method because the fetch app method will call the app received action which will update the UI which I don't want to have happen. I really just want a promise that gets the data and returns it. But there might be another way to re-use the existing fetchApp method rather then creating another one like this.

static_src/test/unit/util/poll.spec.js
```diff
@@ -0,0 +1,46 @@
...
+
+    setImmediate(function() {
+      expect(value).toEqual(true);
+      done();
+    });
+  });
+
+  it('should re-send the promise if the condition is false', function() {
+    // TODO not sure how to test
+  });
```
adborden: Yeah, this is pretty tricky, you could do something do something like you have above and use [sinon's fake timers](http://sinonjs.org/docs/#clock) to tick the clock forward to trigger the poll. Here's some pseudo code:

```javascript
let lastCapturedPromise;
let clock = sinon.useFakeTimers();
function capturedPromiseFactory () { // return a promise, save the resolve/reject to lastCapturedPromise }

it(funciton() {
  pollPromise = poll(identity, capturedPromiseFactory);
  lastCapturedPromise.resolve(false);
  clock.tick(1);
  // poll runs again, lastCapturedPromise is updated
  // assert pollPromise not resolved
  lastCapturedPromise.resolve(true)
  clock.tick(1)
  // assert pollPromise is resolved
});
```

I think the promises resolve async, so that adds more complexity because you have to do your asserts async as well :(

static_src/test/unit/stores/app_store.spec.js
```diff
@@ -262,12 +262,19 @@ describe('AppStore', function() {
   });
 
   describe('on app restarted', function() {
-    it('should emit a change', function() {
-      const spy = sandbox.spy(AppStore, 'emitChange');
+    it('should poll until running instances is greater then 0', function() {
```
adborden: Should this fire an action on resolve? Maybe the app received action shouldn't emit a change?

adborden: Hmm, the comment about firing an action on resolve was referring to this line, https://github.com/18F/cg-dashboard/pull/729/files#diff-d45ddca2d392b7558d8957b40dd036ccR242 I'm not sure how it got up here. anyway, sounds good to me.

PR #733, opened 2016-11-17
Set HttpOnly & Secure flags on session cookies; Remove Session Key from code

cnelson:
[FedRamp requires us](https://github.com/18F/cg-produc/issues/470) to set cookies with the [secure](https://github.com/18F/cg-product/issues/443) and [HttpOnly](https://github.com/18F/cg-product/issues/442) flags where appropriate.

This PR, sets the `HttpOnly` flag to true for all session cookies, and sets the `secure` flag to true based on the value of the `SECURE_COOKIES` environment variable.

While I was looking at the session handling code I noticed the key for securing sessions is not stored in an environment variable, so this PR also includes a change to move that secret into the `SESSION_KEY` envvar / user-provided-service.

cc: https://github.com/18F/cg-product/issues/470

### DIFF
--- a/server.go
+++ b/server.go
@@ -31,6 +31,12 @@ func loadEnvVars() helpers.EnvVars {
 	envVars.PProfEnabled = os.Getenv(helpers.PProfEnabledEnvVar)
 	envVars.BuildInfo = os.Getenv(helpers.BuildInfoEnvVar)
 	envVars.NewRelicLicense = os.Getenv(helpers.NewRelicLicenseEnvVar)
+	envVars.SecureCookies = os.Getenv(helpers.SecureCookiesEnvVar)
+	envVars.SessionKey = os.Getenv(helpers.SessionKeyEnvVar)
+	// set a default session key if one isn't provided
+	if envVars.SessionKey == "" {
+		envVars.SessionKey = "some key"
+	}
 	return envVars
 }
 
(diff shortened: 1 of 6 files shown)


server.go
```diff
@@ -31,6 +31,12 @@ func loadEnvVars() helpers.EnvVars {
 	envVars.PProfEnabled = os.Getenv(helpers.PProfEnabledEnvVar)
 	envVars.BuildInfo = os.Getenv(helpers.BuildInfoEnvVar)
 	envVars.NewRelicLicense = os.Getenv(helpers.NewRelicLicenseEnvVar)
+	envVars.SecureCookies = os.Getenv(helpers.SecureCookiesEnvVar)
+	envVars.SessionKey = os.Getenv(helpers.SessionKeyEnvVar)
+	// set a default session key if one isn't provided
+	if envVars.SessionKey == "" {
```
msecret: `Getenv` will default to `''` if var isn't set correct?

cnelson: [Getenv retrieves the value of the environment variable named by the key. It returns the value, which will be empty if the variable is not present.](https://golang.org/pkg/os/#Getenv)

cnelson:
WDYT @adborden ?

server.go
adborden: Actually, can we just throw here and require the SessionKey? I'd rather not have it too easy to misconfigure the environment variables and accidentally use a weak key without ever knowing.

Issue #743, opened 2016-11-17
Dashboard build process doesn't correctly display JSX errors


msecret:
Background: A while ago, we had to change our build because our tests were not stopping when there was a syntax error in the test file. Instead, the test file would be skipped over, so the developer could commit code to production that had broken tests. Unfortunately this change introduce a  new bug where the build won't accurately or legibly display syntax errors in JSX and other files.

An example of error being displayed incorrectly:
```
Error: "BabelLoaderError: SyntaxError: Unexpected token (26:2)\n\n\u001b[0m  24 |   active\u001b[1m:\u001b[22m \u001b[36mtrue\u001b[39m\u001b[1m,\u001b[22m\n  25 |   loadingDelayMS\u001b[1m:\u001b[22m LOADING_TIME\n> 26 |   style\u001b[1m:\u001b[22m \u001b[31m'cover'\u001b[39m\n     |   ^\n  27 | \u001b[32m}\u001b[39m\u001b[1m;\u001b[22m\n  28 | \n  29 | \u001b[36mclass\u001b[39m Loading \u001b[36mextends\u001b[39m React\u001b[1m.\u001b[22mComponent \u001b[32m{\u001b[39m\u001b[0m\n"

```

Example of how same error should look:
```
BabelLoaderError: SyntaxError: Unexpected token (26:2)

  24 |   active: true,
  25 |   loadingDelayMS: LOADING_TIME
> 26 |   style: 'cover'
     |   ^
  27 | };
  28 |
  29 | class Loading extends React.Component {
```

Fix: 
A. Have configuration to only turn on the feature that stops tests with a syntax error on single test runs OR test runs that run on a CI server. This ensures that broken test files will never be committed because will be caught at CI level while devs still have easier debugging during local development. Less risky option, easier to configure.
B. Attempt to fix build so tests stop on syntax errors AND errors display correctly. I currently don't know how this would be done. Riskier option because more unknown.

suprenant:
Discussed, thinking Fix A is way to go.

adborden:
Confirmed fix.

PR #746, opened 2016-11-17
Bug jsx build

msecret:
The karma warnings plugin is great at ensuring syntax errors in tests
don't stop the test file from being tested at all, but creates really
bad error messages during the build. This fix would make that plugin
only run on CI, when we really need to know if tests aren't being run
due to syntax errors.

Ref #743

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,7 +6,6 @@ machine:
     CF_ORGANIZATION: "cf"
     CF_API_GC: "https://api.fr.cloud.gov"
     CF_ORGANIZATION_GC: "cloud-gov"
-    NODE_ENV: "prod"
   post:
     - cd cg-dashboard && nvm install && nvm use && nvm alias default $(nvm current)
     - mkdir -p download
@@ -39,7 +38,7 @@ test:
     - if ! go get github.com/golang/tools/cmd/cover; then go get golang.org/x/tools/cmd/cover; fi
   override:
     - cd $WS && ./codecheck.sh -u
-    - cd $WS && npm test
+    - NODE_ENV="test" cd $WS && npm test
 
 deployment:
   deploy:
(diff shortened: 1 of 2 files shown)


msecret:
I tested this by purposely committing a syntax error. Circle did stop the build as expected, but I still was able to see accurate errors locally:

https://circleci.com/gh/18F/cg-dashboard/556

adborden:
Why do we even need the `WebpackKarmaWarningsPlugin`? If we leave it out, will we also get better message formatting in circle?

msecret:
@adborden if we leave it out, and there's a syntax error in a test file, it will continue running the tests, skipping the broken file completely.

adborden:
Hmm, I see. Instead of hooking into `NODE_ENV==production` for tests which seems weird to me, can we create a new condition around `NODE_ENV==test`? That's a more common convention I've seen.

adborden:
Or if it's really specific to Circle/CI we can use the `CI` env variable https://circleci.com/docs/environment-variables/#basics

msecret:
I changed it to use "test"

circle.yml
```diff
@@ -37,7 +37,9 @@ test:
     - if ! go get github.com/golang/tools/cmd/cover; then go get golang.org/x/tools/cmd/cover; fi
   override:
     - cd $WS && ./codecheck.sh -u
-    - cd $WS && npm test
+    - cd $WS && npm test:
+        environment:
+          NODE_ENV: "test"
```
adborden: yaml fanciness, how does this look as json? Mind if we write this instead:

```yaml
- cd $WS && NODE_ENV=test npm test
```

adborden:
Quoting Marco Segreto (2016-11-23 10:38:41)
> 
> Sure, I was hoping I could apply the environment to the whole `test` block, but couldn't anyway, so that's fine.

Hmm, yeah that would be nice. One option would be creating a sh script, or
an npm script.

--
Aaron D Borden
Human and Hacker

PR #748, opened 2016-11-18
Restart app error handling

msecret:
Read commit messages for info.

### DIFF
--- a/static_src/util/cf_api.js
+++ b/static_src/util/cf_api.js
@@ -261,7 +261,7 @@ export default {
   postAppRestart(appGuid) {
     return http.post(`${APIV}/apps/${appGuid}/restage`).then(() => {
       appActions.restarted(appGuid);
-    });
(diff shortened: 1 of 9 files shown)


static_src/util/cf_api.js
```diff
@@ -262,7 +262,7 @@ export default {
   postAppRestart(appGuid) {
     return http.post(`${APIV}/apps/${appGuid}/restage`).then(() => {
       appActions.restarted(appGuid);
-    });
+    }).catch((err) => handleError(err, appActions.error.bind(this, appGuid)));
```
adborden: why the `handleError`? if we're handling the error via an action, doesn't it mean we're handling the error twice?

msecret: The `handlerError` function takes care of some data handling in all error cases: https://github.com/18F/cg-dashboard/pull/748/files/3c8e204600f8c1ddd92c2b848f9f1170f59899a4#diff-d45ddca2d392b7558d8957b40dd036ccL19.

PR #752, opened 2016-11-20
Add CSRF protection

jmcarp:
Based on #733.

### DIFF
--- a/controllers/root.go
+++ b/controllers/root.go
@@ -2,16 +2,30 @@ package controllers
 
 import (
 	"fmt"
+	"html/template"
 	"net/http"
+	"os"
+	"path/filepath"
 
 	"github.com/18F/cg-dashboard/helpers"
 	"github.com/gocraft/web"
+	"github.com/gorilla/csrf"
 	"golang.org/x/oauth2"
 )
 
 // Context represents the context for all requests that do not need authentication.
 type Context struct {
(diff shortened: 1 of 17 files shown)


controllers/root.go
```diff
@@ -14,6 +18,16 @@ type Context struct {
 	Settings *helpers.Settings
 }
 
+var basePath = os.Getenv("BASE_PATH")
+var templates = template.Must(template.ParseFiles(filepath.Join(basePath, "static/index.html")))
```
jcscottiii: filepath.Join(basePath, "static", "index.html")

controllers/root.go
```diff
@@ -14,6 +18,16 @@ type Context struct {
 	Settings *helpers.Settings
 }
 
+var basePath = os.Getenv("BASE_PATH")
```
jcscottiii: could we put this in the EnvVars struct?

controllers/root.go
jcscottiii: also, if we do the above and put the extra env var, when we run `InitApp`, we can add a member variable to `Context` named `templates` and when we initialize Context, we can assign the template there instead of here as a global variable to controller package

jmcarp:
@jcscottiii: thanks for reviewing, I'll revise in a few.

PR #756, opened 2016-11-21
Remove default Session Key

cnelson:
@adborden It looks like https://github.com/18F/cg-dashboard/pull/733 got merged without addressing your feedback.  This PR removes the default `SESSION_KEY` and the server will fail to start if one is not provided.

At the risk of stating the obvious: Don't merge/deploy this until your user provided service in production has been updated to add a session key.


adborden:
This also means we're going to invalidate everyone's session. Should be fine, but I'll just test that those with cookies can at least log back in without issue.

adborden:
NM, I thought we were encrypting the session cookie, not the case.

adborden:
session keys set in e/w and govcloud.

PR #774, opened 2016-11-30
Overview with org quick looks

msecret:
Relies on https://github.com/18F/cg-style/pull/195, https://github.com/18F/cg-style/pull/196, https://github.com/18F/cg-style/pull/197


msecret:
What it looks like so far

screen shot 2016-11-29 at 5 45 08 pm

msecret:
Latest update

screen shot 2016-11-30 at 3 23 22 pm

static_src/components/app_count_status.jsx
```diff
@@ -0,0 +1,62 @@
...
+    this.styler = createStyler(style);
+  }
+
+  worstStatus(apps) {
+    return apps.reduce((previousState, app) => {
+      if (rankWorseState(previousState) < rankWorseState(app.state)) {
+        return app.state;
+      }
+      return previousState;
+    }, 'NONE');
```
adborden: I feel like returning `null` here is more appropriate.

static_src/components/app_count_status.jsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  render() {
+    const props = this.props;
+    let status = 'NONE';
```
adborden: I think `NONE` should be in `appStates`, even if it is an invalid state.

adborden:
Have some suggestions, but looks great :+1:

PR #792, opened 2016-12-05
Check for outdated dependences with gemnasium.

jmcarp:
eom

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,11 @@
 
 [![CircleCI](https://circleci.com/gh/18F/cg-dashboard.svg?style=svg)](https://circleci.com/gh/18F/cg-dashboard)
 [![Code Climate](https://codeclimate.com/github/18F/cg-dashboard/badges/gpa.svg)](https://codeclimate.com/github/18F/cg-dashboard)
+[![Dependency Status](https://gemnasium.com/18F/cg-dashboard.svg)](https://gemnasium.com/18F/cg-dashboard)
 
 Environments: [Production](https://dashboard.cloud.gov)
 [Master](https://dashboard-master.apps.cloud.gov)
 [Staging](https://dashboard-staging.apps.cloud.gov)
-
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


adborden:
I have used gemnasium before but code climate is new to me. @msecret I think you're right, that gemnasium only adds the ability to see outdated deps and is largely duplicative of what code climate gives us.

One question though, @msecret are you getting notifications for [these code climate issues](https://codeclimate.com/github/18F/cg-dashboard/issues)? I like the constant notifications that gemnasium sends about security issues, so I wouldn't want to miss that with code climate, but looks like we might be missing them.

msecret:
I'm getting notifications when the build fails, but maybe it's not as constant as what you want?

I think many of the code climate issues are somewhat irrelevant, as their in packages that are dev dependencies, so don't really have a security impact on production. Maybe a solution there woudl be to separate our deps and dev deps and ask code climate to test on dev deps? No idea if that's possible.

adborden:
Quoting Marco Segreto (2016-12-06 16:46:05)
> I'm getting notifications when the build fails, but maybe it's not as constant as what you want?
> 
> I think many of the code climate issues are somewhat irrelevant, as their in packages that are dev dependencies, so don't really have a security impact on production. Maybe a solution there woudl be to separate our deps and dev deps and ask code climate to test on dev deps? No idea if that's possible.

I see a little button on the issues to disable the check, let's try
that. Gemnasium also had that where you could "acknowledge" an issue so
it would stop bothering you.

When the node security project identifies a new issue, will Code
Climate notify you then?

--
Aaron D Borden
Human and Hacker

msecret:
Yea in Code Climate, in order to disable a specific check, you have to commit code, which can be a little annoying (but probably good to for record keeping)

adborden:
@jmcarp are we missing anything? I'm inclined to close this since it feels like it only duplicates what we already get with Code Climate.

msecret:
Im' going to close this as haven't heard anything in a while, will keep branch open.

PR #793, opened 2016-12-05
Allow for slightly different content for different deployments

tessereth:
Based on the conversation from https://github.com/AusDTO/cg-dashboard/pull/2, this provides the ability to make small modifications to content.

I detailed a procedure for managing these files [on the other PR](https://github.com/AusDTO/cg-dashboard/pull/2#issuecomment-264345170). Is that something that should be added to the repo somewhere or am I over thinking it?

PS. I renamed CG_SKIN to CF_SKIN as it seemed more appropriate.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,4 @@
 machine:
-  node:
-    version: 6.7.0
   environment:
     GODIST: "go1.6.linux-amd64.tar.gz"
     WS: "/home/ubuntu/.go_workspace/src/github.com/18F/cg-dashboard"
@@ -9,6 +7,7 @@ machine:
     CF_API_GC: "https://api.fr.cloud.gov"
     CF_ORGANIZATION_GC: "cloud-gov"
   post:
+    - cd cg-dashboard && nvm install && nvm use && nvm alias default $(cat .nvmrc)
     - mkdir -p download
     - test -e download/$GODIST || curl -o download/$GODIST https://storage.googleapis.com/golang/$GODIST
     - sudo rm -rf /usr/local/go
--- a/circle.yml
+++ b/circle.yml
@@ -1,4 +1,6 @@
 machine:
+  node:
+    version: 6.7.0
   environment:
     GODIST: "go1.6.linux-amd64.tar.gz"
     WS: "/home/ubuntu/.go_workspace/src/github.com/18F/cg-dashboard"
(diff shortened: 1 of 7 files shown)


tessereth:
Looks like the circle build fails unless I also pull across the changes to `circle.yml` I needed to make this build for us. I have no idea why these changes are required. In particular I had to change
```
machine:
  post:
    - cd cg-dashboard && nvm install && nvm use && nvm alias default $(nvm current)
```
to
```
machine:
  node:
    version: 6.7.0
```
This change is now included on this branch so you can see that the tests pass. I can remove it before merge if required.

circle.yml
```diff
@@ -1,4 +1,6 @@
 machine:
+  node:
+    version: 6.7.0
```
msecret: I'm concerned about this change because it would mean we'd have to update this in two places when we change node version, which would likely not happen. I can try and look into the build breaking.

tessereth:
So the reason I'm using Immutable is because I wanted a deep-merge function. A token amount of googling suggested a few different packages. In particular https://facebook.github.io/react/docs/update.html (https://github.com/kolodny/immutability-helper) came up. But then I found you were already using immutable so I figured I should use that to keep things consistent. But I agree it's a bit clunky. Do you have any other suggestions that would make it easy to override specific parts of the config?

tessereth:
So the circle config thing is strange. Looking at your circle logs, I see
```
$ cd cg-dashboard && nvm install && nvm use && nvm alias default $(nvm current)
...
Found '/home/ubuntu/cg-dashboard/.nvmrc' with version <6.7>
Now using node v6.7.0 (npm v3.10.3)
default -> v6.7.0
```
But when I run it, I get
```
$ cd cg-dashboard && nvm install && nvm use && nvm alias default $(nvm current)
...
Found '/home/ubuntu/cg-dashboard/.nvmrc' with version <6.7>
Now using node v6.7.0 (npm v3.10.3)
default -> system
```
So `nvm use` appears to be running correctly but `nvm current` returns `v6.7.0` for you and `system` for me. So my default alias ends up going to system and the build fails later because it can't find `npm`. After some experimentation, it seems I can use `$(cat .nvmrc)` rather than `$(nvm current)`. Does this seem like a viable solution?

adborden:
@tessereth I've seen the nvm issue before but I haven't dug too deep. The difference is our circleci is using ubuntu precise (12.04), and yours is using trusty (14.04). For some reason the `nvm alias default` command isn't setting the default to node `6.7.0` like it is on precise like you describe.

I think the `$(cat .nvmrc)` solution sounds reasonable.

tessereth:
Now updated based on AusDTO/cg-dashboard#3. Thanks @msecret!

PR #802, opened 2016-12-06
Org quick look icons

msecret:
Adds icons to the org name, and space and count status components.
A new component for the entity icons was created to reduce certain
code like setting the right status class, ensuring a dev doesn't
pass the wrong icon type (can only be space, app, org), and setting
defaults for icon color. Refactored the Icon component so it uses
styler interface to keep consistent with newer code.

There was also functionality added to differentiate between stroke
and fill icons. This is a limitation in the SVGs, some of our icons
really have to be stroke and some really have to be fills. The current
set of icons are fills while the previous ones are strokes. I'm opting
to keep the generic `icon` class as stroke icons and adding a new class,
`icon-fill` for these new fill icons so that this isn't a breaking
change for cg-site. I think this is a sane default to set anyway.
Unfortunetely, a dashboard developer will have to know whether their
icon is a stroke or fill when using the Icon class, which is not a great
interface. Lucikly for entity icons, EntityIcon takes care of it.

Requires 18F/cg-style#ms-entity_icon branch on cg-style.

### DIFF
--- a/static_src/components/count_status.jsx
+++ b/static_src/components/count_status.jsx
@@ -5,6 +5,7 @@ import style from 'cloudgov-style/css/cloudgov-style.css';
 
 import createStyler from '../util/create_styler';
 import { appStates } from '../constants.js';
+import EntityIcon from './entity_icon.jsx';
 
 const ICON_TYPES = [
   'space',
@@ -37,7 +38,12 @@ export default class CountStatus extends React.Component {
 
     return (
       <span className={ this.styler('count_status', statusClass) }>
-        <strong>{ props.count }</strong> { props.name }
+        <span style={{ float: 'left' }}>
+          <EntityIcon entity={ props.iconType } state={ props.status } />
+        </span>
+        <span style={{ float: 'right' }}>
+          <strong>{ props.count }</strong> { props.name }
+        </span>
       </span>
     );
   }
--- a/static_src/components/count_status.jsx
+++ b/static_src/components/count_status.jsx
@@ -38,10 +38,10 @@ export default class CountStatus extends React.Component {
 
     return (
       <span className={ this.styler('count_status', statusClass) }>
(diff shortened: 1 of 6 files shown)


msecret:
screen shot 2016-12-06 at 3 33 29 pm

adborden:
Maybe we can give the icons different filenames based on whether they are stroke or fill.

static_src/components/count_status.jsx
```diff
@@ -37,7 +38,12 @@ export default class CountStatus extends React.Component {
 
     return (
       <span className={ this.styler('count_status', statusClass) }>
-        <strong>{ props.count }</strong> { props.name }
+        <span style={{ float: 'left' }}>
```
adborden: I'd like to avoid the one-off styles and keep everything in `cg-style` and use class names. This keeps everything re-usable and documented in the pattern library. Also, forcing ourselves to name these kinds of things helps us keep our design patterns clean and consistent.

static_src/components/entity_icon.jsx
```diff
@@ -0,0 +1,41 @@
...
+import createStyler from '../util/create_styler';
+import style from 'cloudgov-style/css/cloudgov-style.css';
+
+const ENTITIES = ['app', 'space', 'org'];
+
+const STATE_MAP = {};
+STATE_MAP[appStates.running] = 'ok';
+STATE_MAP[appStates.started] = 'ok';
+STATE_MAP[appStates.stopped] = 'inactive';
+STATE_MAP[appStates.crashed] = 'error';
```
adborden: I think you can use the ES6 syntax here:
```javascript
const STATE_MAP = {
  [appStates.running]: 'ok',
  [appStates.started]: 'ok',
  [appStates.stopped]: 'inactive',
  [appStates.crashed]: 'error'
};
```

PR #804, opened 2016-12-08
Border both the app and space icons

msecret:
I decided to keep a list of bordered icons in Icon component, so the
user doesn't have to worry about what icons are supposed to be bordered.

When a new icon is added that has to be bordered, the dev will need
to add it to this list.


adborden:
Maintaining a list of bordered icons seems strange. What makes them always bordered or not bordered? What if instead there was a bordered boolean attribute that controlled this, and everywhere there was an app Icon, the developer would also specify `bordered=true`.

To me, having the explicit list feels like magic that could trip folks up.

PR #806, opened 2016-12-09
Set max items and add link to logs

line47:
Setting the max items on the recent activity and providing a link to logs.cloud.gov.

This relies on https://github.com/18F/cg-style/pull/208

### DIFF
--- a/static_src/components/activity_log.jsx
+++ b/static_src/components/activity_log.jsx
@@ -25,10 +25,13 @@ function stateSetter(props) {
 }
 
 const propTypes = {
-  initialAppGuid: React.PropTypes.string.isRequired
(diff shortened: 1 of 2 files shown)


static_src/components/activity_log.jsx
```diff
@@ -25,10 +25,13 @@ function stateSetter(props) {
 }
 
 const propTypes = {
-  initialAppGuid: React.PropTypes.string.isRequired
+  initialAppGuid: React.PropTypes.string.isRequire,
```
msecret: I think it should be `isRequired`

line47: Updated

PR #831, opened 2016-12-14
Upgrade cg-style to 2.3.0

msecret:


msecret:
Note: the code climate error is the same error from before, just repeated because of dep changes.

PR #842, opened 2016-12-15
Build improvements

msecret:
Fix the `check-style` command so it works better to distinguish when cg-style was installed from github.

Remove CSS modules from dev build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "build": "npm run clean && npm run check-style && node_modules/webpack/bin/webpack.js",
     "build-style": "cd node_modules/cloudgov-style && npm install && npm run build",
-    "check-style": "/bin/bash -c '[[ $(npm ls cloudgov-style) =~ \"github.com\" ]] && npm run build-style || exit 0'",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "build": "npm run clean && npm run check-style && node_modules/webpack/bin/webpack.js",
     "build-style": "cd node_modules/cloudgov-style && npm install && npm run build",
-    "check-style": "/bin/bash -c '[[ $(npm ls cloudgov-style) =~ \"github.com\" ]] && npm run build-style || exit 0'",
+    "check-style": "/bin/bash -c '[[ $(npm ls cloudgov-style) =~ \"git://github.com\" ]] && npm run build-style || exit 0'",
```
adborden: Hmm, I think I usually see `git+https://` in `package.json`. It's also uncommon, but folks might also put an ssh uri too. What does adding the protocol solve?

msecret: So currently, if I run `check-style` locally in any circumstance, it will trigger because some local path is `github.com/18F/cg-dashboard`. This fixes that.

PR #843, opened 2016-12-15
Refactor Action component to allow for HREF

line47:
Added HREF to the action component. 
Updated the login/log out button to use HREF action.
Removed about and contact links from top navigation.

Relies on https://github.com/18F/cg-style/pull/221

### DIFF
--- a/static_src/components/action.jsx
+++ b/static_src/components/action.jsx
@@ -1,3 +1,4 @@
+
 import React from 'react';
 import style from 'cloudgov-style/css/cloudgov-style.css';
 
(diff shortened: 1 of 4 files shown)


static_src/components/action.jsx
```diff
@@ -22,13 +22,14 @@ const BUTTON_TYPES = [
...
   clickHandler: React.PropTypes.func,
-  children: React.PropTypes.any
+  disabled: React.PropTypes.bool,
+  href: React.PropTypes.string,
+  label: React.PropTypes.string,
+  style: React.PropTypes.oneOf(BUTTON_STYLES),
+  type: React.PropTypes.oneOf(BUTTON_TYPES)
 };
 
 const defaultProps = {
```
adborden: How about `null`? I'm not a fan of empty string.

msecret: I don't think React will let you do that, but it'd be good to check.

PR #853, opened 2016-12-19
Setting icon size in the icon component

line47:
Setting up icon size variable and values for small medium and large.

### DIFF
--- a/static_src/components/entity_icon.jsx
+++ b/static_src/components/entity_icon.jsx
@@ -38,6 +38,7 @@ export default class EntityIcon extends React.Component {
       <Icon
         name={ this.props.entity }
         styleType={ stateClass }
+        iconSize="small"
         iconType="fill"
         bordered={ ['app', 'space', 'service'].includes(this.props.entity) }
       />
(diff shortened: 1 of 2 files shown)


line47:
Setting this icon size also gets us closer to merging https://github.com/18F/cg-style/pull/225

msecret:
Ahh I didn't want to do this

msecret:
I want to discuss this more and potentially take this commit out:

I designed the icons for a really easy: the conform to the size of the text they're contained in. This is an easier developer interface as the developer doesn't need to know abstract icon size (how big is `small`?) and can just drop icons and have them relatively well sized. Even if the sizes aren't perfect concerning the design, I think this is still a better system and want to push for it.

thisisdano:
I will argue against this

msecret:
Ok, how about we don't make it default to anything. 

This code is actually broken as is, I don't think the correct size class is being passed to the component because it's not passed as an array (which is interesting that the previous method of doing this was so successful that we didn't even realize this wasn't working).

So by default, no size will be defined but the develop can pass a size if necessary.

thisisdano:
That sounds like a good compromise — thanks!

PR #854, opened 2016-12-19
Styling updates to services panel

line47:
Updated alignment of columns and added borders to rows.
Updated adding a service button text.
Added a link in the panel head to the space the service lives in.

### DIFF
--- a/static_src/components/service_instance.jsx
+++ b/static_src/components/service_instance.jsx
@@ -185,8 +185,7 @@ export default class ServiceInstance extends React.Component {
               { serviceInstance.name }
             </span>
           </span>
-          <span className={ this.styler('panel-column', 'panel-column-less',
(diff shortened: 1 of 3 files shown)


static_src/components/service_instance.jsx
```diff
@@ -186,7 +185,8 @@ export default class ServiceInstance extends React.Component {
               { serviceInstance.name }
             </span>
           </span>
-          <span className={ this.styler('panel-column') }>
+          <span className={ this.styler('panel-column', 'panel-column-less',
```
msecret: This sucks to have to use `last` classes, because then you have to change them if the content changes. What are the new styles that require this?

line47: This was for the cost of the service aligning it to the right. Is there another class that should be used in this scenario?

PR #855, opened 2016-12-21
Updating usage panel and other panel headers

line47:
Updating the semantics of the headers used in panels

### DIFF
--- a/static_src/components/app_container.jsx
+++ b/static_src/components/app_container.jsx
@@ -140,7 +140,7 @@ export default class AppContainer extends React.Component {
           <h2>{ this.fullTitle } { this.statusUI }</h2>
           { this.restart }
           <Panel title="Usage and allocation">
-              <span>View more usage data at <a href="https:logs.cloud.gov">logs.cloud.gov</a></span>
+              <span>View more usage data at <a href="https://logs.cloud.gov">logs.cloud.gov</a></span>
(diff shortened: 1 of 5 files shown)


static_src/components/app_container.jsx
```diff
@@ -140,6 +140,7 @@ export default class AppContainer extends React.Component {
           <h2>{ this.fullTitle } { this.statusUI }</h2>
           { this.restart }
           <Panel title="Usage and allocation">
+              <span>View more usage data at <a href="https:logs.cloud.gov">logs.cloud.gov</a></span>
```
adborden: You're missing the `//` in the protocol, no?

adborden:
Are there cg-style changes that are paired with this PR?

line47:
No cg-style updates to pair with for this PR.

line47:
After looking over at cg-style it looks like this is paired with https://github.com/18F/cg-style/pull/229/

adborden:
:sunglasses: that looks much better. In that case, I'd like to call out that I don't think we should be using CSS rules based on elements, but rather create new class-based rules instead. I believe @msecret and @thisisdano are in agreement?

static_src/components/app_container.jsx
line47: Good catch, I've added the `//`

line47:
pairs with https://github.com/18F/cg-style/pull/238

Issue #648, opened 2016-10-05
Modify sizing and layouts of top page headings


msecret:
In order for a developer to clearly know what page they are looking at in the dashboard, the page headings and page layout should be large and clearly separated.

This will be included in: https://github.com/18F/cg-dashboard/issues/745
This is essentially the implementation issue for the breadcrumb: https://github.com/18F/cg-design/issues/24

This work is specific to page header at top of page, including setting up space underneath top nav where this information can go:
 - Overview/home: home icon, "Overview" displays (no breadcrumb)
 - Org/App/Space: breadcrumb nav displayed. Noting that because breadcrumb is wider, Open App and Restart App buttons move into panel below nav

Acceptance criteria:
 - [ ] There is space for a page headers in Overview/Org/App/Space pages
 - [ ] Page content in Overview/Org/App/Space pages has transparent area revealing light gray page background where actions (where available, i.e. App page) are displayed

msecret:
Show what I mean


![screen shot 2016-10-05 at 4 17 47 pm](https://cloud.githubusercontent.com/assets/1701077/19135102/1ee9bd96-8b17-11e6-842f-7a61d5a88711.png)

suprenant:
Feeling more comfortable tackling the nav. Entire header (including navigation) could be thornier

suprenant:
@msecret will take a look here, thinking this could be bigger than our usual standards.

line47:
Not a good issue for a React beginner, which I am.

msecret:
We may have to push this back because it's harder to implement then we wanted due to the react component hierarchy.

adborden:
Here's a relevant invision screen of the App page with actions. https://gsa.invisionapp.com/share/8X8RHAOKS#/screens/208180056
![screenshot from 2016-12-20 10-34-05](https://cloud.githubusercontent.com/assets/509703/21363068/dd82cfa4-c69f-11e6-905e-752cec5ec739.png)

PR #858, opened 2016-12-22
Move page title up to header

adborden:
This is work towards https://github.com/18F/cg-dashboard/issues/648

- Introduces immutable models for objects that don't come from the CF API e.g. `PageMetadata`
- Adds a `PageHeader` component that listens to stores and updates the page title
- Triggers the page load/unload actions from the component lifecycle events

It's definitely not ideal to be listening to so many stores in the `PageHeader`. Ideally, each page container component could call the `PAGE_METADATA_UPDATE` action to update the title itself since it knows the most about what the title of the page should be. This is a limitation of my react knowledge, but the components can't fire new actions during `_onChange` of the dispatch during the store update. I'm not sure if there's a better approach, let me know if you have ideas.

Remaining work:
- [ ] Add component for the page-level actions on the application page

### DIFF
--- a/static_src/components/app_container.jsx
+++ b/static_src/components/app_container.jsx
@@ -61,7 +61,6 @@ export default class AppContainer extends React.Component {
     AppStore.addChangeListener(this._onChange);
     OrgStore.addChangeListener(this._onChange);
     SpaceStore.addChangeListener(this._onChange);
-
     pageMetadataActions.loadPage('app');
   }
 
@@ -69,6 +68,7 @@ export default class AppContainer extends React.Component {
     AppStore.removeChangeListener(this._onChange);
     OrgStore.removeChangeListener(this._onChange);
     SpaceStore.removeChangeListener(this._onChange);
+    pageMetadataActions.unloadPage();
   }
 
   _onChange() {
--- a/static_src/components/app_container.jsx
+++ b/static_src/components/app_container.jsx
@@ -11,6 +11,7 @@ import Loading from './loading.jsx';
 import OrgStore from '../stores/org_store.js';
 import QuotaStore from '../stores/quota_store.js';
 import RoutesPanel from './routes_panel.jsx';
+import pageMetadataActions from '../actions/page_metadata_actions';
 import Panel from './panel.jsx';
 import PanelGroup from './panel_group.jsx';
 import ServiceInstancePanel from './service_instance_panel.jsx';
@@ -60,6 +61,8 @@ export default class AppContainer extends React.Component {
     AppStore.addChangeListener(this._onChange);
     OrgStore.addChangeListener(this._onChange);
     SpaceStore.addChangeListener(this._onChange);
+
+    pageMetadataActions.loadPage('app');
   }
 
   componentWillUnmount() {
--- a/static_src/components/page_header.jsx
+++ b/static_src/components/page_header.jsx
@@ -3,12 +3,14 @@ import style from 'cloudgov-style/css/cloudgov-style.css';
 
 import createStyler from '../util/create_styler';
 import AppStore from '../stores/app_store';
+import OrgStore from '../stores/org_store';
 import { pageKinds } from '../constants';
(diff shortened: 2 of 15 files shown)


adborden:
Currently, this looks like this.
screenshot from 2016-12-21 17-45-41

static_src/components/app_container.jsx
```diff
@@ -60,12 +61,14 @@ export default class AppContainer extends React.Component {
     AppStore.addChangeListener(this._onChange);
     OrgStore.addChangeListener(this._onChange);
     SpaceStore.addChangeListener(this._onChange);
+    pageMetadataActions.loadPage('app');
```
msecret: So rather then having the components fire this, what if they're fired in the main.js file. Then the title bar would set the current page type to `app` and since it's listening to the stores, it would re-update when the data comes in?

static_src/components/page_header.jsx
```diff
@@ -0,0 +1,101 @@
...
+// Prefer title set by props or PageMetadataStore, then fallback to guessing by
+// page kind
+function getTitle(props) {
+  const pageMetadata = PageMetadataStore.current;
+  let title = props.title || pageMetadata.title;
+  if (title) {
+    return title;
+  }
+
+  // TODO this seems strange that the page header needs to know about specific
```
msecret: One way to limit this: make a page title component for each org, space and app that inherit or use shared functionality. Then `main` decides which one to render, and they only need to know about one store each?

static_src/components/app_container.jsx
adborden: If we move this to `main.js`, then where would we put the unload action? That is similar to the problem I had with the breadcrumb prototype. You set a title on a page and then navigate to another page, say marketplace that doesn't set the title, you would be left with the old title.

In contrast, the container components seem like a better place to manage page lifecycle related events than the routes in `main.js` because naturally components come with their own lifecycle that aligns with the page.

static_src/components/page_header.jsx
adborden: Using specialized page header components could work, but how would `main.js` specify the pageheader? `main.js` renders the container component, but the `PageHeader` is actually nested in the `MainContainer` > `Header` components. If I'm understanding this correctly, what would the logic in `MainContainer` look like to render this?

```javascript
// main.js
app() {
  ReactDOM.render(
    <MainContainer>
      <AppPageHeader />
      <AppContainer />
    </MainContainer>);
}
```

```javascript
// main_container.js
render() {
const pageHeader = findPageHeaderElement(this.props.children);
const content = withoutPageHeaderElement(this.props.children);
return (
  <Header>
    <SiteHeader />
    { pageHeader }
  </Header>
  <main>
    { content }
  </main>
);
}
```

What would `findPageHeaderElement` look like? Or is there a better way.

adborden:
Hmm, @msecret you just gave me some ideas.

## Removing MainContainer

So, if `MainContainer` didn't exist and `main` just rendered `AppContainer`, then the page header would be under `AppContainer` and the title could be passed to the page header as props, the react way.

## Hoisting the title state up

So if `MainContainer` kept track of the title, it could provide an `updateTitle` method as a prop to the child container e.g. `AppContainer`, which could then update the title in the `_onChange`, like it should. This kind of breaks the assumptions about [flux containers](https://facebook.github.io/flux/docs/flux-utils.html#container) which is that containers don't have props, they get their data from state and only pass information downwards.

static_src/components/page_header.jsx
msecret: Yea I was thinking that idea, or passing the `<AppPageHeader />` as a prop, which we've done before.

static_src/components/app_container.jsx
msecret: Yea you would essentially have to operate similar to the side nav, unload the page on every route that doesn't have a title?

Also something to keep in mind: are there any pages left without a title now that marketplace and the space users, services routes are gone?

adborden:
This is succeeded by #904

PR #859, opened 2016-12-22
Org overview

msecret:
The org overview page, which successfully re-uses the space quick look. There's a new app quick look, which is a column layout of some simple app stats that can be configured. This is now re-used in the space quicklook and app list which is great. I had to play around with panel layouts to get everything consistent and reduce some complexity. I also moved the app list from a table to the panel layouts to allow for reusing the app quick look.

I also fixed a new set of React warnings.


static_src/components/app_quicklook.jsx
```diff
@@ -0,0 +1,112 @@
...
+const propTypes = {
+  app: React.PropTypes.object.isRequired,
+  orgGuid: React.PropTypes.string.isRequired,
+  spaceGuid: React.PropTypes.string.isRequired,
+  spaceName: React.PropTypes.string,
+  extraInfo: React.PropTypes.arrayOf((propVal) => EXTRA_INFO.includes(propVal))
+};
+
+const defaultProps = {
+  spaceName: '',
```
adborden: I feel like setting empty string as the default for optional props doesn't add much value. React is pretty good at handling `undefined` in that it basically renders as empty string. I would prefer to leave this prop out of the defaults.

static_src/components/org_container.jsx
```diff
@@ -0,0 +1,117 @@
...
+      if (space.apps && space.apps.length) {
+        return all.concat(space.apps);
+      }
+      return all;
+    }, []);
+  }
+
+  render() {
+    const state = this.state;
+    let loading = <Loading text="Loading org" />;
```
adborden: We're inconsistent with these abbreviated names. Here is "org", the empty state is "organizations" cc @thisisdano

static_src/components/org_container.jsx
```diff
@@ -0,0 +1,117 @@
...
+    let content = <div>{ loading }</div>;
+
+    if (state.empty) {
+      content = <h4 className="test-none_message">No organizations</h4>;
+    } else if (!state.loading && state.org) {
+      const allApps = this.allApps();
+      const allServices = this.allServices();
+
+      // TODO repeated pattern space_container, overview
```
adborden: Maybe an `EntityLayout` component?

static_src/components/panel_row.jsx
```diff
@@ -15,9 +15,7 @@ const propTypes = {
   styleClass: React.PropTypes.oneOf(STYLES)
 };
 
-const defaultProps = {
-  styleClass: 'none'
```
adborden: Since the propTypes specifies `styleClass` as an enum, if you don't pass a styleClass, won't you get an error?

static_src/components/panel_row.jsx
```diff
@@ -29,9 +27,10 @@ export default class PanelRow extends React.Component {
   render() {
     const props = this.props;
     const styleClass = props.styleClass && `panel-row-${this.props.styleClass}`;
+    const mainClass = props.styleClass !== 'boxed' && 'panel-row';
 
     return (
-      <div className={ this.styler('panel-row', styleClass) }>
+      <div className={ this.styler(mainClass, styleClass) }>
```
adborden: I liked the way it was working before, using css base and a css modifier class. Why the need for this change?

static_src/main.js
```diff
@@ -56,9 +56,10 @@ function overview() {
 function org(orgGuid) {
   orgActions.toggleSpaceMenu(orgGuid);
   orgActions.fetch(orgGuid);
+  cfApi.fetchSpaces().then(() => spaceActions.fetchAllForOrg(orgGuid));
```
adborden: This sticks out to me, since we fetch the org via action, but the spaces are fetched with the api directly. I'm not totally sure "what's right" but I feel like moving the action handling into cfApi gives the caller more control (because we have access to the request's promise).

static_src/components/org_container.jsx
line47: I prefer the "organizations" to "org" as it's more descriptive

static_src/components/org_container.jsx
msecret: I want to wait for more PL work merged as I think the new overview component might fit in on some of these.

static_src/components/panel_row.jsx
msecret: I'm not getting any errors, let me play around with it.

static_src/components/panel_row.jsx
msecret: Because the boxed component is almost becoming a separate thing. I think I want to split it out into it's own component. Again, I'm waiting on the PL work because some of the work on the overview component might make sense. I actually want to sit down and refactor the whole panel component before we end.

static_src/main.js
msecret: Yea I agree, this is the second of times I've had to do this and I really require something more for it. It is a case where putting the CF api calls in the actions and then returning the promise makes sense, but then I've also seen advice not to do that (although I don't remember why). But we definitely need a way of do this action first and then do this action.

PR #871, opened 2016-12-29
Fix styles of count_status

adborden:
Paired with https://github.com/18F/cg-style/pull/248

- Use count_status-icon
- Use count_status_container
- Specify sizes for EntityIcon (quick fix)

Ideally, the EntityIcon would not need to specify a size, but it currently is causing the entire quick look panel to expand. This is the easier fix.

### DIFF
--- a/static_src/components/space_quicklook.jsx
+++ b/static_src/components/space_quicklook.jsx
@@ -42,7 +42,7 @@ export default class SpaceQuicklook extends React.Component {
       content = (
         <PanelRow>
           <h3>
-            <EntityIcon entity="space" />
+            <EntityIcon entity="space" iconSize="medium" />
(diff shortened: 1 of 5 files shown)


adborden:
screenshot from 2016-12-29 09-14-16

static_src/components/space_quicklook.jsx
```diff
@@ -42,7 +42,7 @@ export default class SpaceQuicklook extends React.Component {
       content = (
         <PanelRow>
           <h3>
-            <EntityIcon entity="space" />
+            <EntityIcon entity="space" iconSize="medium" />
```
msecret: What changed for us to need to set the icon size on all of these?

PR #892, opened 2016-12-30
New panel logic and overview header

thisisdano:
This PR adds logic to the panel component to only include an `h1.panel-title` is the panel is explicitly given a title. If the title is left blank, the h1 will be omitted.

This allows flexibility in the space/org overview panel to include a special panel header that includes status and a bottom rule.

This is something of a pair with https://github.com/18F/cg-style/pull/250#pullrequestreview-14729427

PR #893, opened 2016-12-30
Goodbye sidebar! (The templates to make this work)

thisisdano:
This updates the templates to remove the sidebar, and adds styling hooks for dashboard elements.

**Paired with https://github.com/18F/cg-style/pull/255**
**Based off Icon/Panel work in https://github.com/18F/cg-dashboard/pull/892 and https://github.com/18F/cg-style/pull/250 — Merge them first.**

- - -

screen shot 2016-12-30 at 7 27 00 am

- - -

screen shot 2016-12-30 at 7 27 18 am

### DIFF
--- a/static_src/components/disclaimer.jsx
+++ b/static_src/components/disclaimer.jsx
@@ -21,7 +21,7 @@ export default class Disclaimer extends React.Component {
       flagImg = <img alt={ flagAlt } src={ flag }></img>;
     }
     return (
-      <div className={ this.styler('usa-disclaimer') }>
+      <div className={ this.styler('usa-disclaimer', 'disclaimer-dashboard') }>
         <div className={ this.styler('grid') }>
           <span className={ this.styler('usa-disclaimer-official') }>
             { config.header.disclaimer }
--- a/static_src/components/disclaimer.jsx
+++ b/static_src/components/disclaimer.jsx
@@ -21,7 +21,7 @@ export default class Disclaimer extends React.Component {
       flagImg = <img alt={ flagAlt } src={ flag }></img>;
     }
     return (
-      <div className={ this.styler('usa-disclaimer', 'disclaimer-dashboard') }>
+      <div className={ this.styler('usa-disclaimer', 'disclaimer-no_sidebar') }>
         <div className={ this.styler('grid') }>
           <span className={ this.styler('usa-disclaimer-official') }>
             { config.header.disclaimer }
--- a/static_src/components/header.jsx
+++ b/static_src/components/header.jsx
@@ -35,7 +35,7 @@ export default class Header extends React.Component {
       </Action>
     </HeaderLink>;
     return (
-    <header className={ this.styler('header', 'header-dashboard') }>
+    <header className={ this.styler('header', 'header-no_sidebar') }>
       <div className={ this.styler('header-wrap') }>
         <div className={ this.styler('header-title') }>
           <a href="/" className={ this.styler('logo') } title="Home">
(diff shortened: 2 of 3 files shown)


static_src/components/disclaimer.jsx
```diff
@@ -21,7 +21,7 @@ export default class Disclaimer extends React.Component {
       flagImg = <img alt={ flagAlt } src={ flag }></img>;
     }
     return (
-      <div className={ this.styler('usa-disclaimer') }>
+      <div className={ this.styler('usa-disclaimer', 'disclaimer-dashboard') }>
```
thisisdano: hook to set dashboard-specific width

static_src/components/header.jsx
```diff
@@ -35,7 +35,7 @@ export default class Header extends React.Component {
       </Action>
     </HeaderLink>;
     return (
-    <header className={ this.styler('header') }>
+    <header className={ this.styler('header', 'header-dashboard') }>
```
thisisdano: hook to set dashboard-specific width/styling

static_src/components/main_container.jsx
```diff
@@ -43,11 +43,9 @@ export default class App extends React.Component {
...
   render() {
     let content;
-    let sidebar;
 
     if (this.state.isLoggedIn) {
       content = this.props.children;
-      sidebar = <Nav />;
     } else {
       content = <Login />;
     }
```
thisisdano: removed sidebar references

static_src/components/main_container.jsx
```diff
@@ -57,11 +55,8 @@ export default class App extends React.Component {
       <div>
         <Disclaimer />
         <Header />
-        <div className={ this.styler('sidenav-parent', 'main_content', 'content-dashboard') }>
-          <nav className={ this.styler('sidenav') }>
-            { sidebar }
-          </nav>
-          <main className={ this.styler('sidenav-main', 'usa-content') }>
+        <div className={ this.styler('main_content', 'content-dashboard') }>
+          <main className={ this.styler('usa-content') }>
```
thisisdano: removed sidebar references

adborden:
@thisisdano are the `-no_sidebar` classes temporary? When would we want to remove them? Might be good to leave a comment to remind someone.

thisisdano:
@adborden I wonder what we should do about those styles. I'd originally named them `-dashboard`, but Marco preferred this more 'universal' name. However, the purpose of the special style is to provide a hook specific to the dashboard width, which is probably narrower (not not _necessarily_ narrower) than the width of the marketing home/docs pages.

We will probably want/need a dashboard width that's distinct from the other pages. How best to style? I'd appreciate your thoughts!

adborden:
Idk, with the sidenav sticking around for the docs site, I feel like the sidenav should somehow fit in the grid, then there would be two styles of layout, one "full-width" single column (dashboard) and then the second would be the two-column side-nav + main content layout (docs). If that was the case, then I would pick names like `-full` (single-column) and `-secondary` + `-main` (two-column).

I'm not really sure what to do in the meantime, maybe use `-full` to signify it's the full width, but it's a bit confusing because in contrast when using a sidenav you're using a non-standard grid where the main column is full-width anyway.

adborden:
Hmm, now that I say that, I would make the full-widht version the first-class citizen, and the two-column layout the non-standard, by making that one require the `-with_sidebar`. That way it feels awkward and implies that there is work to do to make the docs site fit with the new grid.

msecret:
@adborden maybe we should make your suggestions tech debt stories?

PR #905, opened 2016-12-31
Redis session backend

### LINKED ISSUE #751: Don't use filesystem store
We're currently using the filesystem session store. That means session data isn't shared across app instances, so if we ever scale to more than one instance, users will be load-balanced between instances with potentially different local session stores. To fix, we can switch to a shared session store, like the redis store. Alternatively, we can use [sticky sessions](https://docs.cloudfoundry.org/concepts/http-routing.html) to avoid users switching between instances, but we'd still lose session data on restart.

jmcarp:
Resolves #751.

### DIFF
--- a/helpers/settings.go
+++ b/helpers/settings.go
@@ -3,6 +3,7 @@ package helpers
 import (
 	"encoding/gob"
 	"errors"
+	"fmt"
 	"net/url"
 	"os"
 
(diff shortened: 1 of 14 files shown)


jmcarp:
Verified that this works locally, but this should be tested on staging before merge.

@adborden @jcscottiii

helpers/settings.go
```diff
@@ -115,3 +136,43 @@ func (s *Settings) InitSettings(envVars EnvVars) error {
...
+func getRedisSettings(env *cfenv.App) (string, string, error) {
+	uri, err := getRedisService(env)
+	if err != nil {
+		uri = os.Getenv("REDIS_URI")
+	}
+	if uri == "" {
+		uri = "redis://localhost:6379"
+	}
+
+	u, err := url.Parse(uri)
```
msecret: What does `u` stand for?

manifests/manifest-base.yml
```diff
@@ -2,8 +2,10 @@
 buildpack: go_buildpack
 memory: 256M
 services:
-  - deck-ups
+- deck-ups
+- deck-redis
```
msecret: Should we be calling these `dashboard` now? Or maybe this is not the right time to make a name change, that's fine too.

jcscottiii:
looks good to me. just want to make sure navigator has time to test this out in staging first

helpers/settings.go
rogeruiz: That Parse method returns a tuple (two things) and the first one is the parsed uri which is set to `u` and the second one is an error set to `err`

manifests/manifest-base.yml
rogeruiz: I'm not sure if renaming a service will affect the way the app parses the VCAP_SERVICES, but you can run this @msecret and submit a PR which updates the VCAP lookup.

https://docs.cloudfoundry.org/devguide/services/managing-services.html#rename_service

rogeruiz: We're going to rename `deck-redis` to `dashboard-redis` but we don't wan to touch `deck-ups`. I'm around to pair on that change if you'd like though @msecret.

Issue #860, opened 2016-12-22
Space overview page shows crashed app as ok


adborden:
I created `adborden-node-crashed` which never opens a port for listening so the health check fails and CF reports the app as crashed. You can see on the Space Overview in the dashboard, `adborden-node-crashed` shows up as green.

This is in the [cf-deck-testing/dev](https://dashboard-staging.apps.cloud.gov/#/org/48b3f8a1-ffe7-4aa8-8e85-94768d6bd250/spaces/82af0edb-8540-4064-82f2-d74df612b794) org/space.

![screenshot from 2016-12-22 14-13-57](https://cloud.githubusercontent.com/assets/509703/21442163/fd93469a-c850-11e6-90eb-b521492a4ca1.png)
![screenshot from 2016-12-22 14-12-53](https://cloud.githubusercontent.com/assets/509703/21442167/02480cac-c851-11e6-9685-112c56205ad9.png)

Notes:
 - We should only say okay if it's fully okay

msecret:
@adborden do you know what `status` show up? In that screenshot, it looks like CF thinks the app is started, not crashed.

adborden:
The app state is "STARTED". 

fwiw, the cf cli uses the condition [`runnning_instances == 0 && state !== STOPPED`](https://github.com/cloudfoundry/cli/blob/ffee70cca68ab15df71f48641f6fa3bb007c4a2b/cf/uihelpers/ui.go#L20) to determine crashed... although I assumed rebooting would also look like that.

Here's the full app summary https://gist.github.com/adborden/548719c53bfd26d4ce46e6bf11bd8f7a

adborden:
Hmm, it looks like there are more state values on the app instance. Maybe we should include that information to determine if an app is crashing https://github.com/cloudfoundry/cli/blob/ffee70cca68ab15df71f48641f6fa3bb007c4a2b/cf/models/application.go#L76

It kind of makes sense because an app is really either starting, stopped, or in some state of running. YOu could have only one instance of several crashed, and that would mean the app is running, but not in a healthy state. cc @thisisdano

adborden:
And here's the cli output for the app details, including the instances:
```
$ cf app adborden-node-crashed
Showing health and status for app adborden-node-crashed in org cf-deck-testing / space dev as aaron.borden@gsa.gov...
OK

requested state: started
instances: 0/1
usage: 64M x 1 instances
urls: adborden-node-crashed.apps.cloud.gov
[...]

PR #920, opened 2017-01-04
Better heuristics around application health

adborden:
This is the first part in [fixing the app health indicator](https://github.com/18F/cg-dashboard/issues/860).

- [ ] This depends on styles in https://github.com/18F/cg-style/pull/276

screenshot from 2017-01-04 15-31-38

### DIFF
--- a/static_src/constants.js
+++ b/static_src/constants.js
@@ -2,14 +2,31 @@
 import keymirror from 'keymirror';
 
 const appStates = {
+  crashed: 'CRASHED',
+  down: 'DOWN',
+  flapping: 'FLAPPING',
+  none: 'NONE',
+  restarting: 'RESTARTING',
   running: 'RUNNING',
   started: 'STARTED',
   stopped: 'STOPPED',
-  crashed: 'CRASHED',
-  restarting: 'RESTARTING',
-  none: 'NONE',
   default: 'STOPPED'
 };
+
+// Descriptive states to represent entity health
+const entityHealth = keymirror({
+  // Healthy entity
+  ok: null,
+  // Something is not right, might want to investigate
+  warning: null,
+  // Something is definitely not right and requires some attention
+  error: null,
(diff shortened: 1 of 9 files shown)


static_src/util/health.js
```diff
@@ -0,0 +1,63 @@
...
+export function appHealth(app) {
+  if (!app) {
+    throw new Error('`app` must be provided.');
+  }
+
+  if (app.state === appStates.started && app.running_instances === app.instances) {
+    return entityHealth.ok;
+  }
+
+  if (app.running_instances === 0) {
```
msecret: Are we sure theres no case where running instances would be more then 1 and app state would be stopped?

static_src/constants.js
```diff
@@ -2,15 +2,32 @@
...
   started: 'STARTED',
   stopped: 'STOPPED',
-  crashed: 'CRASHED',
-  restarting: 'RESTARTING',
-  none: 'NONE',
   default: 'STOPPED'
 };
 
+// Descriptive states to represent entity health
+const entityHealth = keymirror({
```
msecret: I like how we are separating them, it's easier to understand

static_src/constants.js
```diff
@@ -2,15 +2,32 @@
 import keymirror from 'keymirror';
 
 const appStates = {
+  crashed: 'CRASHED',
+  down: 'DOWN',
+  flapping: 'FLAPPING',
```
msecret: ???

adborden: I got this from the CLI, but not sure under what conditions all of these might appear.

static_src/util/health.js
adborden: As far as I can tell, no. I suspect there might be some corner cases to handle. For example, I have seen `running_instances === -1`, when the app is restarting/restaging.

Issue #914, opened 2017-01-04
When route edit form is open, cancel button appears disabled but should not be


suprenant:
What we're after: display cancel/OK buttons here in a way to avoid confusing users.

Notes: 
 - Pattern library should have complete set of button states
 - OK state (hover) color: darker green
 - Cancel: base outline

![screen shot 2017-01-04 at 12 57 05 pm](https://cloud.githubusercontent.com/assets/1452369/21654948/6881bc6c-d27d-11e6-9f78-575e57944ddd.png)

Acceptance criteria:
 - [x] Cancel button is displayed in black with a black outline
 - [x] OK button is filled green

adborden:
LGTM

PR #922, opened 2017-01-05
Remove deprecated buttons

msecret:
Some buttons have been deprecated. This PR removes them from being used in dashboard, fixing button colors where they are incorrect.

It also fixes other React warnings due to incorrect prop usage.

Ref https://github.com/18F/cg-dashboard/issues/914

### DIFF
--- a/static_src/components/route.jsx
+++ b/static_src/components/route.jsx
@@ -104,7 +104,7 @@ export default class Route extends React.Component {
   bindAction(unbind) {
     return (
       <Action key="unbind" label={ (!!unbind) ? 'Unbind' : 'Bind' }
-        style={ (!!unbind) ? 'warning' : 'base' } type={ (!!unbind) ? 'link' : 'button' }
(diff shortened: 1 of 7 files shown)


static_src/components/route.jsx
```diff
@@ -104,7 +104,7 @@ export default class Route extends React.Component {
   bindAction(unbind) {
     return (
       <Action key="unbind" label={ (!!unbind) ? 'Unbind' : 'Bind' }
-        style={ (!!unbind) ? 'warning' : 'outline' } type={ (!!unbind) ? 'link' : 'button' }
+        style={ (!!unbind) ? 'warning' : 'primary' } type={ (!!unbind) ? 'link' : 'button' }
```
line47: In the sketch file the bind buttons are `outline` and not `primary`. Are we changing this?

msecret: Are they black outlines or blue outlines?

line47: Blue outlines

Issue #915, opened 2017-01-04
Modify route, click OK, doesn’t work, click cancel, shows editing app indicator when an error has occurred


suprenant:
What we're after: users trying to modify routes should either not be able to modify routes for which they do not have permissions and/or received clear errors for actions for which they don't have permissions or are otherwise invalid.

Notes: 
 - That this only happens when user doesn't have permission to modify a route
 - Display error where rest of errors are showing in UI
 - We'll need to review how permissions are currently working here

Possible fixes:
1. Users don't have access to change items they don't have permissions for
2. Correctly display error messages all the time

![screen shot 2017-01-04 at 12 57 05 pm](https://cloud.githubusercontent.com/assets/1452369/21654990/90d0fe12-d27d-11e6-91b9-7136690b2c3f.png)

Acceptance criteria:
 - [x] In the case that a modification is attempted without appropriate permissions, an error would be clearly visible

suprenant:
Noting that we're going for option #2 above and will consider #1 if there's time.

adborden:
Error appears correctly.

PR #923, opened 2017-01-05
Fixed route and other error handling

msecret:
A change, maybe in axios ajax module, maybe somewhere else changed
the error responses so the actual data resides in a `response` object
within the error object. This was breaking the error catching. The
fix is the pass the error `response` object to the `handleError`
function so the correct data is supplied to it so it can pass the
error correctly to the UI with the correct information.

Additionally, the route store wasn't resetting the route loading state after
and error, which was causing the editing loading text to appear after any
error.

Ref https://github.com/18F/cg-dashboard/issues/915

### DIFF
--- a/static_src/stores/route_store.js
+++ b/static_src/stores/route_store.js
@@ -218,7 +218,8 @@ class RouteStore extends BaseStore {
         const route = this.get(action.routeGuid);
         if (!route) break;
         const newRoute = Object.assign({}, route, {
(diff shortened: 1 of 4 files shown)


adborden:
Not sure this should be applied globally to all the errors. The `catch` could be catching any error, not just from axios which might not have a response property. Might be better to have a clause in the `handleError` that checks for the `response` property instead.

PR #924, opened 2017-01-05
Better handling of app instance state

adborden:
Fetching stats actually gives us instance data, so we can get richer data if we treat it that way. We merge the instance stats into `app.app_instances`. To calculate instance usage, we average across the instances. To get the total usage, we sum them. To get the app status, we find the worse state within the instances.

This fixes an issue on the app  page where we show the wrong status label and gives us more accurate numbers in the usage and allocation panel.

### DIFF
--- a/static_src/util/cf_api.js
+++ b/static_src/util/cf_api.js
@@ -246,7 +246,7 @@ export default {
 
   fetchAppStats(appGuid) {
     return http.get(`${APIV}/apps/${appGuid}/stats`).then((res) => {
-      appActions.receivedAppStats(appGuid, res.data[0]);
+      appActions.receivedAppStats(appGuid, { app_instances: Object.values(res.data) });
     }).catch((err) => {
       handleError(err);
     });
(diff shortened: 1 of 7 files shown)


static_src/util/cf_api.js
```diff
@@ -246,7 +246,7 @@ export default {
 
   fetchAppStats(appGuid) {
     return http.get(`${APIV}/apps/${appGuid}/stats`).then((res) => {
-      appActions.receivedAppStats(appGuid, res.data[0]);
+      appActions.receivedAppStats(appGuid, { app_instances: Object.values(res.data) });
```
msecret: Why do we only want to pass instances for this now?

adborden: Before, we were basically merging the first instance into the app level which caused some undesired behavior. Mainly this is an issue for `state` because both apps and app instances have state but each has different values. For example, app will only have state `started` or `stopped` to reflect the "requested" state. App instances can have many more states like `starting`, `running`, `crashing`, etc.

So depending on how the page loaded, for a running app, you might see the app label as "STARTED" or as "RUNNING" which was inconsistent. Potentially for a crashing app, you might also see "STARTED" up there.

By keeping the instances in a separate field, we can keep track of the app `state` as well as each instances' `state`.

Issue #869, opened 2016-12-29
Double loading on Overview


adborden:
As the page is loading, we see the spinner, then the org quick look panels come in for a moment, then we get the loading again, and finally we get our loaded quick look panels.

![peek 2016-12-28 17-59](https://cloud.githubusercontent.com/assets/509703/21535294/b0ef4c86-cd27-11e6-9082-d813ac761593.gif)

thisisdano:
Here's another view of this:

![loading](https://cloud.githubusercontent.com/assets/11464021/21777982/44d35956-d656-11e6-8c6d-dd524164a9ec.gif)

adborden:
I'm still seeing this behavior, kicking it back to Ready.

PR #929, opened 2017-01-06
Panels, Columns, and Rows

thisisdano:
**Work in progress**
**Pair with https://github.com/18F/cg-style/pull/284**

@msecret @adborden @line47 looking for initial feedback

Basically, this work tries to regularize the way we use rows and columns — to accommodate for the presentational layout uses (panels and layouts that use rows and columns as a grid layout system — as in app usage) and the semantic (`panel-row`) use.

- `PanelRow` component -> `Row` with vars to pass; more flexible use of gutters
- new `Col` component
- slight rebuilds for `contents-tree`
- more consistent display of actions
- doesn't break existing views _afaik_

Still work to be done to remove extraneous vars in `Row` and `Col` and to make these components as easy to use as possible.

But this allows for very nice overall display — as in this app view:

screen shot 2017-01-06 at 10 14 17 am


thisisdano:
Added Card component as well (for Overview, and perhaps also for Marketplace)

thisisdano:
@adborden can you point me to where `panel-content` exists outside of a panel — I'm having trouble tracking it down. It should not exist outside of a panel.

Columns and rows may exist outside of a panel as they are layout elements, just like the grid. Specific types of rows (like our good friend `panel-row` are intended for panel use only)

adborden:
[Here's an example](https://github.com/18F/cg-dashboard/pull/929/files#diff-11b5954b219cf347a76cca7135985ef8R91) of `panel-content` in the ActivityLog. Why should the ActivityLog have to exist within a panel? Why doesn't the parent have the `panel-content` class and the ActivityLog exists within that?

thisisdano:
Yup. Sorry, I'd thought of that as panel content. Agree that it _needn't_ be. The only reason I can think of that I did it this way was that I probably ran into some react error I couldn't get out of in another way. Totally agree that `panel-content` should live at the panel level for simplicity. I'd have to look into why I thought I couldn't do it that way!

thisisdano:
To be explicit, I'm very _very_ open to others helping on this branch. If you can jump in and make a change that I seem to inept to do on my own, I'd appreciate it.

adborden:
I might also suggest that instead of making the changes globally, maybe you just start with a single component/page so it's easier to focus on the core changes. It also means there's less broad sweeps in case you need to make adjustments or respond to feedback. I can see this becoming another large PR.

We definitely have good tools with SCSS that allow us to easily override the styles for a single component so the new and old styles can exist side by side without you having to update _every_ component.

thisisdano:
I agree in theory...  though since I've already made all the changes globally, idk if I can go back and limit scope. If these weren't all things we were trying to finish before the death of the stream, there'd be reasonable time to work more cautiously.

thisisdano:
OK — I've made a number of additions and improvements over the weekend:
- Simplified `Row`
  - Now only takes `valign` and `gutters` props
  - `panel-row` type now its own component: `PanelEntry` [not in PL yet]
- Simplified `Col`
  - new component `ContentsTreeColumn` for data columns in the contents tree [not in PL yet]
- Regularized contents tree components
  - `AppQuicklook` now `ContentsTreeApp` [documented in PL]
  - `SpaceQuicklook` now `ContentsTreeSpace` [documented in PL]
  - documented
- Changed block loading component
  - now the same as inline loading indicator
  - now can display loading text
  - made a regularized loading container (`.loading-container`) for loading in panels and cards. This prevents panel "jumping" on loading. [documented in PL]
- removed icon from Org in overview card
- removed the gray path from `ContentsTreeApp` [documented in PL]
- Changed `secondary` prop in `Card` (the top-right content) to `helperText` [not in PL yet]
- Gave `main` a min-height to keep the footer at the bottom of the window

loading

Changing loading revealed that expand actions in the overview trigger loads on _all_ open cards.

We also get the disconcerting "double-load" on the Overview as well, with a quick load (perhaps with no status information) that leads to a second, more complete load in quick succession. The preliminary load feels unnecessary and confusing. (I see that this issue is already noted in https://github.com/18F/cg-dashboard/issues/869)

static_src/components/app_container.jsx
```diff
@@ -160,7 +160,7 @@ export default class AppContainer extends React.Component {
               </PageHeader>
             </div>
           </div>
-          <Panel title="Usage and allocation" secondary={ usageSecondary }>
+          <Panel title="App usage and allocation" secondary={ usageSecondary }>
```
line47: Could we name this panel `Usage and allocation` and drop the `App`? 
None of the other panels on this page are using the word `App` before the panel title.

Or on the flip side, add `App` to the other panels. On the other pages we are using `Org`, `Space` etc... in the panel titles.

static_src/components/row.jsx
```diff
@@ -0,0 +1,59 @@
+
+import React from 'react';
+import style from 'cloudgov-style/css/cloudgov-style.css';
+import createStyler from '../util/create_styler';
+
+const STYLES = [
+  'bordered',
+  'boxed',
+  'clean', // TODO this should be reconciled with panel-row and panel-row-bordered
+  'none'
```
line47: What's the difference between clean and none?

static_src/components/app_container.jsx
thisisdano: Yeah, dunno! I added `App` to make it more consistent with the other top panels in the other views, but I didn't think they _all_ needed `App`. But why not? The more the merrier!

static_src/components/row.jsx
thisisdano: I thought I removed this entire STYLES thing.... These are all deprecated.

thisisdano:
It's true — I don't know the best way to fix.

static_src/components/app_container.jsx
```diff
@@ -166,18 +170,22 @@ export default class AppContainer extends React.Component {
...
 
-          <Panel title="Routes">
-            <RoutesPanel />
-          </Panel>
-
-          <Panel title="Services">
-            <ServiceInstancePanel />
-          </Panel>
+          <Row gutters valign="top">
+            <Col flex={ 1 }>
```
msecret: This is an example of where I think we went from a simple interface; just dropping panels side by side when you need, to a much more complex one; Using `Col` with a `flex` property that I wouldn't understand as a developer unless I went into the `Col` code. What are the `Col`s doing that can just be done with the `Panel`?

thisisdano: setting up two columns

thisisdano:
I totally agree that in the case of `panel-content` and `loading-container`, it's DRYer and makes sense to fold duplicative classing into the Component. I think I just need help doing this because it involves a bit more complicated `props`ing than I'm comfortable building on my own.

That said, this kind of issue issue exists outside of this PR as well. Is it something that needs to be addressed now to make progress with this PR? If so, is it complicated? Can I get help? I just don't want a baby/bathwater situation.

msecret:
I'll try and work on this and see what I can do. I'm not totally sure how much time I'll have with the big PI planning though, so it might not get done for a while. We might be able to do this work in the new PI as well and we've talked about doing refactor work. In terms of other PRs that rely on this one, maybe just continue with them and we'll figure out later how to merge them in? At the very least, we can see what you were proposing to do and can rebuild new PRs from that.

If there are layout problems that you want to see fixed sooner, I'd suggest opening new PRs off master for those that use simpler, more one-off solutions?

thisisdano:
Whatever works, I'd appreciate. Thanks! Also, I have plenty of interest in continuing to iron out design issues, even when I am not officially working on this project. I have plenty invested in its successful execution.

PR #935, opened 2017-01-07
Refactor org quicklook loading

adborden:
With this PR I'm introducing some new guidelines about handling async state between components, actions, and stores.
- Each action creator (each method in `*_actions.js`)  should dispatch only
  a single action.
- Any XHR calls should happen within the fetch action creators which
  should return a promise of the request.
- Any `fetch` action should call the appropriate `receive`/`success`/`error` action on the resolve of the fetch's promise.
- Each async action should have a fetch, success, and error sub-action (e.g.
  `TOGGLE`, `TOGGLE_SUCCESS`, and `TOGGLE_ERROR`).

I've refactored the SpaceActions and SpaceStore to follow these principles (minus the error action). The org toggle action follows the fetch/success/error principle.

I feel like with these rules, it's a little more clear on how to manage async state. Fetch actions attach the success/error to the request promise. Success/error actions only fire the actions.

I've also removed the extra loading state from the SpaceQuicklook. Since the toggle is what controls the loading, the Org quicklook needs to conditionally render the Loading or SpaceQuicklook.

**Remaining work**
- [x] account for loading state on the space page, which was formerly wrapped into the space quicklook
- [ ] display the error in the UI

### DIFF
--- a/static_src/actions/space_actions.js
+++ b/static_src/actions/space_actions.js
@@ -5,7 +5,9 @@
  */
 
 import AppDispatcher from '../dispatcher.js';
+import cfApi from '../util/cf_api';
 import { spaceActionTypes } from '../constants.js';
+import SpaceStore from '../stores/space_store';
 
 export default {
   fetch(spaceGuid) {
@@ -13,12 +15,18 @@ export default {
       type: spaceActionTypes.SPACE_FETCH,
       spaceGuid
     });
+
+    // TODO error action should also be specified here
+    return cfApi.fetchSpace(spaceGuid).then(this.receivedSpace);
   },
 
   fetchAll() {
     AppDispatcher.handleViewAction({
       type: spaceActionTypes.SPACES_FETCH
     });
+
+    // TODO error action should also be specified here
+    return cfApi.fetchSpaces().then(this.receivedSpaces);
   },
 
   fetchAllForOrg(orgGuid) {
@@ -26,6 +34,12 @@ export default {
       type: spaceActionTypes.SPACES_FOR_ORG_FETCH,
       orgGuid
     });
+
+    // TODO error action should also be specified here
+    return Promise.all(SpaceStore.getAll()
+      .filter(space => space.organization_guid === orgGuid)
+      .map(space => this.fetch(space.guid))
+    );
   },
 
   receivedSpace(space) {
--- a/static_src/util/cf_api.js
+++ b/static_src/util/cf_api.js
@@ -9,7 +9,6 @@ import loginActions from '../actions/login_actions.js';
 import orgActions from '../actions/org_actions.js';
 import quotaActions from '../actions/quota_actions.js';
 import routeActions from '../actions/route_actions.js';
-import spaceActions from '../actions/space_actions.js';
 import serviceActions from '../actions/service_actions.js';
 import userActions from '../actions/user_actions.js';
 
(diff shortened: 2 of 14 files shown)


adborden:
@msecret still have some work remaining on this but would like your thoughts on this.

static_src/actions/space_actions.js
```diff
@@ -5,27 +5,41 @@
...
 
 export default {
   fetch(spaceGuid) {
     AppDispatcher.handleViewAction({
       type: spaceActionTypes.SPACE_FETCH,
       spaceGuid
     });
+
+    // TODO error action should also be specified here
+    return cfApi.fetchSpace(spaceGuid).then(this.receivedSpace);
```
msecret: Is there a reason this is calling the `receivedSpace` action? Leaving it in the api would be less of a change.

static_src/util/cf_api.js
```diff
@@ -47,7 +46,9 @@ export default {
     return resources.map((r) => this.formatSplitResponse(r));
   },
 
-  fetch(url, action, multiple, ...params) {
+  fetch(url, _action, multiple, ...params) {
+    // Set a default noop action handler
+    const action = typeof _action === 'function' ? _action : () => {};
```
msecret: Could this be dangerous because if a dev forgets to pass in an action, they won't know why something isn't working?

static_src/actions/space_actions.js
adborden: My opinion is that the fetch action should also handle the receive action as the promise resolves. The alternative would be that the UI component calling the action would have to call the receive which feels too boilerplate to me. It's also unnecessary, why would a caller want to fetch and not call the receive? Rather than have purely dumb actions, I think this fetch/receive pattern works well.

adborden: I just updated the list of guidelines to include this opinion. You can see this in action in the [toggleQuicklook](https://github.com/18F/cg-dashboard/pull/935/files#diff-5700de5f748b06353d227e8f6a17417aR76) action. `toggleQuicklook` is like a fetch action (because the toggle needs to fetch spaces for the org) and when the request returns, the appropriate `success`/`error` action is called based on how the promise resolves.

static_src/util/cf_api.js
adborden: Sure, I wouldn't say "dangerous" but yes, there would be no feedback if you didn't include the action callback. This is just like with promises, you wouldn't get any feedback if you forgot to call `.this` on your promise.

I see this method as going away. It's useful to have the fetch utilities to handle multiple and paged responses, but mostly I think we should just use the request promise for handling any followup action rather than in this utility. So the intention is to move to a model where there is less feedback for developer errors -- a promise-based model which I think is familiar to javascript developers.

static_src/actions/space_actions.js
msecret: I thought the alternative would be the cfAPI calling the `receivedSpace` action which is happening now, but could make more sense to move them all to actions.

adborden: I see. I think that by putting the `receivedSpace` action in the cfApi, we open ourselves up to some action mismatches. It becomes possible to want to fetch from the space API, which would trigger the `RECEIVE`, but it's on the api caller to call the `FETCH` leading to possible `RECEIVE` without the initial `FETCH`.

By moving it all to the action, the `FETCH`, `RECEIVE`, and `ERROR` would all be located in the same action creation and decoupled from the actual cfApi call. e.g.
```javascript
function fetchAction() {
  dispatch(FETCH)
  cfApi.fetch().then(() => dispatch(RECEIVE)).catch(() => dispatch(ERROR)).
}
```
I think this makes it much clearer who is responsible for what:
- cfApi: handles API requests and data parsing/formatting
- actions: handle the async flow of cfApi and encapsulate the dispatch of action events
- stores: synchronously updates the loading state based on the action events.

msecret: That sounds fine then.

PR #942, opened 2017-01-09
Fixed content inconsistencies and wrong spaces

msecret:


static_src/components/breadcrumbs.jsx
```diff
@@ -51,7 +51,7 @@ export default class Breadcrumbs extends React.Component {
 
     // Add the parent states
     if (org && org.name) {
-      breadcrumbs.push(<BreadcrumbsItem key="home" url="/#/">overview</BreadcrumbsItem>);
+      breadcrumbs.push(<BreadcrumbsItem key="home" url="/#/">Overview</BreadcrumbsItem>);
```
adborden: I think we want this lowercase since orgs/spaces/apps are generally lowercase. Otherwise it will stand out too much.

thisisdano: I see Britta's point (consistency) but I definitely agree with Aaron here.

brittag: Hm, I found lowercase a teeny bit confusing while browsing around as a user - to me, "Overview" is part of the UI, a distinct thing from orgs/spaces/apps, so it feels right for it to be consistent with the other place I saw this term, rather than consistent with org/space/app names. Defer to you all though!

adborden: I just wanted to make sure this change was intentional because at one point it was explicitly lowercased. I'm a fan of consistency and this looks okay to me.

screenshot from 2017-01-12 15-23-43

PR #944, opened 2017-01-10
Adds UI notificaitons for service form

msecret:
To ensure user knows when the instance is being created and after it
was successfully created.

I decided on a simple timeout in the service instance store so it
only sets certain states after a certain amount of time. This ensures
minimal changes to how the components work. If the timeout was in
the jsx code, it would be complicated to figure out how to configure
the jsx code to wait to remove the instance form component.

### DIFF
--- a/static_src/stores/service_instance_store.js
+++ b/static_src/stores/service_instance_store.js
@@ -172,7 +172,7 @@ class ServiceInstanceStore extends BaseStore {
         this._createdTempNotification = true;
         this.emitChange();
         setTimeout(() => {
-          this._createInstanceForm = null
+          this._createInstanceForm = null;
           this._createdTempNotification = false;
           this.emitChange();
         }, CREATED_NOTIFICATION_TIME_MS);
--- a/static_src/stores/service_instance_store.js
+++ b/static_src/stores/service_instance_store.js
@@ -17,6 +17,7 @@ export const OPERATION_DELETING = 'deleting';
 const OPERATION_PROCESSING = 'processing';
 export const OPERATION_RUNNING = 'running';
 const OPERATION_INACTIVE = 'inactive';
+export const CREATED_NOTIFICATION_TIME_MS = 2000;
 
 const OPERATION_STATES = {};
 OPERATION_STATES[OPERATION_FAILED] = 'Failed';
@@ -41,6 +42,8 @@ class ServiceInstanceStore extends BaseStore {
     this._data = new Immutable.List();
     this._createInstanceForm = null;
     this._createError = null;
+    this._createLoading = false;
+    this._createdTempNotification = false;
   }
 
   get createInstanceForm() {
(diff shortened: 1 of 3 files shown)


static_src/stores/service_instance_store.js
```diff
@@ -150,18 +161,27 @@ class ServiceInstanceStore extends BaseStore {
...
         cfApi.fetchServiceInstance(action.serviceInstance.guid);
-        this._createInstanceForm = null;
+        this._createLoading = false;
+        this._createdTempNotification = true;
         this.emitChange();
+        setTimeout(() => {
+          this._createInstanceForm = null;
+          this._createdTempNotification = false;
+          this.emitChange();
+        }, CREATED_NOTIFICATION_TIME_MS);
```
adborden: I was thinking about this kind of time-based UI change when working on the quicklook loading and FWIW I think this should be done in the action. It would look something like this:
```javascript
instanceCreate(serviceData) {
  // dispatch SERVICE_INSTANCE_CREATE
  return cfApi
    .createServiceInstance(serviceData)
    .then((guid) => cfApi.fetchServiceInstance(guid).then(this.instanceCreated));
}

instanceCreated(instance) {
  // dispatch SERVICE_INSTANCE_CREATED
  setTimeout(/* dispatch SERVICE_INSTANCE_NOTIFICATION_CLEAR */);
}
```

What do you think?

static_src/stores/service_instance_store.js
```diff
@@ -150,18 +161,27 @@ class ServiceInstanceStore extends BaseStore {
...
+        this._createLoading = true;
+        this.emitChange();
         break;
       }
 
       case serviceActionTypes.SERVICE_INSTANCE_CREATED: {
         cfApi.fetchServiceInstance(action.serviceInstance.guid);
-        this._createInstanceForm = null;
+        this._createLoading = false;
+        this._createdTempNotification = true;
```
adborden: I think you want to clear the error on success here, just in case there was a previous failure we were displaying. Although not totally sure if it should be done on the CREATE action...

static_src/stores/service_instance_store.js
```diff
@@ -17,6 +17,7 @@ export const OPERATION_DELETING = 'deleting';
 const OPERATION_PROCESSING = 'processing';
 export const OPERATION_RUNNING = 'running';
 const OPERATION_INACTIVE = 'inactive';
+export const CREATED_NOTIFICATION_TIME_MS = 2000;
```
adborden: FWIW, I think 2 seconds is a bit short. Not sure if we have better guidance around this, but I think 3-5 seconds would be better.

msecret: I took guidance from design, but can move to 3 seconds.

static_src/stores/service_instance_store.js
msecret: I was hesitant to move things to actions in this PR before I see more of how you're doing that. Well, was more thinking what our plan should be there; should we move all our cf api calls to actions or move then one at a time.

I was researching this yesterday and I saw an article that said either call the API from your stores or your actions but not both, which made me hesitant to mix and match.

adborden: I think we're in agreement that we should move all the cfApi calls to the actions.

PR #945, opened 2017-01-12
New method for testing data.

msecret:
Uses more real data for fake testing data. Removes some information from this data and adds in fixtures files consistently created an labeled.

TODO still, not on this PR:
- Get paging working, such as the `events` endpoint, which has many pages that the code pages over.
- Requests other the GET to work. Might have to modify the data to get this working as desired, and ensure we return the correct things.


adborden:
:+1:

PR #957, opened 2017-01-13
Work-around Session Filestore fallback bug

adborden:
gorilla Session creation isn't handled correctly when MaxAge is unset. Work
around upstream bug[1] by explicitly setting MaxAge.

[1] https://github.com/gorilla/sessions/issues/96


jmcarp:
Fixed go fmt.

PR #958, opened 2017-01-14
Release 1.13

adborden:
- Fix issue with the CSS prefixer
- Build staging as we do production to catch any prefixer issues

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -45,6 +45,7 @@ deployment:
     branch: [master, demo, deprecated]
     owner: 18F
     commands:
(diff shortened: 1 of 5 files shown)


thisisdano:
Any reason to prefix at all at this point? What additional benefit does it provide?

adborden:
@thisisdano  I'm not sure if it's giving us any current benefits, I was under the impression that it helped to avoid pulling in styles from uswds. In the future, it will definitely help to make it easier to customize the dashboard for non-cloud.gov instances.

thisisdano:
aha! Interesting...

Issue #947, opened 2017-01-12
Setup a functional test suite using the fake data testing server


adborden:
Once we have fake data, we can create automated functional UI tests. We want to have a well-written example test so that there is little friction to adding new functional tests in the future.

- [x] Remove the old, broken go functional tests
- [x] npm task to run the functional tests separate from the unit tests
- [x] Implement test scenario for editing application usage

adborden:
This is related to #932

adborden:
Added tests for clicking through the org quicklook.

PR #971, opened 2017-01-24
Browser-based functional tests using webdriverio

adborden:
Part of https://github.com/18F/cg-dashboard/issues/947


package.json
```diff
@@ -13,10 +13,13 @@
     "check-style": "/bin/bash -c '[[ $(npm ls cloudgov-style) =~ \"git://github.com\" ]] && npm run build-style || exit 0'",
     "clean": "rm -rf ./static/assets/*",
     "lint": "eslint --ext .js --ext .jsx ./static_src",
-    "test": "export NODE_ENV=test && npm run lint && karma start --single-run",
+    "test": "npm run lint && npm run test-karma && npm run test-functional",
+    "test-functional": "NODE_ENV=test wdio wdio.conf.js",
+    "test-karma": "NODE_ENV=test karma start --single-run",
```
msecret: Can we just call this `test-unit`?

static_src/test/functional/overview_quicklook.spec.js
```diff
@@ -0,0 +1,50 @@
...
+  it('navigates to page', function () {
+    browser.url('/');
+  });
+
+  it('has a title', function () {
+    expect(browser.getTitle()).toBe('cloud.gov dashboard');
+  });
+
+  it('has a page header', function () {
+    const pageHeader = browser.element('.page-header-title');
```
msecret: I have thoughts about how we should target elements for these tests. I wanted to prefix classes for tests with `test-` so that it's very clear to a developer that the element they're working with has a hook to functional tests. I found this greatly reduces regression in functional tests from HTML/element changes.

msecret: We don't need to address this now, we could just discuss at some point and write some documentation on how we want to target elements from tests.

adborden: Okay, so in this case, would I add an additional class, `.test-page-header-title` to the page header?

msecret: yes

PR #977, opened 2017-01-24
App start button

msecret:
This adds a start button to the app page when the app isn't started yet, rather then just the restart button.

Uses the new method of moving cf api calls to the actions code. In this PR, it makes it easier to coordinate various actions calling each other but not dispatching an event until everything makes sense.

This code fixes a small UI issue which existed before, and is not required for code to be merged: https://github.com/18F/cg-style/pull/290.

### DIFF
--- a/static_src/stores/app_store.js
+++ b/static_src/stores/app_store.js
@@ -107,6 +107,18 @@ class AppStore extends BaseStore {
         break;
       }
 
+      case appActionTypes.APP_START: {
+        const app = this.get(action.appGuid);
+        if (app) {
+          const startingApp = Object.assign({}, app,
+            { state: appStates.starting });
+          this.merge('guid', startingApp, (changed) => {
+            if (changed) this.emitChange();
+          });
+        }
+        break;
+      }
+
       case appActionTypes.APP_RESTART: {
         const app = this.get(action.appGuid);
         if (app) {
--- a/static_src/stores/app_store.js
+++ b/static_src/stores/app_store.js
@@ -108,7 +108,6 @@ class AppStore extends BaseStore {
       }
 
       case appActionTypes.APP_RESTART: {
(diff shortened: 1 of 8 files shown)


static_src/stores/app_store.js
```diff
@@ -107,8 +110,19 @@ class AppStore extends BaseStore {
         break;
       }
 
+      case appActionTypes.APP_START: {
+        const app = this.get(action.appGuid);
+        if (app) {
+          const startingApp = Object.assign({}, app,
+            { state: appStates.starting });
```
adborden: btw, since we're using ES6, this could be written as `startingApp = { ...app, state: appStates.starting }`, do you have an opinion on if we want to use the `Object.assign` syntax or spread operator?

static_src/stores/app_store.js
```diff
@@ -107,8 +110,19 @@ class AppStore extends BaseStore {
...
+          this.merge('guid', startingApp, (changed) => {
+            if (changed) this.emitChange();
+          });
```
adborden: The callback shouldn't be necessary, `merge` uses the [defaultChangedCallback](https://github.com/18F/cg-dashboard/blob/master/static_src/stores/base_store.js#L13) which handles this for you.

static_src/test/unit/stores/app_store.spec.js
```diff
@@ -267,43 +280,17 @@ describe('AppStore', function() {
...
-
-      appActions.restarted('zxlcvkjklv');
-
-      expect(spy).toHaveBeenCalledOnce();
-      let args = spy.getCall(0).args;
-      expect(args[0]).toBeFunction();
-      expect(args[0](expectedRes)).toBeTruthy();
-      expect(args[1]).toBeFunction();
-    });
+    // TODO still needed?
```
adborden: I think this `describe` can be removed.

PR #978, opened 2017-01-25
Fix show more activity button

msecret:
So it doesn't show when there are less activity log items then the
max to show.


static_src/components/activity_log.jsx
```diff
@@ -77,6 +77,12 @@ export default class ActivityLog extends React.Component {
     if (this.state.empty) {
       content = <h5 className="test-none_message">No recent activity</h5>;
     } else {
+      let showMore = (this.state.activity.length >= this.props.maxItems) &&
```
adborden: This should probably be `this.state.maxItems`. I think it's possible for `props` and `state` to be out of sync, which could cause subtle rendering bugs down the road.

PR #980, opened 2017-01-25
Open app button

msecret:
Adds a button to top of app page that opens the first bound route or doesn't show if there are no bound routes.

### DIFF
--- a/static_src/components/app_container.jsx
+++ b/static_src/components/app_container.jsx
@@ -8,6 +8,7 @@ import { appHealth, worstAppInstanceState } from '../util/health';
 import { appStates } from '../constants';
 import AppStore from '../stores/app_store.js';
 import Breadcrumbs from './breadcrumbs.jsx';
+import RouteStore from '../stores/route_store.js';
 import EntityIcon from './entity_icon.jsx';
 import ErrorMessage from './error_message.jsx';
 import Loading from './loading.jsx';
@@ -110,15 +111,18 @@ export default class AppContainer extends React.Component {
 
   get openApp() {
     if (AppStore.isRunning(this.state.app)) {
+      const route = RouteStore.getRouteURLForApp(this.state.app);
       return (
-        <Action
-          style="primary"
-          href=""
-          label="open app"
-          type="outline"
-        >
-          <span>Open app</span>
-        </Action>
+        <div>
+          <Action
+            style="primary"
+            href={ route }
+            label="open app"
+            type="outline"
+          >
+            <span>Open app</span>
+          </Action>
+        </div>
       );
     }
     return null;
--- a/static_src/components/app_container.jsx
+++ b/static_src/components/app_container.jsx
@@ -113,21 +113,19 @@ export default class AppContainer extends React.Component {
 
   get openApp() {
     const route = RouteStore.getRouteURLForApp(this.state.app);
-    if (route) {
-      return (
-        <div>
-          <Action
-            style="primary"
-            href={ `https://${route}` }
-            label="open app"
-            type="outline"
-          >
+    if (!route) return null;
+    return (
+      <div>
+        <Action
+          style="primary"
+          href={ `https://${route}` }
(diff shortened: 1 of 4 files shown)


static_src/stores/route_store.js
```diff
@@ -40,6 +42,26 @@ class RouteStore extends BaseStore {
     return !!route.app_guid;
   }
 
+  getRouteURLForApp(app) {
+    const allRoutes = this.getAll();
+    const boundRoutes = allRoutes.filter((route) => route.app_guid === app.guid);
```
adborden: Hmm, how does this work if you have multiple apps bound to the same route?

static_src/test/unit/stores/route_store.spec.js
```diff
@@ -29,6 +30,40 @@ describe('RouteStore', function() {
     });
   });
 
+  describe('gerRouteURLForApp()', function() {
```
adborden: typo, `getRouteURLForApp`

static_src/test/unit/stores/route_store.spec.js
```diff
@@ -29,6 +30,40 @@ describe('RouteStore', function() {
...
+    it('should a full url for a route with no domain', function() {
```
adborden: "should be"?

static_src/stores/route_store.js
msecret: I thought we found that was not possible at some point.

msecret: Yea we did, it would throw and error.

static_src/components/app_container.jsx
```diff
@@ -108,6 +111,28 @@ export default class AppContainer extends React.Component {
     );
   }
 
+  get openApp() {
+    if (AppStore.isRunning(this.state.app)) {
```
msecret: Dont' we only want the button if the app is actually running? Otherwise we'd be sending a user to a broken URL?

static_src/stores/route_store.js
adborden: To bind two apps to a single route? Yeah, that is possible. That's how the blue/green [autopilot cf plugin](https://github.com/contraband/autopilot#method) works.

I currently have this setup in case you want to test. https://adborden-node.apps.cloud.gov/ routes to `adborden-node` and `adborden-nodejs-venerable` in the cf-deck-testing space.

static_src/components/app_container.jsx
adborden: Assuming you had [two apps bound to the same route](https://github.com/18F/cg-dashboard/pull/980#discussion_r98577410), one might not be running but it could still make sense for the link to appear. I think we should just remove the check and worst case no apps are running and you just get a 404 from Cloud Foundry. Seems reasonable to me.

static_src/stores/route_store.js
msecret: I checked out your two apps, they both still have the correct open app route.

msecret: I think it's because we set the `app_guid` when the request is made on the app page.

adborden: Probably something we will want to fix down the line, we'll get bit by this eventually if we're not careful about how we're loading the routes.

static_src/components/app_container.jsx
```diff
@@ -108,6 +111,25 @@ export default class AppContainer extends React.Component {
...
+            href={ `https://${route}` }
+            label="open app"
+            type="outline"
+          >
+          <span>Open app</span>
+        </Action>
+      </div>
+      );
+    }
+    return null;
```
adborden: Sorry, one more ask, can you reverse the `if` condition? I prefer the fail-fast because it keeps the indentation flat which improves legibility, and it keeps all the checks at the top of the function.
```javascript
get openApp() {
 const route = //...
  if (!route) {
    return null;
  }

  return <Action />;
}
```

PR #987, opened 2017-01-27
Add PageStore to handle initial data loading

adborden:
Fixes https://github.com/18F/cg-dashboard/issues/869

The problem is that when we cascade fetches (fetch all spaces, then fetch individual space details), there is a small gap between these where the state of the stores is not loading, so the loading indicator disappears, then a moment later re-appears. By explicitly controlling the state from the route via a new `PageStore` and `pageActions`, we can be sure the loading state doesn't change until we really want  it to.

I didn't add this to all the routes yet, wanted to get this bug fixed first.

### DIFF
--- a/static_src/actions/page_actions.js
+++ b/static_src/actions/page_actions.js
@@ -1,0 +1,29 @@
+
+import AppDispatcher from '../dispatcher.js';
+import { pageActionTypes } from '../constants.js';
+
+export default {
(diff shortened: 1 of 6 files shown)


adborden:
Yeah, I think eventually all the routes should end up using it. It would also give us better control for resetting state on page transitions.

PR #990, opened 2017-01-31
Panel refactor documentation and actions

msecret:
Refactors all the panel code to use the correct panel actions and panel documentation. This ensures consistency in visual style and spacing.

At this point, it does break some layout, so we can wait until those things are fixed to merge.

Requires https://github.com/18F/cg-style/pull/293


msecret:
I fixed this code so it doesn't cause any major visual regressions. Only visual differences I've seen are some content in the panel rows are differently placed before, which will be fixed by the next refactor PRs. I did this so this could be safely merged.

PR #993, opened 2017-02-01
Panel refactor rows

msecret:
First pass at refactoring panel rows by trying it on space quick look (which has nested list) and routes panel (which has headers).

Requires https://github.com/18F/cg-style/pull/294.


msecret:
Currently, the main thing left to fix is the complex usage and limits UI, which will likely require another component.

PR #994, opened 2017-02-02
Add error action for activity logs fetch

adborden:
Adds an action for log fetch failure, so that the app can display a better message than just "No recent activity".

I realized that the current `handleError` in `cf_api` doesn't work well for promises, so I created a new simpler one. I expect the existing `handleError` to disappear once all the actions are removed from `cf_api`.

### DIFF
--- a/static_src/components/activity_log.jsx
+++ b/static_src/components/activity_log.jsx
@@ -28,7 +28,9 @@ function stateSetter(props) {
 
   return {
     activity,
-    empty: (ActivityStore.fetched && activity.length === 0)
+    empty: (ActivityStore.fetched && activity.length === 0),
+    hasErrors: ActivityStore.hasErrors,
+    errors: ActivityStore.errors
   };
 }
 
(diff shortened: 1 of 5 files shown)


static_src/components/activity_log.jsx
```diff
@@ -74,7 +76,9 @@ export default class ActivityLog extends React.Component {
   render() {
     let content = <div></div>;
 
-    if (this.state.empty) {
+    if (this.state.empty && this.state.hasErrors) {
+      content = <h5 className="test-none_message">An error occured fetching recent activity</h5>;
```
msecret: In the current case of the activity log, an error does occur, but some activity items are still fetched (at least thats what appears to happen in my UI). Would it be better to display any items that we have and the error, or should we just display the error?

I'm worried the UI will actually get worse if we just show the error, because this error seems to happen a lot on the app page (we might want to figure out why), so we could be making the activity log unusable.

adborden: > In the current case of the activity log, an error does occur, but some activity items are still fetched (at least thats what appears to happen in my UI).

Yes, this is definitely the case. Events and logs are fetched separately so, it is possible one succeeds and the other fails (it's usually the logs that fail). I think what I've implemented addresses your concerns about UI clutter. The main use case I am aiming for is when there is no activity to show **and** there was an error. Previously in that case we were showing "No recent activity", which feels really misleading to me when there's actually an error.

I didn't want to get too complicated where one succeeds with data and then we would be showing data + error message so, in that case, we would show the data we have and hide any error (unless you're seeing different and that would be a bug).

I think this solution is actually the cleanest without any significant design changes. Does that make sense?

msecret: Yes I see that now, that makes sense to me from a UI perspective.

Issue #973, opened 2017-01-24
Move API calls to actions and out of the stores


adborden:
As per https://github.com/18F/cg-dashboard/pull/935#issue-199339263, we should be consistent about using the API in the actions and not the stores. These are the principles:

- Each action creator (each method in `*_actions.js`)  should dispatch only
  a single action.
- Any XHR calls should happen within the fetch action creators which
  should return a promise of the request.
- Any `fetch` action should call the appropriate `receive`/`success`/`error` action on the resolve of the fetch's promise.
- Each async action should have a fetch, success, and error sub-action (e.g.
  `TOGGLE`, `TOGGLE_SUCCESS`, and `TOGGLE_ERROR`).

- [x] activity_store
- [x] app_store
- [x] org_store
- [x] service_store
- [x] service_instance_store
- [x] service_binding_store
- [ ] route_store
- [ ] domain_store
- [ ] quota_store
- [x] service_plan_store
- [ ] user_store

PR #995, opened 2017-02-02
Move api calls from app store to app action

adborden:
Part of [Navigator 2370](https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-2370) (and https://github.com/18F/cg-dashboard/issues/973)

### DIFF
--- a/static_src/actions/app_actions.js
+++ b/static_src/actions/app_actions.js
@@ -9,14 +9,14 @@ import { appActionTypes } from '../constants';
 import cfApi from '../util/cf_api.js';
 import poll from '../util/poll.js';
 
-export default {
+const appActions = {
   fetch(appGuid) {
     AppDispatcher.handleViewAction({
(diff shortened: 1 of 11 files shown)


static_src/actions/app_actions.js
```diff
@@ -54,27 +74,35 @@ export default {
       appGuid,
       app
     });
+
+    return Promise.resolve(app);
```
msecret: what's the pattern of what promise data should be returned in actions? So if `receivedAppStats` just took an `appGuid` should that be returned or nothing? It'd be helpful to know the convention.

adborden: There's no hard rule. For fetch actions, it should be the entity that is being fetched. For receive actions, it should be the input so that it is chainable. The exception is error actions, I think it's best to return void here although I haven't made a final opinion on it.

PR #997, opened 2017-02-06
Refactor service actions

adborden:
Part of https://github.com/18F/cg-dashboard/issues/973

Refactors services, service instances, service plans, and service bindings.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -154,6 +154,30 @@ To update a single package to a specific version
     to do so.
 - CSS in component files should opt to use CSS Modules first, and will use normal CSS class architecture after.
 
+### Working with React/Flux
+
+
+#### Action creators
+
+- Each action creator (each method in `*_actions.js`)  should dispatch only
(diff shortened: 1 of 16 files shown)


static_src/actions/org_actions.js
```diff
@@ -5,61 +5,72 @@
...
+    return Promise.resolve(orgGuid);
   },
 
   fetch(orgGuid) {
     AppDispatcher.handleViewAction({
       type: orgActionTypes.ORG_FETCH,
       orgGuid
     });
+
+    return cfApi.fetchOrg(orgGuid).then(orgActions.receivedOrg);
```
msecret: Shouldn't `orgActions` be `this`?

adborden: I'm trying to avoid using `this` in the actions because it's possible the action will get passed as a function (like above) and when called, won't be bound to `this`. Explicitly referring to `orgActions` here means we never have to worry about `this` being unbound.

PR #998, opened 2017-02-06
Refactor usage and limits

msecret:
Changes stat to new more constrained layout. Required some re-org
of thes stat component, like moving the title of it inside
the component to set it as a certain width.

Also fixed react error when ResourceUsage was passing a string to
secondaryInfo prop of Stat, which should be an Element.

Requires 18F/cg-style#ms-panel_refactor_usage_limits.

### DIFF
--- a/static_src/components/resource_usage.jsx
+++ b/static_src/components/resource_usage.jsx
@@ -56,6 +56,7 @@ export default class ResourceUsage extends React.Component {
       };
     } else {
       properties = { ...properties,
+        primaryStat: props.amountTotal,
         editable: props.editable,
         max: props.max,
         min: props.min,
--- a/static_src/components/resource_usage.jsx
+++ b/static_src/components/resource_usage.jsx
@@ -44,36 +44,31 @@ export default class ResourceUsage extends React.Component {
 
   render() {
     const props = this.props;
-    let title = <h2 className={ this.styler('stat-header')}>{ props.title }</h2>;
     let stat = (
       <Stat
         editable={ props.editable }
         max={ props.max }
         min={ props.min }
         onChange={ props.onChange }
+        title={ props.title }
         name={ props.name }
         primaryStat={ props.amountTotal }
-        secondaryInfo={ props.secondaryInfo }
+        secondaryInfo={ <span>{ props.secondaryInfo }</span> }
       />
     );
 
     if (props.amountUsed && props.amountTotal) {
-      title = <h2 className={ this.styler('stat-header')}>{ props.title }</h2>;
       stat = (
         <Stat
           primaryStat={ props.amountUsed }
+          title={ props.title }
           secondaryInfo={ <span>{this.available()} available</span> }
           statState={ this.statState(props.amountUsed, props.amountTotal) }
         />
       );
     }
 
-    return (
-      <div style={{ width: '100%' }}>
-        { title }
-        { stat }
-      </div>
-    );
+    return stat;
   }
 }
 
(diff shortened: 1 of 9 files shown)


msecret:
screen shot 2017-02-06 at 3 07 31 pm

static_src/test/functional/overview_quicklook.spec.js
```diff
@@ -43,7 +43,9 @@ describe('Overview page', function () {
     });
 
     it('is expanded', function () {
-      expect(quicklookElement.isExpanded()).toBe(true);
+      browser.waitUntil(function() {
+        return quicklookElement.isExpanded();
+      }, 2000, 'quicklook does not expand');
```
adborden: We should add an `expand()` method to the `quicklook.element` that calls `click` and then does the wait. This kind of stuff should be encapsulated in the page objects since the wait will be needed anytime you want to trigger the click in a test.

msecret: So would the `expand` return a promise? And then you'd run the test after the promise completes?

adborden: webdriver.io is all sync anyway so you don't need to return a promise. Basically you'll move this `waitUntil` into a new `expand` method in `quicklook.element` and then on [line 42](https://github.com/18F/cg-dashboard/pull/998/commits/9d76f5fe0bb90dd06b7b95a5b51fdf2a51f2d27d#diff-1a9224665779b2b0186cf33b4f843d2dR42) you'll call `expand` instead of `click`. That click test block will handle the waiting. Does that make sense?

static_src/components/resource_usage.jsx
```diff
@@ -44,36 +44,31 @@ export default class ResourceUsage extends React.Component {
...
       <Stat
         editable={ props.editable }
         max={ props.max }
         min={ props.min }
         onChange={ props.onChange }
+        title={ props.title }
         name={ props.name }
         primaryStat={ props.amountTotal }
-        secondaryInfo={ props.secondaryInfo }
+        secondaryInfo={ <span>{ props.secondaryInfo }</span> }
```
adborden: Seems strange that the type for `secondaryInfo` on `ResourceUsage` is `string`, but for `Stat` it's `element`. We should change both to `node` which would be [any renderable type](https://facebook.github.io/react/docs/react-api.html#react.proptypes.node) including both strings and elements.

static_src/components/resource_usage.jsx
```diff
@@ -44,36 +44,31 @@ export default class ResourceUsage extends React.Component {
...
       );
     }
 
-    return (
-      <div style={{ width: '100%' }}>
-        { title }
-        { stat }
-      </div>
-    );
+    return stat;
```
adborden: This makes ResourceUsage pretty thin, I think we can clean this up by using a props object like
```javascript
let properties = { /* common properties */ }
if (props.amountUsed && props.amountTotal) {
    properties = { ...properties, /* additional/override properties */ }
}

return <Stat { ...properties } />;
```

static_src/components/resource_usage.jsx
```diff
@@ -44,31 +44,27 @@ export default class ResourceUsage extends React.Component {
...
+        statState: this.statState(props.amountUsed, props.amountTotal)
+      };
+    } else {
+      properties = { ...properties,
+        editable: props.editable,
+        max: props.max,
+        min: props.min,
+        onChange: props.onChange,
+        name: props.name
+      };
```
adborden: Hope my suggestion didn't make this more confusing! `primaryStat` and `secondaryInfo` should be overridden in this case, they're different depending on whether both amountUsage and amountTotal are available or not.

PR #1004, opened 2017-02-08
Fix missing spaces

msecret:
Fixes bugs where users with admin permissions (users who can see all orgs) were not able to see certain spaces.

Bug was due to how data from the promise code was being put together. It was only adding
in the data from the last two requests rather then all of them. I also wrote a test case for this to ensure it doesn't happen again.

Refs https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-2876

### DIFF
--- a/static_src/util/cf_api.js
+++ b/static_src/util/cf_api.js
@@ -115,8 +115,8 @@ export default {
       const reqs = urls.map((u) => http.get(u).then((r) => r.data.resources));
 
       return Promise.all(reqs)
-        .then((all) => all.pop())
-        .then((all) => [].concat.call([], res.data.resources, all))
+        .then((all) => [].concat.apply([], all))
+        .then((all) => res.data.resources.concat(all))
         .then((all) => action(this.formatSplitResponses(all), ...params))
         .catch((err) => handleError(err));
     });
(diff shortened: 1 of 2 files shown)


static_src/util/cf_api.js
```diff
@@ -115,8 +115,8 @@ export default {
       const reqs = urls.map((u) => http.get(u).then((r) => r.data.resources));
 
       return Promise.all(reqs)
-        .then((all) => all.pop())
-        .then((all) => [].concat.call([], res.data.resources, all))
+        .then((all) => [].concat.apply([], all))
```
adborden: This can just be `[].concat(res.data.resources, all)`, otherwise you're creating two empty arrays and throwing one away, right?

static_src/util/cf_api.js
```diff
@@ -115,8 +115,8 @@ export default {
...
+        .then((all) => res.data.resources.concat(all))
```
adborden: Why were we popping before? are we requesting n+1 items?

static_src/util/cf_api.js
msecret: Correct, I think I was just being careful yesterday to ensure it was fixed.

static_src/util/cf_api.js
msecret: I'm not sure

PR #1007, opened 2017-02-09
Fix layout broken on prod build

msecret:
In order to build the nested complex lists, I was using the `name`
property of the React component's `type` field. I though this was
a React defined property, but it might actually be the function
name property from standard JS. This `name` property seems to change
in the production build when build with Uglify.

This change uses the standard way of checking type as I've read about
in various places.


adborden:
FWIW, ES6 defines a `name` property for classes, although transpilers might have some bugs around this. I know that I was trying to assign a name to a constructor with babel and babel would throw an error.

adborden:
We should apply this to the release branch.

PR #1009, opened 2017-02-10
Reduce number of listeners on ServiceInstanceStore

adborden:
Adds a warning when we are getting close to maxing out the number of listeners on a store. Currently `DomainStore` and `RouteStore` also have listeners within `ActivityLogItem` and could be hoisted up in the same way.

### DIFF
--- a/static_src/stores/base_store.js
+++ b/static_src/stores/base_store.js
@@ -9,6 +9,10 @@ import Immutable from 'immutable';
 import AppDispatcher from '../dispatcher.js';
 import LoadingStatus from '../util/loading_status.js';
 
+// TODO clean up the listeners and reduce the MAX_LISTENERS to 15
+const MAX_LISTENERS = 20;
+const MAX_LISTENERS_THRESHOLD = 10;
+
 function defaultChangedCallback(changed) {
   if (changed) this.emitChange();
 }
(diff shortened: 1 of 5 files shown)


static_src/stores/base_store.js
```diff
@@ -9,6 +9,10 @@ import Immutable from 'immutable';
 import AppDispatcher from '../dispatcher.js';
 import LoadingStatus from '../util/loading_status.js';
 
+// TODO clean up the listeners and reduce the MAX_LISTENERS to 15
+const MAX_LISTENERS = 20;
+const MAX_LISTENERS_THRESHOLD = 10;
```
msecret: Why should 10 me the threshold? It seems like the max is there to catch possible leaks, but we're knowingly using this man, so I'd expect to not be warned for it. Or is there other problems with having so many listeners?

adborden: You mean why not set an unlimited number of listeners? In this case, we have a dynamic number of ActivityLogItems on the page, so it seems strange to me that the number of listeners explodes on the app page. That's the scenario I'm trying to catch, but you're right that it's not a leak so why should we care?

It feels anti-flux to me, since the container component's are responsible to listening to state changes and passing that down the component hierarchy. I think it's reasonable that each panel acts as a component and would each could have it's own listeners but I don't think smaller components should listen to the stores.

msecret: Yea that's the pattern I've been going for recently, there are some old small components that don't follow it.

adborden:
Rebased; added a commit to remove the other stores out of ActivityLogItem. @msecret can you take a look at the last commit?

PR #1011, opened 2017-02-10
Fixes route form cancel button refreshing page

msecret:
It seems liek React requires the action to be blank in this
case. I tried other things, this was only thing that got it
to stop refreshing.

Refs https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-2494

### DIFF
--- a/static_src/components/action.jsx
+++ b/static_src/components/action.jsx
@@ -90,7 +90,7 @@ export default class Action extends React.Component {
           aria-label={ this.props.label }
           onClick={ (ev) => this.props.clickHandler(ev) }
           disabled={this.props.disabled}
-          type={ this.props.type }
+          type={ this.props.type === 'submit' ? this.props.type : null }
         >
           { this.props.children }
         </button>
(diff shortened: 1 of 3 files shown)


adborden:
Hmm, yeah very strange. The whole form is a little weird and generates some warnings and errors. The things that stick out to me most is:

1. we're using `type` on the cancel `button` for styling, I wonder if the browser is interpreting that as `type="submit"`?
1. we attach the `_onSubmit` to both the submit `button` and the `form`
1. React complains the form starts as uncontrolled, and then becomes a controlled form

I can't even seem to reproduce this anymore... but I still get warnings.

msecret:
I didn't try messing with the button inputs, I'll try that.

I also never saw the React warning but probably wasn't paying attention.

msecret:
I tried ensuring the buttons dont' have submit type. I also tried playing around with the form so we're only binding the `onSubmit` method once. None of it got the form to stop submitting when clicking the cancel button.

Issue #953, opened 2017-01-12
Rethink and implement health states for application activity log


adborden:
From the [app walkthrough](https://docs.google.com/document/d/1Hx3QXfkmxdpX-rPxu3XNp2VRuGc8BaNryzshE43dPu0/edit) we identified some issues around the application activity log component.

### [3] Display more information/change states
Updating an app appears as a warning (gold). Maybe this shouldn’t be the case
Change messages so they’re more descriptive of what the action is

- [x] Only errors should have the status-pill
- [x] No status-pill for any other events (start/update), no ok or warning state at all

### [2] Not clear that activity entries are clickable and expandable

- [x] Entire row should be clickable for the expand target
- [x] Row should have better indication that it is clickable, make the hover cursor a pointer

### [2] Change language from ambiguous “updating” to something like modified the app resources

- [x] For the update event, change copy to "modified the application's resource allocation"

### [2] Add label describing the raw data upon expansion: “Event log”

- [x] When log event is expanded, add copy: "Raw event log"

### [3] “Show more activity” action

It does not do anything, should fetch the next set of events, should not be displayed if there are no longer any events to fetch (addressed in https://github.com/18F/cg-dashboard/issues/949)

- [x] “Show more activity” button should be outline style.

adborden:
For the clickable indicator, not sure if we just need to change the pointer or if we need to add some kind of chevron/arrow.

msecret:
Checking this out on demo (which has activity-log-ux branch), I found certain log entries are off in their left alignment. See screenshot.

Is that just from cg-style differences, or is it a problem?

msecret:
Checked it out on staging, the UI thing isn't there.

PR #1013, opened 2017-02-11
Activity log UX fixes

adborden:
Fixes https://github.com/18F/cg-dashboard/issues/953

Depends on https://github.com/18F/cg-style/pull/302

screenshot from 2017-02-10 18-46-36

### DIFF
--- a/static_src/util/cf_api.js
+++ b/static_src/util/cf_api.js
@@ -100,25 +100,37 @@ export default {
     return this.fetch(url, action, true, ...params);
   },
 
-  fetchAllPages(url, action, ...params) {
-    return http.get(APIV + url).then((res) => {
-      const urls = [];
+  // fetchAllPages(url, data = {}, action = () => {})
+  fetchAllPages(url, ...args) {
+    let [data, action] = args;
+    if (typeof data === 'function') {
+      action = data;
+      data = {};
+    }
+
+    const path = `${APIV}${url}`;
+    return http.get(path, { params: data }).then((res) => {
+      const pages = [];
 
       if (!res.data.next_url) {
         return action(this.formatSplitResponses(res.data.resources));
       }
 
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -58,7 +58,7 @@
     "babel-preset-react": "^6.16.0",
     "babel-runtime": "^6.20.0",
     "classnames": "^2.1.5",
-    "cloudgov-style": "^2.5.4",
+    "cloudgov-style": "git://github.com/18F/cg-style.git#ab/activity-log-ux",
```
msecret: This should be changed back before merge.

static_src/util/cf_api.js
```diff
@@ -198,8 +210,15 @@ export default {
     return this.fetchOne(`/spaces/${spaceGuid}/summary`);
   },
 
-  fetchSpaceEvents(spaceGuid) {
-    return this.fetchAllPages(`/spaces/${spaceGuid}/events`, results => results);
+  fetchSpaceEvents(spaceGuid, options) {
```
msecret: If we're only passing `appGuid` here right now, why should it be an `objects` object rather then just an optional `appGuid`?

adborden: Yep, valid point. The `q` parameter takes a handful of [values and operators](https://apidocs.cloudfoundry.org/245/spaces/list_all_events_for_the_space.html) for filtering, so I wanted to start moving our action API closer to something that gives us the flexibility offered by the CF api. Since other APIs also accept a `q` parameter I wanted to establish a convention rather than just sticking arbitrary second, third, etc arguments into these kinds of actions.

I'll have a stronger opinion once we start optimizing other APIs as well, but for now I'll let you make the call.

msecret: I like your thinking

PR #1015, opened 2017-02-15
Automated accessibility testing

msecret:
I did this with Axe-core rather then pa11y, a11y or other tools
because it supported running in selenium. I tried to just make it
part of funtional suite, but couldn't get it working with axe. To
get it working, I used axe builder which sets of the whole env
for the tests.

THis currently just logs the results, I have to write code to
fail a test if there's certain errors and maybe report the errors
in a better format then json.


msecret:
Not doing right now

PR #1019, opened 2017-02-21
Global fetch error pattern

msecret:
Demonstrates the working global fetch error pattern for the app page. This pattern works by having an action catch any fetch errors and then calling an error action for important data that had a fetch error. The error action appends important message information that is global to all fetch errors. An error store keeps track of fetch errors (or other errors eventually). If there are numerous errors, over 5, it will cancel them out and show just one global connection error. A global error container will loop through the errors in the errors store and display a global error for each. This global error is in charge of handling when an error is dismissed by passing an action back to the error store and rendering the actual notifications which show the errors at the top of the page.

A bunch of cf api error tests had to be redone so the expectations come in the `catch` block because now fetch errors return an error.

There is still further customization that could be done:

- The error store should eventually handle different types of errors and should show generic error messages for different types of errors rather then just fetch errors. 
- When data that is required for page to even render (such as app fetch) fails, the notification could be a red "error" type rather then yellow "warning". The point still gets across for now.
- Different actions could be supplied to the global error component so different actions appear in the notification rather then just "refresh". Refresh is fine for now as fetch errors are the only ones using this component.
- The notification or global error component could really use a fade out on destruction. Currently, when a user is closing a notification it can be confusing to see the next one replace it.

Requires https://github.com/18F/cg-style/pull/308


static_src/test/functional/pageobjects/notification.element.js
```diff
@@ -0,0 +1,41 @@
+
+import BaseElement from './base.element';
+
+// https://www.martinfowler.com/bliki/PageObject.html
+//
+// Represents a OrgQuicklookElement for making assertions against. This makes it
```
adborden: Here, too, we'd like to remove the reference to OrgQuicklookElement :)

static_src/components/notification.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+const STATUSES = Object.assign({}, entityHealth, keymirror({
+  info: null
+}));
+
+
+const propTypes = {
+  message: React.PropTypes.string,
+  status: React.PropTypes.oneOf(Object.keys(STATUSES)),
+  actions: React.PropTypes.arrayOf(React.PropTypes.object),
```
msecret: I'm not sure I agree with that as to create each action, it needs an object with a `text` and `clickHandler` fields.

static_src/stores/app_store.js
```diff
@@ -22,7 +22,7 @@ export class AppStore extends BaseStore {
   }
 
   get loading() {
-    return this._fetchAll || this._fetchApp || this._fetchAppStats;
+    return this._fetchAll || this._fetchApp;
```
msecret: I think I took this out to test something

static_src/stores/error_store.js
```diff
@@ -0,0 +1,62 @@
...
+        this.push(err);
+        this.checkForMaxFetchErrors();
+        break;
+      }
+
+      case errorActionTypes.DISMISS: {
+        const errIdx = this.getAll().findIndex((err) => err === action.err);
+        if (errIdx) {
+          // TODO little unsafe to access data here?
+          this._data = this._data.delete(errIdx);
```
msecret: There is one. The problem is everything uses guids, which errors don't have.

static_src/components/notification.jsx
adborden: Let me clarify, if you made it type `node`, you could do this

```javascript
const actions = (
  <span>
    <Action clickHandler={ handleRefresh }>Refresh</Action>
    <Action clickHandler={ handleCancel }>Cancel</Action>
  </span>
);
<Notification actions={ actions } />
```

Arrays might also count as `node`, so you wouldn't even need the `span` in that case. I'm not sure.

static_src/stores/error_store.js
adborden: Oh, I see. How about you assign a guid to error? The guid could be any string. You could keep a global id in ErrorStore and increment it on error creation. Then all the BaseStore stuff just works.

PR #1022, opened 2017-02-23
This restores parallel tests with webdriverio

adborden:
This allows the test server to be executed with command line arguments. The
arguments will be spawned as a child process. This allows us to wrap commands
with a test server. The server is started and connection information is
appended to the environment for the child process. Once the child process exits,
the testing server will be closed.

I call this a "sidecar" pattern based on Netflix's Prana[1] except that the
services provided by the sidecar is really only making the test server
available.

[1] http://techblog.netflix.com/2014/11/prana-sidecar-for-your-netflix-paas.html


adborden:
Heads up, this is based on [ms-global_error](https://github.com/18F/cg-dashboard/pull/1019).

static_src/test/server/server.js
```diff
@@ -1,6 +1,62 @@
 
+// This testing server acts as a "sidecar" for a main process given as command
+// line arguments, or simply starts the testing server.
+//
+// Sidecar pattern comes from Netflix's Prana[1] but instead of providing
+// platform services, we just provide access to the testing server.
+//
+// TODO the testing server does not require any cleanup nothing is done to stop
```
msecret: This seems more like an observation, was confused about what actually has to be done?

adborden: Most servers have some kind of [stop/close](https://hapijs.com/api#server-events) event so that you can clean up any resources like database connections. In this case, we would call `server.stop()` to trigger the shutdown process. Since nothing is listening to the stop event anyway, it's not a big deal.

I'll update the TODO to be more specific: It should listen to SIGTERM/SIGINT events and call `server.stop()`.

adborden:
@msecret I ended up just tackling the TODO

PR #1027, opened 2017-03-01
FormStore

adborden:
This introduces a `FormStore` to manage forms. Each form must be assigned a unique guid. Components using form can hook into the `onSubmit` handler of the form, or they can subscribe to `FormStore` changes to react to individual form field changes.

Initially we discussed putting the validators in the form model itself, but this ended up being difficult, because it resulted in cascading changes (value in fieldA changes -> validator of formB changes -> error of formB changes). More work would be needed to do this. The advantage is that the actions could handle the success/error actions more transparently, simplifying `FormElement`. The downside is that it's a bit awkward for the component to juggle FormStore changes and then update the validators through a series of form change events (potentially one for each form field).

### DIFF
--- a/static_src/components/create_service_instance.jsx
+++ b/static_src/components/create_service_instance.jsx
@@ -9,6 +9,7 @@ import ReactDOM from 'react-dom';
 import Box from './box.jsx';
 import Action from './action.jsx';
 import { Form, FormText, FormSelect, FormElement, FormError } from './form';
+import FormStore from '../stores/form_store';
 import Loading from './loading.jsx';
 import OrgStore from '../stores/org_store.js';
 import SpaceStore from '../stores/space_store.js';
@@ -36,9 +37,11 @@ export default class CreateServiceInstance extends React.Component {
       createError: ServiceInstanceStore.createError
     };
 
+    // Create the form in the store
+    FormStore.create('create-service-form');
+
     this.validateString = validateString().bind(this);
     this._onChange = this._onChange.bind(this);
-    this._onValidateForm = this._onValidateForm.bind(this);
     this._onValidForm = this._onValidForm.bind(this);
     this._onCancelForm = this._onCancelForm.bind(this);
     this.styler = createStyler(style);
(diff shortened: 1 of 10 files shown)


static_src/components/create_service_instance.jsx
```diff
@@ -113,12 +114,11 @@ export default class CreateServiceInstance extends React.Component {
     return (
       <div className = { this.styler('actions-large') }>
         { createError }
-        <Form action="/service_instances"
+        <Form
+          guid="create-service-form"
```
msecret: Should this be a const var, because it's used multiple places?

static_src/components/form/form_element.jsx
```diff
@@ -11,77 +12,62 @@ export default class FormElement extends React.Component {
...
   }
 
   onChange(e) {
-    this.setState({ value: e.target.value }, () => {
-      this.validate();
+    const value = e.target.value;
+    const err = this.props.validator(value, this.props.name);
+    this.setState({ err, value }, () => {
+      formActions.changeField(this.props.formGuid, this.props.name, value)
+        .then(() => {
```
msecret: I was confused to see this in the component. What was your thinking it putting it here rather then error actions?

msecret: Ah because the `err` is here. So this code will set the current state of the form, so now it renders with the new value and an error. Then it will change the field so the form model in the store is updated. Then it will set an error or success of the field through the form actions?

static_src/components/usage_and_limits.jsx
```diff
@@ -217,9 +204,21 @@ export default class UsageAndLimits extends React.Component {
     );
   }
 
-  _onSubmit(e) {
-    e.preventDefault();
-    appActions.updateApp(this.props.app.guid, this.state.partialApp);
+  _onSubmit(errors, form) {
+    if (errors.length) {
+      console.warn({ errors }); // eslint-disable-line no-console
```
msecret: Is there a reason we want the errors in the console here? Are we not doing anything with the errors in this case?

static_src/components/form/form_element.jsx
adborden: Right, exactly.

static_src/components/usage_and_limits.jsx
adborden: No reason, _onSubmit taking an error is left over from the old `onValid` handler. By default, on submit, the `Form` component will display a message if there are errors. This behavior already existed and I didn't want to change it, although I think it needs some rethinking. Here's what it looks like. I'll remove the `console.warn` for now.

peek 2017-03-01 16-55

PR #1029, opened 2017-03-03
Global error enhancements

msecret:
- Modified the message so it optionally show a cloud.gov status
link if one exists, so user can check if cloud.gov is down.
- Changed the messaging a bit based on usability testing.
- Remove the loading indicator in case of errored app stats
fetch request.


static_src/actions/error_actions.js
```diff
@@ -38,10 +38,10 @@ export default {
   },
 
   importantDataFetchError(err, entityMessage) {
-    let msg = 'Page failed to load, please try again';
+    let msg = 'Connection issue, please try again';
```
adborden: This is a little terse, maybe something more like:
> There was an issue connecting to dashboard.cloud.gov, please try again later.

static_src/components/global_error.jsx
```diff
@@ -35,10 +36,21 @@ export default class GlobalError extends React.Component {
 
   render() {
     const err = this.props.err;
+    const link = (config.docs.status) && (
+      <span> check <a target="_blank" href={ config.docs.status }>
+        cloud.gov's status</a> or
```
adborden: We want to avoid hard-coding "cloud.gov" here, maybe:
> ... check [the platform's status] or ...

msecret: oh yea forgot, good catch.

static_src/actions/error_actions.js
msecret: I like that, but from the user testing, it seems like the specific messaging was useful, so I want to  keep the "app usage data might not be available" in there.

PR #1030, opened 2017-03-08
Release 1.16

msecret:
wait nvm, this needs global errors.


adborden:
you could keep the PR open and then just make sure the change gets onto the right branch ;)

PR #1031, opened 2017-03-08
Add more import fetch error handling

msecret:
To more actions

### DIFF
--- a/static_src/actions/org_actions.js
+++ b/static_src/actions/org_actions.js
@@ -6,6 +6,7 @@
 
 import AppDispatcher from '../dispatcher.js';
 import cfApi from '../util/cf_api.js';
+import errorActions from './error_actions.js';
 import { orgActionTypes } from '../constants';
 import spaceActions from './space_actions';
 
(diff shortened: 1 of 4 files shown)


static_src/actions/org_actions.js
```diff
@@ -26,7 +27,14 @@ const orgActions = {
...
     });
 
-    return cfApi.fetchOrg(orgGuid).then(orgActions.receivedOrg);
+    return cfApi.fetchOrg(orgGuid)
+      .then(orgActions.receivedOrg)
+      .catch((err) =>
+        errorActions.importantDataFetchError(
+          err,
+          'organization data may be incomplete'
+        )
```
adborden: Just a minor thing, but it's worth adding a helper to errorActions like, `defaultFetchErrorHandler` or something that returns a function that calls the action. Then you could have:
```javascript
    return cfApi.fetchOrg(orgGuid)
      .then(orgActions.receivedOrg)
      .catch(errorActions.defaultFetchErrorHandler('organization data may be incomplete'));
```

msecret: So what would the benefit be of that? It looks like that doesn't require the `err` being passed?

adborden:
Quoting Marco Segreto (2017-03-08 17:41:28)
> msecret commented on this pull request.
> 
> 
> 
> > @@ -26,7 +27,14 @@ const orgActions = {
>        orgGuid
>      });
>  
> -    return cfApi.fetchOrg(orgGuid).then(orgActions.receivedOrg);
> +    return cfApi.fetchOrg(orgGuid)
> +      .then(orgActions.receivedOrg)
> +      .catch((err) =>
> +        errorActions.importantDataFetchError(
> +          err,
> +          'organization data may be incomplete'
> +        )
> 
> So what would the benefit be of that? It looks like that doesn't require the `err` being passed?

Yeah, that's it, just a little syntactic sugar.

--
Aaron D Borden
Human and Hacker

PR #1036, opened 2017-03-14
Complete Local Dev Environment

jseppi:
# Complete Local Dev Environment

## Value

In order to:

1. Get developers up quickly
1. Minimize variations between machines
1. Minimize the number of tools to install.
1. Prevent developers from having to do all their experiments against production data.

It would be helpful to have a self contained environment where developers can do all their development locally on.

## Solution

For points 1-3, use a dockerized environment that contains tools (mailcatcher, redis) and the dashboard.
For point 4, use PCFDev which deploys a local cloud foundry instance on their computer.

## Feature
Added instructions for how to setup your local dev environment with PCFDev + Docker (w/ docker-compose).
Ability and documentation to do visual debugging via VNC.
Ability and documentation to do npm linking between this repo and cg-style.
Documentation to do one-off commands with tools that don't have to be downloaded locally.

## Cleanup
Moved old setup instructions from main README to a new place.

## Misc. code changes:
Go:
- Disable SSL check given CF_LOCAL env var. Necessary for pcfdev endpoints.
- Safeguard: Double check that if you aren't using CF_LOCAL, you have to SECURE_COOKIES (helps ensure we don't do anything unsafe in production)
- Added tests
- The Go tests already spin up Docker containers. But now that it's trying to run the tests inside docker (i.e. Docker inside Docker), we have to create a network `servicesTestNetwork` so that the tests can reach the containers. We added a helper function named `connectToNetwork`. Then make the mail and redis tests use it.
- Reorder reading Redis credentials. Read from REDIS_URI before reading from redis service instance. (makes it a lot easier to test)

#### Shortened Startup

The app starts up, we login to the PCFDev UAA server and it redirects us back to the dashboard.

jul-07-2017 15-27-28

#### Visual Debugging of Tests Running
We run the functional tests and Chrome starts up and runs the tests.

jul-07-2017 15-51-10

This was a bonus feature but deemed necessary so developers can see what's going on in the container in case of a failure.

## Future Plans

Previously, we had integration tests that ran against a live CF deployment. A developer could run them locally. The problem was when we had multiple developers on the team, the data could get messed up if two developers were running the tests against the same CF deployment at the same time (there was only the one prod E/W CF deployment back then). (Also, this was before the dashboard was redesigned). 
Now, since we have a local CF deployment (via PCFDev), we can bring them back and a developer can run the integration tests against their local CF deployment and we can have CI run integration against staging CF and production CF deployments and have no race condition problems.


msecret:
Thanks so much @jseppi . I'll hold off to write comments, but please let me know if you want feedback on anything specific or have any questions.

jseppi:
I'm definitely fine with running-feedback/commentary, so go ahead and comment anywhere that strikes your fancy :)

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM node:6
```
msecret: At some point, it'd be good to get the exact node version that's in `.nvmrc`

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: '2'
+services:
+  app:
+    build: .
+    volumes:
+    - .:/usr/src/app
+    - node-modules:/usr/src/app/node_modules/
```
msecret: I'm guess this makes sure node_modules is cached, or loaded from existing system? That sounds cool.

jseppi: yeah, it stores `node_modules` in a docker volume

Dockerfile
jseppi: Do you mean to grab it dynamically, or to just specify what is currently in `.nvmrc` (ie, `FROM node:6.7`)?

I don't believe there is a way to grab it dynamically in the Dockerfile itself, though we could potentially make a container that has `nvm` and have it use the repo's `.nvmrc` to turn on whichever node version is specified.

adborden:
Thanks! We have plans to make the dashboard a general Cloud Foundry dashboard, so other CF instances can use it too with their own brand on it. If/when we go down that route, I suspect that cloud.gov would create a new repo containing manifests and specific configuration settings with the general cf-dashboard as a git submodule or maybe a golang package or even an npm module (we haven't worked out those details yet).

Anyway, in that case, do you think the docker setup would be part of the open source repo or would it be part of the cloud.gov repo?

dockerizing-notes.md
```diff
@@ -0,0 +1,15 @@
+# Dockerizing Notes
+
+## Running
+
+1. `docker-compose build`
+1. `docker-compose run app npm install --no-shrinkwrap`
+  because the shrinkwrap has darwin-specific packages listed
```
adborden: Hmm, this is interesting, I thought we had removed those. I'm able to `npm install` on ubuntu. CircleCI also doesn't have any trouble with it. Wonder if npm3 takes care of it?

Dockerfile
toolness: another option, which we do in calc, is to leave it hard coded but add a unit test that verifies the Dockerfile is in sync with .nvmrc.

msecret: I think either of those options is fine.

jseppi:
@adborden:

> Anyway, in that case, do you think the docker setup would be part of the open source repo or would it be part of the cloud.gov repo?

Hrm...I suppose that's up to you all, probably based on how well it works and how you want the maintenance of it to be handled :)

jseppi:
@msecret: I'm not quite sure what the requirements are for the `npm link`'ing of cg-styles is. Could you clarify that process/need? If a call is easier than writing it up we can do that instead.

msecret:
@jseppi here are the instructions for npm linking https://github.com/18F/cg-style/blob/master/CONTRIBUTING.md#setup

jseppi:
Excellent - thanks!

adborden:
Quoting James Seppi (2017-03-16 09:58:48)
> Hrm...I suppose that's up to you all, probably based on how well it works and how you want the maintenance of it to be handled :)

Yeah, I'm leaning toward not including it in the common repo but
curious if other folks have experience with it.

--
Aaron D Borden
Human and Hacker

afeld:
Why not both? 😉

msecret:
@jseppi @toolness I haven't had a chance to try this, but will hopefully get to it soon.

jcscottiii:
rebased and added pcfdev and golang

jcscottiii:
heavily inspired by https://github.com/18F/omb-eregs

rememberlenny:
A.M.A.Z.I.N.G!

jcscottiii:
You can try it out via running https://github.com/18F/cg-dashboard/blob/dockerize/devtools/docker-setup.md

devtools/docker-setup.md
```diff
@@ -0,0 +1,231 @@
...
+```sh
+docker-compose stop app
+docker-compose up -d app
+```
+
+It will replace the existing app container with a new one, thus recompiling the backend.
+
+### Full Setup: Running One-Offs
+
+Running one-offs is especially important if you don't have the proper tools
```
el-mapache: I may have missed something earlier that explain this, but what are the proper tools being referenced here?

.cfignore
```diff
@@ -9,3 +9,8 @@ cg-deck
 cg-dashboard
 /screenshots/
 static/tests/
+.org.chromium*
+# only exclude these index.html files from the VNC service.
+# they only appear in the root folder.
+/index.html*
```
el-mapache: I see that we are ignoring these index.html files, but running the commands specified in the docker README for the first time generated something like 250 index.html files. Is there a way we can reduce that number?

Or are those all safe to delete? If so, maybe we could add a script to the package.json to manage that?

devtools/docker-setup.md
jcscottiii: oh I meant Go and node. I'll clarify that.

.cfignore
jcscottiii: all of them are safe to delete! :) do we want to add a script to package.json? or just tell people to run `rm index.html*`

also, do you know of any tools that work like `watch` but instead of compiling we could delete them? then the user wouldn't have to worry about it at all.

el-mapache: I think a watch util would probably be best, then we don't have to run any commands. I found this [script](https://gist.github.com/senko/1154509), but it requires `inotify`, which I don't believe is available on a mac. I think the analog for OSX is `fsevents`. I also found this library on npm: https://www.npmjs.com/package/watchman. There is also `chokidar` (https://github.com/paulmillr/chokidar), which would probably do what we want. Its used by most of the build libraries (gulp, webpack, etc). It should be pretty easy to add a hook to listen for `index.html` file creation and then run something to delete them.

jcscottiii: that's awesome! the great part is that i can put it in the container (which is linux) and it will clean up and then it will automatically reflect on the mac's disk

jcscottiii: i'll incorporate that

devtools/docker-setup.md
```diff
@@ -0,0 +1,231 @@
...
+- [docker-compose](https://docs.docker.com/compose/install/)
+
+### Quick-Start: Running
+
+```sh
+docker-compose up frontend_testing_server
+```
+
+It will say the server is up when everything is done. At the point, you can
+navigate to `http://localhost:8001` and see the testing server.
```
el-mapache: This is probably a good place to add a note about ignoring the security warning and just clicking through the `proceed to...` link? Maybe bold it or add small header to call it out.

devtools/docker-setup.md
jcscottiii: done

devtools/docker-setup.md
jcscottiii: i added the note here: https://github.com/18F/cg-dashboard/pull/1036/commits/1abf66766cde43cef5e684687bd3ee24fec7a749
 
in the quick-start, you don't have to login (since it doesn't use the real CF API).

Let me know your thoughts

.cfignore
jcscottiii: done!

el-mapache:
This is great! Thanks for making this happen!

rememberlenny:
Should we merge now?

el-mapache:
Yes, merging!

PR #1037, opened 2017-03-16
Fixes infinite loading bug on services

msecret:
The reason the bug was happening:
The service instance store had a `loading` state set to true even after
it was done loading. I couldn't reproduce, but this may have been the
case with the plan store too.

The stores were not manually emitting
a change on the received events. They assume the `merge` methods
would emit a change. But the `merge` functions only emit changes if
something actually changed. So in stores like the instance store, where
separate fetches are happening, I think the instances were already,
meaning nothing had to be merged, meaning emit change didn't get called.
this meant that the loading state was still true on the store.

### DIFF
--- a/static_src/stores/service_instance_store.js
+++ b/static_src/stores/service_instance_store.js
@@ -141,16 +141,20 @@ export class ServiceInstanceStore extends BaseStore {
       case serviceActionTypes.SERVICE_INSTANCE_RECEIVED: {
         this._fetching = false;
         const instance = action.serviceInstance;
-        this.merge('guid', instance, () => { });
-        this.emitChange();
+        this.merge('guid', instance, () => {
+          // Always emitchange as fetch state was changed.
+          this.emitChange();
+        });
         break;
       }
 
       case serviceActionTypes.SERVICE_INSTANCES_RECEIVED: {
         this._fetchAll = false;
         const services = action.serviceInstances;
-        this.mergeMany('guid', services, () => { });
-        this.emitChange();
+        this.mergeMany('guid', services, () => {
+          // Always emitchange as fetch state was changed.
+          this.emitChange();
+        });
         break;
       }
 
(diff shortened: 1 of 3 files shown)


adborden:
This doesn't look right, both `merge` and `mergeMany` already call `emitChange`. Can you confirm why BaseStore wasn't triggering a change when it should? https://github.com/18F/cg-dashboard/blob/master/static_src/stores/base_store.js#L16

adborden:
Ah, I see, some of those Action handlers change the fetch state outside of the models. If the model didn't change, the `emitChange` isn't fired. In that case we should add a comment for those cases that we're doing an explicit `emitChange` because the fetch state was changed.

static_src/stores/service_instance_store.js
```diff
@@ -141,14 +141,16 @@ export class ServiceInstanceStore extends BaseStore {
       case serviceActionTypes.SERVICE_INSTANCE_RECEIVED: {
         this._fetching = false;
         const instance = action.serviceInstance;
-        this.merge('guid', instance);
+        this.merge('guid', instance, () => { });
+        this.emitChange();
```
adborden: Let's move this into the callback of merge with a comment.

static_src/stores/service_plan_store.js
```diff
@@ -54,7 +54,8 @@ export class ServicePlanStore extends BaseStore {
       case serviceActionTypes.SERVICE_PLAN_RECEIVED: {
         const servicePlan = action.servicePlan;
         const servicePlanReceived = Object.assign({}, servicePlan, { fetching: false });
-        this.merge('guid', servicePlanReceived);
+        this.merge('guid', servicePlanReceived, () => { });
+        this.emitChange();
```
adborden: This emit change shouldn't be necessary, if the `servicePlanReceived` was changed then `merge` will emit for us. We're not modifying any other state here outside of the model.

msecret: You're correct. It should only be necessary when some other state is fetched outside of the merge.

PR #1038, opened 2017-03-17
Simplify deploy script; disable e/w updates

adborden:
- Remove deprecated cg-deck deploy
- Allow cf environment to be specified as parameter
- Remove unused prepare_deploy.sh script
- Remove unnecessary `cd`'s in circleci config


msecret:
We should probably deploy one last time to get rid of those two bugs? Maybe we'll merge this after that last deploy?

adborden:
Sure, we can also grab those branches with only the fixes and merge them into `release-1.16` and tag/deploy that. That way we don't have to worry about what is on master.

PR #1039, opened 2017-03-17
Expressive empty state for creating apps

adborden:
When you've got an empty space (no apps) we provide some hints as to how the user would create a new app.

This depends on:
- [ ] https://github.com/18F/cg-dashboard/pull/1038
- [ ] https://github.com/18F/cg-style/pull/312

### DIFF
--- a/static_src/components/info_app_create.jsx
+++ b/static_src/components/info_app_create.jsx
@@ -15,7 +15,7 @@ const defaultProps = {
 };
 
 
-export default class SpaceContainer extends React.Component {
+export default class InfoAppCreate extends React.Component {
   constructor(props) {
     super(props);
     this.styler = createStyler(style);
@@ -61,5 +61,5 @@ export default class SpaceContainer extends React.Component {
   }
 }
 
-SpaceContainer.propTypes = propTypes;
-SpaceContainer.defaultProps = defaultProps;
+InfoAppCreate.propTypes = propTypes;
+InfoAppCreate.defaultProps = defaultProps;
--- a/static_src/components/info_app_create.jsx
+++ b/static_src/components/info_app_create.jsx
@@ -1,0 +1,65 @@
+
+import React from 'react';
+
+import style from 'cloudgov-style/css/cloudgov-style.css';
+import { config } from 'skin';
+
+import createStyler from '../util/create_styler';
+
+const propTypes = {
+  org: React.PropTypes.object,
+  space: React.PropTypes.object
+};
+
+const defaultProps = {
+};
+
+
+export default class SpaceContainer extends React.Component {
+  constructor(props) {
+    super(props);
+    this.styler = createStyler(style);
+  }
+
+  render() {
+    const org = this.props.org || {};
+    const space = this.props.space || {};
+
+    return (
+      <div className={ this.styler('info', 'info-app_create') }>
+        <p>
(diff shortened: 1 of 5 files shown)


static_src/components/info_app_create.jsx
```diff
@@ -0,0 +1,65 @@
...
+const propTypes = {
+  org: React.PropTypes.object,
+  space: React.PropTypes.object
+};
+
+const defaultProps = {
+};
+
+
+export default class SpaceContainer extends React.Component {
```
msecret: This should be called `InfoAppCreate` not `SpaceContainer`

static_src/components/info_app_create.jsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  render() {
+    const org = this.props.org || {};
+    const space = this.props.space || {};
+
+    return (
+      <div className={ this.styler('info', 'info-app_create') }>
+        <p>
+          Learn how to <a href={ config.docs.deploying_apps }>deploy a new app</a>.
```
msecret: What happens when the `docs.deploying_apps` is not in the config?

msecret: Same for other times it's being used here.

static_src/components/info_logs.jsx
```diff
@@ -15,7 +16,7 @@ export default class InfoLogs extends React.Component {
     return (
       <PanelActions>
         <p>
-          View more logs at <a href="https://logs.cloud.gov">logs.cloud.gov</a> (for East/West environment) or <a href="https://logs.fr.cloud.gov">logs.fr.cloud.gov</a> (for GovCloud environment).
+          View more logs at <a href={ config.platform.logs.url }>{ config.platform.logs.name }</a>.
```
msecret: Should we only show this link when both the name and url exist?

static_src/components/space_container.jsx
```diff
@@ -101,8 +101,7 @@ export default class SpaceContainer extends React.Component {
 
           <AppList />
           <PanelActions>
-            <span>Learn how to <a href={ config.docs.deploying_apps }>deploy a
-            new app</a>.</span>
+            <InfoAppCreate space={ space } org={ this.state.currentOrg } />
```
msecret: This technically shouldn't be in panel actions. I think I changed it on my branch.

static_src/components/info_app_create.jsx
adborden: derp.

static_src/components/info_app_create.jsx
adborden: The app would crash, or the link would just be null. I think this is really a problem with the current skin config design, so I'd rather not fix it here. The skin config should make it so it's okay to have missing keys and get sane defaults or at least not crash.

If you wanted to fix it here, (and everywhere the config is used) it looks like this:
```javascript
import { config } from 'skin';

// For each key you're using:
const deployingApps = config.docs && config.docs.deploying_apps || "https://cloudfoundry.org";
```

static_src/components/info_logs.jsx
adborden: I think this would also be addressed by a new skin config design, so I wouldn't touch it. I would like to write something like this:
```javascript
if (config.platform.logs) {
  return <a ... />;
}
// else nothing
```

PR #1040, opened 2017-03-20
Release 1.16

adborden:
Bugfixes

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "cloudgov-dashboard",
-  "version": "1.14.0",
+  "version": "1.16.1",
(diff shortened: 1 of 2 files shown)


npm-shrinkwrap.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cloudgov-dashboard",
-  "version": "1.14.0",
```
fureigh: @adborden Should I be concerned that this was changing from `1.14.0` and not from `1.16.0`? If not, LGTM.

adborden: Good catch, it's supposed to match package.json, so it was an oversight that it wasn't updated to 1.16.0 previously.

Issue #965, opened 2017-01-19
Expressive empty state for no applications


adborden:
As a new user, you will likely see the dashboard with no applications. An empty expressive state could prompt you on how to push your first application. Because it's currently not possible to push an application from the dashboard, having some help text, a link to the [Your first deploy][first-deploy] and a sample `cf push` command could go a long way.

As an example, here's how Heroku describes deploying via command line. _Note: Heroku CLI deploy uses git, which is different than CF so the steps and commands will be quite different._

![screenshot from 2017-01-19 09-34-03](https://cloud.githubusercontent.com/assets/509703/22117950/baa476f2-de2a-11e6-942e-2cf70e2d1052.png)

[first-deploy]: https://cloud.gov/docs/getting-started/your-first-deploy/

adborden:
Here's a draft at the cli steps. API, org, space should all be dynamic since the user will be looking at a particular org/space/cloud foundry instance in the dashboard.

From your application's directory:

    $ cf login -a api.fr.cloud.gov --sso
    $ cf target -o org -s space
    $ cf push my-app

adborden:
@msecret I think this is worth taking a look at for one of our upcoming sprints.

adborden:
An example empty state design https://gsa.invisionapp.com/d/main#/console/8652272/208089392/preview

adborden:
Some sample copy:

> This space has no apps.
> Learn how to [deploy an new app](https://cloud.gov/docs/apps/deployment/).
> If you already have your application code setup, you can run these commands from your application directory:
```
$ cf login -a api.fr.cloud.gov --sso
$ cf target -o org -s space
$ cf push my-app
```
> If this is [your first deploy](https://cloud.gov/docs/getting-started/your-first-deploy/), you can checkout these tips to get started.

adborden:
Gonna call this done.

PR #1041, opened 2017-03-21
Expressive empty states

msecret:
Adds empty states for:
- no orgs
- no users
- just one user
- no spaces

Modifies the existing no apps state for pages like overview where there are many side by side.

Still needs services, which is a larger one and not required for usability testing, so don't need for this iteration.

Merged in #965 
Requires https://github.com/18F/cg-style/pull/313
Ref https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-556

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "test-unit": "NODE_ENV=test karma start --single-run",
     "test-selenium-install": "selenium-standalone install",
     "testing-server": "PORT=8001 npm run testing-server-run",
-    "testing-server-run": "nodemon --watch ./static_src/test/server ./static_src/test/server/server.js",
+    "testing-server-run": "node ./static_src/test/server/server.js",
     "watch-server": "npm run testing-server & npm run watch",
     "watch-test": "NODE_ENV=test karma start --browsers Chrome",
     "watch": "webpack --watch"
--- a/package.json
+++ b/package.json
@@ -90,7 +90,6 @@
     "moment": "^2.17.1",
     "moment-timezone": "^0.5.10",
     "node-libs-browser": "^0.5.2",
-    "nodemon": "^1.11.0",
     "normalize.css": "^3.0.3",
     "react": "^15.4.2",
     "react-dom": "^15.4.1",
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "test-unit": "NODE_ENV=test karma start --single-run",
     "test-selenium-install": "selenium-standalone install",
     "testing-server": "PORT=8001 npm run testing-server-run",
-    "testing-server-run": "node ./static_src/test/server/server.js",
+    "testing-server-run": "nodemon --watch ./static_src/test/server ./static_src/test/server/server.js",
     "watch-server": "npm run testing-server & npm run watch",
     "watch-test": "NODE_ENV=test karma start --browsers Chrome",
     "watch": "webpack --watch"
@@ -90,6 +90,7 @@
     "moment": "^2.17.1",
     "moment-timezone": "^0.5.10",
     "node-libs-browser": "^0.5.2",
+    "nodemon": "^1.11.0",
     "normalize.css": "^3.0.3",
     "react": "^15.4.2",
     "react-dom": "^15.4.1",
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "test-unit": "NODE_ENV=test karma start --single-run",
     "test-selenium-install": "selenium-standalone install",
     "testing-server": "PORT=8001 npm run testing-server-run",
-    "testing-server-run": "node ./static_src/test/server/server.js",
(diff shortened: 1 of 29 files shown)


static_src/components/app_list.jsx
```diff
@@ -65,7 +69,11 @@ export default class AppList extends React.Component {
     );
 
     if (this.state.empty) {
-      content = <h4 className="test-none_message">No apps</h4>;
+      content = (
+        <EntityEmpty callout="You have no apps in this space">
+          <InfoAppCreate space={ this.state.currentSpace } org={ this.state.currentOrg } />
```
adborden: Do we want to check the user roles here?

static_src/css/main.css
```diff
@@ -36,3 +36,4 @@ input[type="checkbox"] + label::before,
 input[type="radio"] + label::before {
   display: none;
 }
+
```
adborden: Hmm, I didn't even know we had CSS in here...

static_src/stores/user_store.js
```diff
@@ -249,19 +310,20 @@ class UserStore extends BaseStore {
...
+  get isLoadingCurrentUser() {
+    return this._loading.currentUser === true;
   }
 
-  currentUserHasSpaceRole(role) {
-    return this._hasRole(role, 'space_roles');
-  }
+  hasRole(userGuid, entityGuid, roleToCheck) {
+    let wrappedRoles = roleToCheck;
+    if (!roleToCheck.length) {
```
adborden: We should check the type here, both strings and arrays have length.

static_src/stores/user_store.js
```diff
@@ -249,19 +310,20 @@ class UserStore extends BaseStore {
...
+      wrappedRoles = [roleToCheck];
+    }
 
-  currentUserHasOrgRole(role) {
-    return this._hasRole(role, 'organization_roles');
+    const key = entityGuid;
+    const user = this.get(userGuid);
+    const roles = user && user.roles && user.roles[key] || [];
+    return !!roles.find((role) => wrappedRoles.includes(role));
```
adborden: We should add documentation for this method, I was thinking if you passed an array of roles, the check would only succeed if the user has ALL roles, but this is checking for ANY role which is a big difference.

static_src/test/server/api.js
```diff
@@ -19,9 +19,17 @@ var spaceRoutes = require('./fixtures/space_routes');
...
+var userOrganizations = require('./fixtures/user_organizations.js');
+var userSpaces = require('./fixtures/user_spaces.js');
 
 var BASE_URL = '/v2';
 
+const ENV_NO_ORGS = process.env.NO_ORGS || false;
+const ENV_NO_SPACES = process.env.NO_SPACES || false;
+const ENV_NO_APPS = process.env.NO_APPS || false;
+const ENV_NO_ORG_USERS = process.env.NO_ORG_USERS || false;
+const ENV_NO_SPACE_USERS = process.env.NO_SPACE_USERS || false;
```
adborden: Hmm, I wonder if we can use the smocks admin to pre-program this. Not sure how smocks persists its data...

adborden:
We should be writing more component tests too, I'll add one for `InfoAppCreate`

package.json
```diff
@@ -18,7 +18,7 @@
     "test-unit": "NODE_ENV=test karma start --single-run",
     "test-selenium-install": "selenium-standalone install",
     "testing-server": "PORT=8001 npm run testing-server-run",
-    "testing-server-run": "nodemon --watch ./static_src/test/server ./static_src/test/server/server.js",
+    "testing-server-run": "node ./static_src/test/server/server.js",
```
adborden: Don't forget to remove it from `package.json`

static_src/actions/user_actions.js
```diff
@@ -122,11 +125,172 @@ export default {
...
+  receivedUser(user) {
+    AppDispatcher.handleServerAction({
+      type: userActionTypes.USER_RECEIVED,
+      user
+    });
+
+    return Promise.resolve(user);
+  },
+
+  fetchUserSummary(userGuid) {
```
msecret: I may have left it in.

static_src/components/app_list.jsx
msecret: I think I moved the roles to `InfoAppCreate`.

static_src/css/main.css
msecret: Yea we were using it for quick styles at first, but found that putting it in cg-style was easier.

static_src/stores/user_store.js
msecret: Keep forgetting that.

static_src/test/server/api.js
msecret: eya I keep forgetting about that admin

static_src/components/app_list.jsx
adborden: Yeah, I think you still need to pass in a user to `props` for it to do the check.

PR #1042, opened 2017-03-21
User testing scenarios

msecret:
WIP,

Sets up correct testing scenarios for things like expressive empty states.


msecret:
I experimented more with the smocks API, and found that to be a better use case, so will use that instead of this.

PR #1043, opened 2017-03-21
Dockerize w/ npm link to cg-style

toolness:
**Note: this is a PR against #1036, not the main branch.**

This adds support for `npm link`ing cg-style with cg-dashboard.

It assumes, though, that you have cg-style checked out relative to the dashboard directory at `../cg-style`.

It runs OK on docker OS X native, but might not run on `dinghy` due to the differences between the way that the two do file permissions and stuff.

I also reverted the `nvm` stuff because it made the docker setup easier to understand (oy sorry @jseppi)... I think a better solution, unless nvm is *really* needed in the container, may be to simply add a static test to ensure that `.nvmrc` and `Dockerfile` specify the same node version (we do this in CALC all over the place).

### DIFF
--- a/dockerizing-notes.md
+++ b/dockerizing-notes.md
@@ -1,16 +1,17 @@
 # Dockerizing Notes
 
-- `nvm` is run when the container starts (via `docker-entrypoint.sh`) and will
-  use the project's `.nvmrc` file to determine which node version to install
-  and use.
 - Fixed the issue with `npm install` that respects `npm-shrinkwrap.json`. the
   problem was not being able to build the same fsevents due to not having
   python in the container.
 
 ## Running
 
+You will first need to ensure that `../cg-style` exists relative
+to your cg-dashboard checkout and contains a checkout of
+[`cg-style`](https://github.com/18F/cg-style).
+
 1. `docker-compose build`
-1. `docker-compose run app npm install`
+1. `docker-compose run app ./setup-npm-link.sh`
 1. Visit `localhost:8001` (or `app.cgdashboard.docker` if using `dinghy`)
 
 ## Other stuff:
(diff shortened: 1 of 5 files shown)


dockerizing-notes.md
```diff
@@ -19,6 +20,7 @@
 
 ## TODO:
 
-* figure out what is needed for also developing `cg-style` (team usually uses `npm link`)
 * figure out why the shrinkwrap doesn't work -- looks like it has to do
```
jseppi: whoops, I had gotten this working over in #1036, but forgot to remove from these notes. Is it not working as part of `setup-npm-link.sh`?

msecret:
I think making the cg-style path modifiable/an option would help as many of us have this project installed in a strange location due to Go's requirements on where you put it's packages. So most people won't have cg style at `../cg-style`.

toolness:
Ah yeah, sorry, I meant to mention that--I believe we should be able to make that configurable through the use of environment variables.

dockerizing-notes.md
toolness: Oh, um, actually I am not sure. I thought I saw the error on my system, where `npm install` failed because of that one `fsevents` darwin dependency, and then I re-ran the thing with `--no-shrinkwrap` or whatever and it worked. But maybe I had accidentally removed some stuff that I thought was `nvm` related but was actually related to your shrinkwrap fix?

One frustrating thing right now is that it's cumbersome to test this because `npm install` takes so long, especially when it needs to be run *twice* (once in each repo). 😞

jseppi: Oh yeah, I think you need to have this stuff in there:

```dockerfile
RUN apt-get update && apt-get install -y -q --no-install-recommends \		
        build-essential \			
        python \		
        python-dev \		
    && rm -rf /var/lib/apt/lists/*
```

Haven't tested -- you might be able to get rid of `build-essential`, but I'm pretty sure you need `python` and `python-dev`.

PR #1044, opened 2017-03-27
Show alternative expressive empty-state in empty space when user is not SpaceDeveloper

adborden:
When the user is not a SpaceDeveloper, we don't want to prompt them to create an app if they don't have permission to. Instead, we show them tips about assigning roles and how permissions work.


msecret:
I think all these changes are now in https://github.com/18F/cg-dashboard/pull/1041

adborden:
Yepp

PR #1046, opened 2017-03-27
Modify global errors to display just 1

msecret:
Because user's were often reporting angst and confusion when more
then one error was appearing.

### DIFF
--- a/static_src/main.js
+++ b/static_src/main.js
@@ -24,6 +24,8 @@ import routeActions from './actions/route_actions.js';
 import spaceActions from './actions/space_actions.js';
 import serviceActions from './actions/service_actions.js';
 import SpaceContainer from './components/space_container.jsx';
+import { appHealth } from './util/health.js';
+import { entityHealth } from './constants.js';
 import { trackPageView } from './util/analytics.js';
 import uaaApi from './util/uaa_api.js';
 import userActions from './actions/user_actions.js';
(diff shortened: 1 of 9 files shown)


adborden:
I'm not sure we want to lower `maxErrors` on the store. When we hit that max, we replace every error in the store with a generic one. I think that will just make things more confusing to the user because we're taking specific error messages and then making them less actionable. Or maybe even misleading because the issue might not actually be connection related, but that's what the error says.

msecret:
So my thinking was that so far, when users are getting more then 1 errors, they're usually getting large failures. That being said, I'm down to bump this up to 3.

adborden:
Yeah, I agree, in fact probably a single error means something catastrophic is wrong, hence the prompt to refresh. Maybe we should just get rid of the replace logic, I feel like the component only showing one error is really the "right" fix.

It also makes me think that our error handling isn't quite right. Rather than having these global handlers on the fetch actions, errors for significant data should halt the page, preventing further errors. We are a little ways from that, but I think we want to start building more thoughtful error handling in that direction, rather than just letting things fall to the global error handler.

static_src/main.js
```diff
@@ -121,8 +124,11 @@ function app(orgGuid, spaceGuid, appGuid) {
   activityActions.fetchAppLogs(appGuid);
   quotaActions.fetchAll();
   appActions.changeCurrentApp(appGuid);
-  appActions.fetch(appGuid);
-  appActions.fetchStats(appGuid);
+  appActions.fetch(appGuid).then((res) => {
+    if (appHealth(res) === entityHealth.ok) {
+      appActions.fetchStats(appGuid);
```
adborden: Comment here would be nice. I'm also curious if a healthy app is the only one where we can fetch stats for.

static_src/main.js
```diff
@@ -176,6 +182,9 @@ const router = new Router(routes);
 router.configure({
   before: checkAuth,
   notfound: notFound,
-  on: () => trackPageView(window.location.hash)
+  on: () => {
+    errorActions.clearErrors();
```
adborden: Good call, I think `after` or `before` might be a better choice, why `on`?

msecret: oh yea, before seems to make sense.

PR #1051, opened 2017-03-29
Try at tests

msecret:
I want to make a request to the api server to change the variant so
the request data can change to not have apps. This would allow tests
to change conditions in the API server.

I can't get it to work. Doing an ajax request with axios from the
node context fails with ECONNREFUSED. Trying with fetch from inside
fails with "TypeError: Uncaught error: Cannot read property 'variant' of null"
that's returned from smocks running.

Is there a way to do this through the server API rather the ajax? How
would I even gain access to the server from here?

See https://github.com/18F/cg-dashboard/pull/1051/files#diff-9894cb6e373d34d674fd20c227303074R19

### DIFF
--- a/static_src/test/functional/expressive_empty_state.spec.js
+++ b/static_src/test/functional/expressive_empty_state.spec.js
@@ -1,0 +1,40 @@
+
+import http from 'axios';
+import dedent from 'dedent';
+
+import EmptyElement from './pageobjects/empty.element';
+import OrgQuicklookElement from './pageobjects/org_quicklook.element';
+
+describe('Expressive empty state', function () {
+  let appsEmptyState;
+
+  it('navigates to page', function () {
+    browser.url('/');
+  });
+
+  it('changes to app emtpy state', function(done) {
+    const server = browser.getUrl().split('/')[2];
+
+    const result = browser.executeAsync(function(url, done) {
+      var req = fetch(`http://${url}/_admin/api/route/space-summary`, {
+        method: 'POST',
+        body: { "variant": "space-summary-empty-state" }
+      })
+      .then((res) => { done({msg: 'success', data: res}) })
+      .catch((err) => { done({msg: 'error', data: err, req: req}) });
+    }, server);
+
+    console.log('result', result);
+    console.log(result.value);
+  });
+
+  describe('for apps on overview', function() {
+    beforeEach(function () {
+      appsEmptyState = new EmptyElement( browser, EmptyElement.primarySelector);
+    });
+
+    it('exists', function () {
+      expect(appsEmtpyState.exists()).toBe(true);
+    });
+  });
+});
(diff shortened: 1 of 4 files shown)


static_src/test/functional/expressive_empty_state.spec.js
```diff
@@ -0,0 +1,40 @@
...
+
+  it('navigates to page', function () {
+    browser.url('/');
+  });
+
+  it('changes to app emtpy state', function(done) {
+    const server = browser.getUrl().split('/')[2];
+
+    const result = browser.executeAsync(function(url, done) {
+      var req = fetch(`http://${url}/_admin/api/route/space-summary`, {
```
adborden: I would avoid using `fetch` here, I'm not sure it allows a JSON object for the body (I think you need to pass it a binary type like `Blob). I tried using axios here and it works fine.

static_src/test/functional/expressive_empty_state.spec.js
```diff
@@ -0,0 +1,40 @@
...
+import OrgQuicklookElement from './pageobjects/org_quicklook.element';
+
+describe('Expressive empty state', function () {
+  let appsEmptyState;
+
+  it('navigates to page', function () {
+    browser.url('/');
+  });
+
+  it('changes to app emtpy state', function(done) {
```
adborden: You're not calling this `done` callback which I think is throwing some things off.

static_src/test/functional/expressive_empty_state.spec.js
```diff
@@ -0,0 +1,40 @@
...
+      .catch((err) => { done({msg: 'error', data: err, req: req}) });
+    }, server);
+
+    console.log('result', result);
+    console.log(result.value);
+  });
+
+  describe('for apps on overview', function() {
+    beforeEach(function () {
+      appsEmptyState = new EmptyElement( browser, EmptyElement.primarySelector);
```
adborden: FYI, `BaseElement` doesn't take selector strings as an argument, you need to wrap this with `browser.element(EmptyElement.primarySelector)`.

BUT I added support for selectors in https://github.com/18F/cg-dashboard/pull/1056 because it's a good idea :D

adborden:
We chatted about this in slack, but just for the record, webdriverio runs each spec file in parallel, so by using the smocks feature we're actually changing global state which will cause race conditions. I briefly tried to see what it would take to run a separate smocks server for each spec, but you can't override the webdriveio baseUrl at runtime. That means all the `url()` requests would need the full url. I tried writing a helper for this, but it started getting ugly, monkey patching webdriverio. I gave up before I got it working. I guess we're going against the grain by having multiple servers.

rememberlenny:
Anything I can do to help with this?

msecret:
I think we can table this for now, as I'm not sure we have the bandwidth to work on this. Will keep the branch open though.

PR #1052, opened 2017-03-31
Improved auth flow

adborden:
This should address these issues around auth:
- https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-4127
- https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-3830
- https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-3831

Generally, these issues come up because:
1. We're not specifying `path` when setting session cookies, so the user could have multiple session cookies depending on what route they hit.
1. We start the login flow anytime we detect the user is unauthorized. This results in multiple login flows starting at the same time if there are multiple requests to the server.
1. The client doesn't wait for the auth check before fetching data, resulting in a cascade of login redirects.
1. Any unauthorized request to the API results in a redirect to the login flow, causing login.fr.cloud.gov to return `403` due to failed CORS check because of our `x-csrf-token` header.

To address this, we:
1. Specify a `path` and `expiry` on our session tokens.
1. Move logout to a public URL so that even if you have an invalid token, you can still clear your session.
1. Move `LoginRequired` middleware to `secureRouter` so no whitelist is required and public URLs that were forgotten aren't stuck behind the middleware.
1. Return `401` when unauthorized and let the client figure out what to do.
1. On the client, wait for the checkAuth before loading the page. If unauthorized, start the login flow. If authorized, continue loading the page.

## Callouts

Specifically, I'd like to call out `checkAuth` as a critical section. We're using promises and have to differentiate between authorized, unauthorized, and error, each with different handling. Not sure if there's a cleaner way to write this code.

There's a note about caching with static assets, but I think moving the `LoginRequired` middleware fixes all this and makes our static assets cachable again. What I'm not sure of, is if we should version our javascript before shipping this to avoid folks getting stale javascript.

We're not doing anything to fix users in the stuck state. If they have multiple session cookies, some valid, some not, this will not unstuck them. They'll have to clear all their cookie data for the entire domain. In chrome, this must be done through the dev tools.

This is currently on https://dashboard-demo.app.cloud.gov/#/ to try out.


controllers/secure.go
```diff
@@ -35,6 +35,34 @@ func (c *SecureContext) OAuth(rw web.ResponseWriter, req *web.Request, next web.
...
+		next(rw, r)
+		return
+	}
+
+	// Don't cache anything
+	// right now, there's a problem where when you initially logout and then
+	// revisit the server, you will get a bad view due to a caching issue.
+	// for now, we clear the cache for everything.
+	// TODO: revist and cache static assets.
+	rw.Header().Set("cache-control", "no-cache, no-store, must-revalidate, private")
```
jcscottiii: did you try removing this to verify it could fix the caching problem? (could leave that for another PR)

static_src/stores/login_store.js
```diff
@@ -7,25 +7,44 @@ import BaseStore from './base_store.js';
...
+
       case loginActionTypes.RECEIVED_STATUS:
-        this._isAuthenticated = action.status;
+        this._isAuthenticated = action.authStatus.status === 'authorized';
         this.emitChange();
         break;
 
+      case loginActionTypes.ERROR_STATUS:
+        // Login status is unknown. If we have a login status, leave it as is
+        // and hope things go smooth. A notification is handled elsewhere.
```
jcscottiii: comment on where is `elsewhere`?

static_src/actions/login_actions.js
```diff
@@ -5,13 +5,40 @@
...
+import cfApi from '../util/cf_api';
 import { loginActionTypes } from '../constants';
 
-export default {
-  receivedStatus(status) {
+const loginActions = {
+  fetchStatus() {
+    AppDispatcher.handleViewAction({
+      type: loginActionTypes.FETCH_STATUS,
+      status
```
jcscottiii: should this be `authStatus`? (as a reactjs newbie, let me know so i can understand)

static_src/components/header.jsx
```diff
@@ -28,9 +28,9 @@ export default class Header extends React.Component {
       <Action href="/handshake" label="Login" type="outline">
         Login
       </Action>
-    </HeaderLink>:
+    </HeaderLink> :
```
jcscottiii: is there a reason for the extra space beyond the tag? (style question i don't know)

static_src/components/header.jsx
```diff
@@ -41,15 +41,18 @@ export default class Header extends React.Component {
           <a href="/" className={ this.styler('logo') } title="Home">
             <svg className={ this.styler('logo-img') }>
               <use
-                xlinkHref={ this.getImagePath('logo-dashboard') }>
+                xlinkHref={ this.getImagePath('logo-dashboard') }
```
jcscottiii: reason this line was broken up? (style question i don't know)

static_src/main.js
```diff
@@ -130,22 +143,72 @@ function app(orgGuid, spaceGuid, appGuid) {
...
+
+        // Stop the routing
+        next(false);
+
+        // Hang the promise chain to avoid additional loading and API calls
+        const hang = new Promise();
+        return hang;
+      }
+
+      if (!authStatus) {
```
jcscottiii: do you want to make this the first conditional? then you could simplify the other conditional, (i.e remove `authStatus` from the other conditional)

static_src/stores/login_store.js
```diff
@@ -7,25 +7,44 @@ import BaseStore from './base_store.js';
 import { loginActionTypes } from '../constants.js';
 
 
-class LoginStore extends BaseStore {
+export class LoginStore extends BaseStore {
   constructor() {
     super();
     this.subscribe(() => this._registerToActions.bind(this));
+    // TODO this should probably be false, but we need to account for the
```
jcscottiii: out of scope for this PR, but we should really look into using something like enums to get achieve something close to a tri state boolean.

controllers/root.go
```diff
@@ -34,7 +34,8 @@ func (c *Context) Ping(rw web.ResponseWriter, req *web.Request) {
 func (c *Context) LoginHandshake(rw web.ResponseWriter, req *web.Request) {
 	if token := helpers.GetValidToken(req.Request, c.Settings); token != nil {
 		// We should just go to dashboard if the user already has a valid token.
-		http.Redirect(rw, req.Request, "/#/dashboard", http.StatusFound)
+		dashboardURL := fmt.Sprintf("%s%s", c.Settings.AppURL, "/#/dashboard")
```
adborden: It's not needed, I prefer to be strict about what we send the client. It's the whole robustness principle "Be conservative in what you do, be liberal in what you accept from others". Unless there's a feature where we're supporting a whitelist of multiple allowed hosts, I think we should just use absolute URLs.

controllers/secure.go
adborden: I did not try. I'm not sure what the bad caching behavior was so I wasn't sure I would be able to spot it.

I'm also not really sure how the CF API handles caching. Will it respond with cache headers of it's own or respond appropriately if the client includes caching headers? We might want to update the TODO to be more about how the API should be cached (or not).

static_src/stores/login_store.js
adborden: Yeah, that sounds like a good idea. I'm fine with using `null` and just being explicit in the getters (`null ===true`, `null==false` both fail) but using enums would make this much more explicit.

static_src/stores/login_store.js
adborden: Yeah, it happens in `checkAuth`, but I might update the comment to be more like:
> If necessary, the action caller is responsible for notifying the user of an error.

That way the comment doesn't need to know where elsewhere is.

static_src/actions/login_actions.js
adborden: Ah, good catch. Since it's a fetch, we don't have any auth status, so this is undefined no matter what and should just be removed.

static_src/components/header.jsx
adborden: Just style. It's enforced by our eslint config.

static_src/components/header.jsx
adborden: Yeah, also for style, enforced by the eslint config. I'm not crazy about this rule, but that's what airbnb says.

static_src/main.js
adborden: Yeah, great idea, especially because its an error condition, I like having those at the top too. Test your invariants! ;)

static_src/util/cf_api.js
```diff
@@ -28,6 +51,36 @@ function handleError(err, errHandler = errorActions.errorFetch) {
   }
 }
 
+function parseError(resultOrError) {
```
msecret: Eventually, should all errors be parsed like this?

adborden: Yeah, I believe so because if the axios request is successful, we have a response with data from CF API (and maybe a non-200 status). But if there's a network issue, it would just be an error.

adborden: @msecret I'm still not quite sold on how we want to represent errors in the app. Because of the CF API, we expect most errors to have a `description` property, but most JS errors will only have `message` and we don't have consistent handling to move `description` to `message` or vice versa.

I'm starting to think we should just expect the standard `message`, and make sure to fill in something descriptive based on the specific error type we have. Kind of like how [this](https://github.com/18F/cg-dashboard/pull/1054/commits/6d6cc3b74b92a2dbdf0fcfb791b39ecb2f7e387e#diff-3fd0a313bfc0cb4679cc0af796e67a94R15) works, but to populate `message` instead.

I'm a fan of using custom error types (extending `Error`) so it's easier to change the error handling based on the type of error (using `instanceof Type` checks) but babel makes it a little awkward--you can't use `class extends` with native types apparently.

Any other thoughts?

PR #1061, opened 2017-04-17
Fix type that breaks org related permission changes

rememberlenny:
cc: @jcscottiii @msecret

### DIFF
--- a/static_src/components/users.jsx
+++ b/static_src/components/users.jsx
@@ -94,7 +94,7 @@ export default class Users extends React.Component {
   }
 
   get resourceGuid() {
-    const resourceGuid = this.state.current === ORG_NAME ?
+    const resourceGuid = this.state.currentType === ORG_NAME ?
       this.state.currentOrgGuid : this.state.currentSpaceGuid;
     return resourceGuid;
   }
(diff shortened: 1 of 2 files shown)


static_src/components/users.jsx
```diff
@@ -94,7 +94,7 @@ export default class Users extends React.Component {
   }
 
   get resourceGuid() {
-    const resourceGuid = this.state.current === ORG_NAME ?
+    const resourceGuid = this.state.currentType === ORG_NAME ?
```
msecret: This might be a good opportunity to add a unit test for this JSX component. Or a functional test. In the past we didn't do unit tests for components, but we have one or two examples now that do shallow rendering with enzyme.

rememberlenny: @msecret Updated with working test!

static_src/test/unit/components/users.spec.jsx
```diff
@@ -0,0 +1,56 @@
...
+  beforeEach(function () {
+    sandbox = sinon.sandbox.create();
+  });
+
+  afterEach(function () {
+    sandbox.restore();
+  });
+
+  describe('smoke test store', function () {
+    it('with UserStore', function () {
```
msecret: I don't think we need these smoke tests anymore

msecret (approved):
Just one comment but looking good

PR #1062, opened 2017-04-17
New admin check for permissions

rememberlenny:


msecret (changes requested):
I think we should add tests for the new user actions, the new case statement in the user store and the new method in the user store.

rememberlenny:
Added test for `new user actions`.

rememberlenny:
Added `case` conditions for `CURRENT_UAA_INFO_FETCH`.

rememberlenny:
Added the `isAdmin()` related checks.

rememberlenny:
Finished up the tests around the points mentioned. Good to go! @msecret

static_src/test/unit/stores/user_store.spec.js
```diff
@@ -433,6 +433,73 @@ describe('UserStore', function () {
     });
   });
 
+  describe('on CURRENT_UAA_INFO_RECEIVED', function () {
+    it('should not fail when groups is missing in uaaInfo', function () {
+      const currentUaaInfo = { guid: '1234' };
```
msecret: These tests in each `it` block seem to have repeated information. Is it possible to move the data to a `beforeEach` like you did below?

msecret (approved):
One comment about making tests less repeated.

PR #1067, opened 2017-04-19
Create the functional tests around user roles

rememberlenny:


static_src/test/functional/user_role.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    it('has a page header', function () {
+      const pageHeader = browser.element('.test-page-header-title');
+      expect(pageHeader.getText()).toBe('fake-cf-deck-testing');
+    });
+
+    describe('org manager for org X then they should', function () {
+      it('be able to edit roles for org X', function () {
+        browser.setCookie({ 'name': 'testing_user_role', 'value': 'org_manager_org_x' });
```
msecret: So here, it kinda sounds like you need to set the cooke, then reload or go the the org page. At that point you'll be in the correct state of the user being able to edit the roles for the org.

The user roles page element, should have an `isEditable` method which checks that the current UI is editable. So here you would select the correct user roles, and check that.

msecret:
Trying to think through the data you need:
- A user that's an OrgManager in a certain org.
- A user that's not an OrgManager in a certain org.
- A user that's a SpaceManager in a certain space.
- A user that's not a SpaceManager in a certain space.
- The user info and uaa info for each of those users.

In the code, you're creating new orgs, `fake-cf-user_role-org_x-testing`, `fake-cf-user_role-org_y-testing`. Is it possible to use the existing `fake-cf-deck-testing` and `fake-cf` instead? That might simplify the data side a bit. This is the same for spaces. In the space summaries fixture file, there was a lot of data that was removed for existing spaces, which might break some things.

msecret:
Another thing I'd suggest is to not create the data by hand, but attempt to re-create the cases in the actual dashboard data, and then copy that data to fixtures.

rememberlenny:
@msecret Thats a great suggestion.

I'd be interested to hear @adborden and @jcscottiii thoughts.

The role related conditions are actually really difficult to test, so the explicit guid and user references help significantly in debugging. I see the value in simplifying the tests around the number of orgs, but I also strongly feel there is future potential of too much technical debt in bundling tests with varying functionality.

I see the trade off for a smaller code base, but I feel it comes at the cost of future code legibility.

Any thoughts around this point?

static_src/test/server/api.js
```diff
@@ -139,17 +73,16 @@ module.exports = function api(smocks) {
...
-        name: "fake-personb@gsa.gov"
-      });
+      let userRoleObject;
+      if(req.state['testing_user_role'] && userRoles[req.state['testing_user_role']]){
+        userRoleObject = userRoles[req.state['testing_user_role']];
+      } else {
+        userRoleObject = userRoles['default'];
+      }
+      if (req.state['show_user_info']) {
+        // console.log(userRoleObject);
```
jcscottiii: remove this

static_src/test/server/api.js
```diff
@@ -313,7 +246,41 @@ module.exports = function api(smocks) {
...
+      let organizationUserRolesResponse, orgResponseName;
+      const guid = req.params.guid;
+      if ( organizationUserRoles[guid] ) {
+        organizationUserRolesResponse = organizationUserRoles[guid];
+        orgResponseName = guid;
+      } else {
+        organizationUserRolesResponse = organizationUserRoles['default'];
+        orgResponseName = 'default';
+      }
+      // console.log('Running org for: ', orgResponseName);
```
jcscottiii: remove this

PR #1072, opened 2017-04-21
Fix env var breaking go build

msecret:
We upgraded go recently. Now we're getting the error:

```
**ERROR** GO15VENDOREXPERIMENT is set, but is not supported by go1.7 and later.
Run 'cf unset-env <app> GO15VENDOREXPERIMENT' before pushing again.
**ERROR** Invalid vendor config: unsupported GO15VENDOREXPERIMENT
```

I don't think this env var should be set yet.


adborden:
Aw bummer. Sorry about that. I had Googled around to figure out if it was still necessary but couldn't find anything on it. I thought things were working locally, but I forgot that it's set by the manifest so I wasn't testing it properly.

It might be worth making sure that all the env variables set in manifest are also included in `env.sample`. (there would still be exceptions though for new relic and google analytics).

PR #1076, opened 2017-04-24
Add waits to tests breaking

msecret:
Still not sure why these are required now, but they fix the tests
on my computer.

### DIFF
--- a/static_src/test/functional/user_role.spec.js
+++ b/static_src/test/functional/user_role.spec.js
@@ -31,7 +31,6 @@ describe('User roles', function () {
     });
 
     it('has a page header', function () {
-      browser.screenshot();
       browser.waitForExist('.test-page-header-title', 2000);
       const pageHeader = browser.element('.test-page-header-title');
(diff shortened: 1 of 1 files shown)


static_src/test/functional/user_role.spec.js
```diff
@@ -31,6 +31,8 @@ describe('User roles', function () {
     });
 
     it('has a page header', function () {
+      browser.screenshot();
```
fureigh: @msecret Can you walk me through the purpose this line serves?

msecret: You're correct I can take that out

static_src/test/functional/user_role.spec.js
```diff
@@ -31,7 +31,6 @@ describe('User roles', function () {
     });
 
     it('has a page header', function () {
-      browser.screenshot();
       browser.waitForExist('.test-page-header-title', 2000);
```
rememberlenny: Can we do `browser.waitForExist('.test-page-header-title');`?

PR #1080, opened 2017-05-03
Create ability to make UAA invite user requests from dashboard

rememberlenny:


static_src/components/users_invite.jsx
```diff
@@ -0,0 +1,73 @@
+
```
msecret: I think this PR should focus on the backend and we should remove these frontend components for now

jcscottiii:
closed in favor of #1082 (closed for now) and #1083 (backend part)

PR #1086, opened 2017-05-04
Create an endpoint that sends an email to new cloud.gov users

rememberlenny:
screen shot 2017-05-04 at 6 20 33 pm


rememberlenny:
Need to add tests for:

- Query on invite endpoint without email (FAIL) 
- Query on invite endpoint without url (FAIL) 
- Query on invite endpoint when not logged in (FAIL) 
- Query on invite with all param (SUCCESS)

Need to add a proxy page to prevent links from being prematurely clicked by URL checker

msecret:
I noticed earlier you were trying to get mandrill or mailchimp access. Would we be putting the actual email in that system at some point, calling that API from here?

rememberlenny:
@msecret Thats an interesting idea that Im open to! On the UAA_Extras_app we were creating the email on the application itself.

controllers/uaa.go
```diff
@@ -49,3 +98,476 @@ func (c *UAAContext) UaaInfo(rw web.ResponseWriter, req *web.Request) {
...
+
+Then set up your cloud.gov access and get started.
+
+If you run into problems or have any questions, please email us at cloud-gov-support@gsa.gov.
+Thank you,
+The cloud.gov team
+`
+
+const inviteEmailTpl = `
+<html><head><base target="_top">
```
msecret: Where did the code come from for this email? This doesn't look fun to maintain, although emails in general are not fun to maintain and are hard to write.

I think this is why I was asking about moving this into Mailchimp. Usually those services give devs the ability to feed CSS and HTML separately and then the service will take care of inlining all the style rules which is required for an email. If this is how we usually do it, what are your thoughts on ensuring this email is maintainable in the future?

controllers/uaa.go
```diff
@@ -38,6 +42,51 @@ func (c *UAAContext) InviteUsers(rw web.ResponseWriter, req *web.Request) {
 	c.uaaProxy(rw, req, reqURL, true)
 }
 
+// SendInvite sends users an email with a link to the UAA invite
+func (c *UAAContext) SendInvite(rw web.ResponseWriter, req *web.Request) {
```
msecret: This function is pretty large and isn't unit tested. It might make sense to write some tests for specific functionality and start splitting up this function?

controllers/uaa.go
jcscottiii: this should be moved to a separate file

controllers/uaa.go
```diff
@@ -49,3 +98,476 @@ func (c *UAAContext) UaaInfo(rw web.ResponseWriter, req *web.Request) {
 		rw.Write([]byte("{\"status\": \"Bad request\", \"error_description\": \"Missing valid guid.\"}"))
 	}
 }
+
+// InviteEmail provides inviteTextTpl and inviteEmailTpl the value to access invite_url query
+type InviteEmail struct {
+	URL string
+}
+
+const inviteTextTpl = `
```
jcscottiii: does't look like we are using this, right?

circle.yml
```diff
@@ -20,6 +20,7 @@ dependencies:
     - sudo apt-get update; sudo apt-get install libicu52
     - go version
     - go get github.com/Masterminds/glide
+    - go get github.com/jordan-wright/email
```
jcscottiii: you add these to the glide files

circle.yml
```diff
@@ -32,6 +33,7 @@ test:
   pre:
     - go get github.com/axw/gocov/gocov
     - go get github.com/mattn/goveralls
+    - go get github.com/jordan-wright/email
```
jcscottiii: you add these to the glide files

controllers/routes.go
```diff
@@ -53,7 +53,8 @@ func InitRouter(settings *helpers.Settings, templates *template.Template) *web.R
 	uaaRouter.Middleware((*UAAContext).OAuth)
 	uaaRouter.Get("/userinfo", (*UAAContext).UserInfo)
 	uaaRouter.Get("/uaainfo", (*UAAContext).UaaInfo)
-	uaaRouter.Post("/invite_users", (*UAAContext).InviteUsers)
+	uaaRouter.Get("/invite/email/send", (*UAAContext).SendInvite)
+	uaaRouter.Post("/invite/users/create", (*UAAContext).InviteUsers)
```
jcscottiii: Usually, POST denotes something created and you don't need the /create path.

controllers/uaa.go
rememberlenny: @jcscottiii This is being used in the "text" content: https://github.com/18F/cg-dashboard/pull/1086/files#diff-f800b1e943e88b0a2d3081775265a1f2R56

rememberlenny: The file was hidden by a `click to view diff` link in the `files changed` tab.

rememberlenny:
@jcscottiii Updated the glide file!

server.go
```diff
@@ -53,8 +58,16 @@ func replaceEnvVar(envVars *helpers.EnvVars, envVar string, value interface{}) {
 			envVars.ClientSecret = stringValue
 		case helpers.NewRelicLicenseEnvVar:
 			envVars.NewRelicLicense = stringValue
-		case helpers.SessionKeyEnvVar:
```
jcscottiii: this shouldn't have been removed

rememberlenny: Good catch

controllers/routes.go
```diff
@@ -55,7 +55,7 @@ func InitRouter(settings *helpers.Settings, templates *template.Template, mailer
 	uaaRouter.Middleware((*UAAContext).OAuth)
 	uaaRouter.Get("/userinfo", (*UAAContext).UserInfo)
 	uaaRouter.Get("/uaainfo", (*UAAContext).UaaInfo)
-	uaaRouter.Get("/invite/email/send", (*UAAContext).SendInvite)
+	uaaRouter.Post("/invite/email", (*UAAContext).SendInvite)
```
jcscottiii: also, since we are doing a POST, we can just send a JSON payload now.

rememberlenny: Sounds better. Can you show an example how?

rememberlenny:
@jcscottiii Is this good to go?

jcscottiii:
No, we should do the POST requests with JSON instead of query parameters

rememberlenny:
Getting performance related test failures again. Rebased on master and re-pushed to try again. Ran tests without cache, but still getting performance test errors.

Example fail: https://circleci.com/gh/18F/cg-dashboard/1625

@msecret @jcscottiii

rememberlenny:
Working now.

jcscottiii:
Approving this given that mailer will be more tested soon after. (Use mailcatcher or MailHog)

PR #1087, opened 2017-05-09
Add Detailed Ping about components.

jcscottiii:
In order to make sure we have detailed info about uptime of dependent components, we want to report that via the /ping endpoint


jcscottiii:
To see how this works, check out the new Ping tests.

PR #1093, opened 2017-05-10
update to use deployer accounts

jcscottiii:
In order to easily update the credentials for the deployer, let's use our broker.

Also:
Prior to this, we had a dedicated user account `cgdeck-deployer` that did the deploys. Those creds may have been in TravisCI during the latest leak (even though it has been forever since we used Travis). But instead of trying to that account's password changed, we can just get the creds rotated whenever by using the deployer broker.

That `cgdeck-deployer` account has been deleted.

We also stop deploying to E/W

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,9 +181,15 @@ uaac client add <your-client-id> \
 
 
 ### CI
-This project uses CircleCI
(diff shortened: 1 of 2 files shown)


rememberlenny:
Thanks for doing this!!

PR #1101, opened 2017-05-12
Add the components and test associated to user actions for invite new user frontend

rememberlenny:


static_src/test/unit/stores/user_store.spec.js
```diff
@@ -173,6 +173,37 @@ describe('UserStore', function () {
     });
   });
 
+  describe('on org user association received', function() {
+    it('should emit a change event if data changed', function() {
+      var spy = sandbox.spy(UserStore, 'emitChange');
+
+      AppDispatcher.handleViewAction({
```
msecret: Is there a reason to not call the userOrgAssociationReceived action here rather then dispatching an event? It might be easier and more concise.

static_src/components/users_invite.jsx
```diff
@@ -0,0 +1,73 @@
+
+/**
+ * Renders a users pages that allows to switch between all users in a space
```
jcscottiii: not sure about this comment

rememberlenny: @jcscottiii Fixed up

static_src/test/functional/pageobjects/user_invite.element.js
```diff
@@ -0,0 +1,27 @@
+
+import BaseElement from './base.element';
+
+// https://www.martinfowler.com/bliki/PageObject.html
+//
+// Represents a UserRoleElement for making assertions against. This makes it
```
jcscottiii: `UserRoleElement` -> `UserInviteElement`

static_src/test/functional/user_invite.spec.js
```diff
@@ -0,0 +1,36 @@
...
+      userInviteElement = new UserInviteElement(browser, browser.element('.test-users-invite'));
+      browser.waitForExist('.test-users-invite');
+
+      expect(userInviteElement.isVisible()).toBe(true);
+    });
+
+    it('should be able to input content into invite form', function () {
+      userInviteElement.inputToInviteForm(email);
+      expect(email).toBe(userInviteElement.getInviteFormValue());
+    });
```
jcscottiii: did we want to test what happens upon successful submission?

rememberlenny: @jscottiii we do, but not yet. The UI isnt set, so the test would break soon.

rememberlenny: That being said, I can mock up the backend data and we could test that.

Ill do that.

static_src/components/users_invite.jsx
```diff
@@ -0,0 +1,73 @@
+
+/**
+ * Renders a form that allows org users to invite new users
+ * to cloud.gov
+ */
+
+import React from 'react';
+
+import { Form, FormText } from './form';
```
msecret: I think imports should be organized by:
- external libs (like react)
- space
- Capitalized names in alpha order
- capitalized names in alpha order

So it would be more like this:
```
import React from 'react';

import Action from './action.jsx';
import { Form, FormText } from './form';
import FormStore from '../stores/form_store';
import PanelDocumentation from './panel_documentation.jsx';
```

msecret: It's hard to enforce unfortunately, because I don't think our linter supports it. It's also not a perfect rule because of the `{ }` es6 syntax, but should at least be kinda organized if that makes sense?

static_src/components/users_invite.jsx
```diff
@@ -0,0 +1,73 @@
...
+    super(props);
+    this.state = {};
+    FormStore.create(USERS_INVITE_FORM_GUID);
+
+    this.validateString = validateString().bind(this);
+    this._onChange = this._onChange.bind(this);
+    this._onValidForm = this._onValidForm.bind(this);
+  }
+
+  _onChange() {
```
msecret: This should be removed if it doesn't do anything. As well as the bind on 28

static_src/components/users_invite.jsx
```diff
@@ -0,0 +1,73 @@
...
+import userActions from '../actions/user_actions';
+
+import { validateString } from '../util/validators';
+
+const USERS_INVITE_FORM_GUID = 'users-invite-form';
+
+export default class UsersInvite extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
```
msecret: If we're not using state, it's better sometimes to not set it at all here.

static_src/components/users_invite.jsx
```diff
@@ -0,0 +1,73 @@
...
+          />
+          <Action label="submit" type="submit">Invite new user</Action>
+        </Form>
+      </div>
+    );
+  }
+
+}
+
+UsersInvite.propTypes = { };
```
msecret: Although not all of our files do this, it helps to have propTypes and defaultProps as variables set up top and then set them equal here. This makes the component a little easier to read because it's propTypes are at the top. So

```
// top of file, below imports or so:
const propTypes = {};
const defaultProps = {};

// bottom of file, below component class

}

UsersInvite.propTypes = propTypes;
UsersInvite.defaultProps = defaultProps;
```

This is again one of the annoying things that I'm not sure we can configure our linter for, and a bunch of our files aren't doing it. It'd be nice to configure the linter to catch it and fix all the old files.

static_src/stores/user_store.js
```diff
@@ -66,6 +67,16 @@ export class UserStore extends BaseStore {
         break;
       }
 
+      case userActionTypes.USER_ORG_ASSOCIATED: {
+        const user = Object.assign({}, { orgGuid: action.orgGuid }, action.user);
+        this._inviteInputActive = true;
```
msecret: I'll figure this part out.

static_src/components/users_invite.jsx
rememberlenny: @msecret Can we pair over these things? This is all new for me.

static_src/components/users_invite.jsx
rememberlenny: 👍

static_src/components/users_invite.jsx
msecret: Definitely! I can also pass some reading resources to you:

- https://facebook.github.io/react/docs/typechecking-with-proptypes.html
- https://themeteorchef.com/tutorials/what-are-proptypes

static_src/test/functional/pageobjects/user_invite.element.js
```diff
@@ -0,0 +1,34 @@
...
+  inputToInviteForm(input) {
+    return this.element('.test-users_invite_name').setValue(input);
+  }
+
+  getInviteFormValue() {
+    return this.element('.test-users_invite_name').getValue();
+  }
+
+  submitInviteForm() {
+    this.element('[type="submit"]').click();
```
msecret: Why did you remove the wait here? I had this working

static_src/test/functional/pageobjects/user_invite.element.js
```diff
@@ -0,0 +1,39 @@
...
+  inputToInviteForm(input) {
+    return this.element('.test-users_invite_name').setValue(input);
+  }
+
+  getInviteFormValue() {
+    return this.element('.test-users_invite_name').getValue();
+  }
+
+  submitInviteForm() {
+    const self = this;
```
msecret: You don't need to set self anymore in es6. You can do this instead:

```
browser.waitUntil(() =>
  this.countNumberOfUsers() > existingUserCount
, 10000);
```

PR #1102, opened 2017-05-16
Add more env vars that can be loaded via UPS

jcscottiii:
Currently, there's no way to read sensitive SMTP creds via UPS. This PR fixes that.

Also, it's blocking successful deploys without this code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,10 +152,10 @@ In each space that you plan on deploying, you need to create a `user-provided-se
 Run:
 ```
 # For applications without New Relic monitoring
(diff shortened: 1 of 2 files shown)


msecret:
@jcscottiii would it also make sense for some sort of test to fail if env vars aren't present so it fails at build time? Or I guess if it's failing at deploy time that might be the same difference?

jcscottiii:
@msecret yeah, currently the app will fail to start without the envvars (which is happening now) and that's desired behavior.

The only pain point about this whole process is that the error was in the logs. But since we use zero-downtime-push, the failed app doesn't stick around so I can't run `cf logs app_name --recent`. I had to use logs.fr.cloud.gov and go back in time to see that logs.

PR #1103, opened 2017-05-23
Add a notifier for when roles are changing

rememberlenny:
saving


static_src/components/saving.jsx
```diff
@@ -0,0 +1,56 @@
+
```
msecret: What's the reason we couldn't combine this with the loading component, and allow that component to have text like "Saving"?

static_src/components/user_list.jsx
```diff
@@ -119,16 +121,25 @@ export default class UserList extends React.Component {
   }
 
   render() {
+    let savingContent;
+    let saving = <Saving />;
```
msecret: I don't think you need both `saving` and `savingContent`. You could do:

```js
let saving;

if (this.state.saving) {
  saving = <Saving text={ this.state.saving } />;
}
```

If `saving` is undefined when you try and render it, React will ignore it.

static_src/css/main.css
```diff
@@ -40,3 +40,24 @@ input[type="radio"] + label::before {
   display: none;
 }
 
+.test-user_list .complex_list-item.processing {
```
msecret: This has to be moved to `cg-style` in an appropriate component before we merge.

static_src/stores/user_store.js
```diff
@@ -338,6 +355,36 @@ export class UserStore extends BaseStore {
     return role;
   }
 
+  addRoleToSavingArray(roleId) {
```
msecret: So I feel liek this has created a whole new, complicated data structure to keep saving state when we could use additional data structure of of the array of users instead. I think I tried to explain this the other day but wasn't clear enough, sorry about that.

So we have a list of users in this store. on `USER_ROLES_DELETE` and `USER_ROLES_ADD`, just add a `saving: true` property to the user object with the guid of the user who's role is being deleted or added. Then on `USER_ROLES_DELETED`, `USER_ROLES_ADD` or an error condition, just set `saving: false` on that same user object. Now the data for whether a specific user is being saved is right on that user object.

This will work well with the UI because we'll probably put a loading indicator on each user object rather the the whole user list (Think if you have a long user list; you won't see the loading state at the top as your editing roles at the bottom). We may need a global saving state for actions that aren't tied to a specific user, like inviting a new user.

static_src/stores/user_store.js
```diff
@@ -338,6 +355,36 @@ export class UserStore extends BaseStore {
...
+    this._saving.push(roleId);
+    this.emitChange();
+  }
+
+  removeRoleFromSavingArray(roleId) {
+    // These timeouts prevent a flashed text change, respective to state.
+    const minTimeForSaving = 500;
+    const minTimeForSaved = 2000;
+    let index;
+    window.setTimeout(() => {
```
msecret: I'd look into if we could put timing related things in the JSX components rather then in the store. That way, there's no confusing about what the actual state is, the timing is at the view layer instead.

controllers/secure.go
```diff
@@ -4,6 +4,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	//"net/http/httputil"
```
jcscottiii: remove this

controllers/secure.go
```diff
@@ -109,6 +110,8 @@ func (c *SecureContext) submitRequest(rw http.ResponseWriter, req *http.Request,
 	if res != nil {
 		defer res.Body.Close()
 	}
+	//b, _ := httputil.DumpRequest(request, true)
```
jcscottiii: remove these

controllers/uaa.go
```diff
@@ -20,82 +24,267 @@ type UAAContext struct {
...
+	requestObj := inviteUAAUserRequest{[]string{inviteUserToOrgRequest.Email}}
+	inviteUAAUserBody, jsonErr := json.Marshal(requestObj)
+	if jsonErr != nil {
+		err = &UaaError{http.StatusInternalServerError,
+			[]byte("{\"status\": \"failure\", \"data\": \"" +
+				jsonErr.Error() + "\"}"),
+		}
+		return
+	}
+	log.Println("hmm")
```
jcscottiii: remove all log statements

controllers/uaa.go
```diff
@@ -20,82 +24,267 @@ type UAAContext struct {
...
+type InviteUserToOrgRequest struct {
+	Email string `json:"email"`
+}
+
+// ParseInviteUserToOrgReq will return InviteUserToOrgRequest based on the data
+// in the request body.
+func (c *UAAContext) ParseInviteUserToOrgReq(req *http.Request) (
+	inviteUserToOrgRequest InviteUserToOrgRequest, err *UaaError) {
+	err = readBodyToStruct(req.Body, &inviteUserToOrgRequest)
+	log.Println(inviteUserToOrgRequest)
```
jcscottiii: remove logs

controllers/uaa.go
```diff
@@ -20,82 +24,267 @@ type UAAContext struct {
...
+	log.Println("showme")
+	return
+}
+
+// InviteUserToOrg will invite user in both UAA and CF, send an e-mail and
+// associate then to the org.
```
jcscottiii: doesn't associate anymore, remove this last part of this line

rememberlenny:
Closing this PR in place of:
- https://github.com/18F/cg-dashboard/pull/1107
- https://github.com/18F/cg-dashboard/pull/1108

cc: @jcscottiii @msecret

PR #1104, opened 2017-05-24
Release 1.16.4

jcscottiii:


msecret:
Can we add release notes in the same fashion as previous ones too?

PR #1105, opened 2017-05-24
User invite error handling

msecret:
Adds error handling for user invite feature. Ensures at any part of the 3 step request process, that if something goes wrong, there's an error added in the form for the user.

### DIFF
--- a/static_src/components/users_invite.jsx
+++ b/static_src/components/users_invite.jsx
@@ -7,6 +7,7 @@
 import React from 'react';
 
 import Action from './action.jsx';
+import ErrorMessage from './error_message.jsx';
 import FormStore from '../stores/form_store';
 import { Form, FormText } from './form';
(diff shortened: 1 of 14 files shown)


static_src/components/users_invite.jsx
```diff
@@ -7,7 +7,6 @@
 import React from 'react';
 
 import Action from './action.jsx';
-import ErrorMessage from './error_message.jsx';
```
jcscottiii: is there a way to lint for unused dependencies?

rememberlenny: Good question. 

This looks like it does it: https://github.com/depcheck/depcheck

jcscottiii: when does it get called? i didn't see an error for it in the build before it was fixed.

PR #1108, opened 2017-06-01
Resolve admin permission issues in creating new user through user invite

rememberlenny:


static_src/stores/user_store.js
```diff
@@ -67,18 +68,27 @@ export class UserStore extends BaseStore {
...
         this.emitChange();
         break;
       }
 
-      case userActionTypes.USER_ORG_ASSOCIATED: {
-        const user = Object.assign({}, { orgGuid: action.orgGuid }, action.user);
-        this._inviteInputActive = true;
+      case userActionTypes.USER_ASSOCIATED_ORG_DISPLAYED: {
+        let orgUser;
+        action.orgUsers.map(function(item){
```
msecret: A potentially easier way to do this could be to use es6's `[].find`.

```js
const orgUser = action.orgUsers.find((orgUser) => orgUser.guid === action.userGuid);
```

That finds the user when it's guid is equal to the action user guid.

static_src/components/users_invite.jsx
```diff
@@ -16,21 +16,31 @@ import { validateString } from '../util/validators';
 const USERS_INVITE_FORM_GUID = 'users-invite-form';
 
 const propTypes = {
+  inviteDisabled: React.PropTypes.bool,
+  currentUserAccess: React.PropTypes.bool,
   error: React.PropTypes.object
 };
 const defaultProps = {};
 
+function stateSetter(props) {
```
msecret: I don't think you need to use state setter at all, unless I'm missing something. You can access error through `this.props.error` like the other things, and don't have to set state at all.

msecret: An example component that doesn't have state at all, and is controlled through props: https://github.com/18F/cg-dashboard/blob/master/static_src/components/org_quicklook.jsx.

static_src/actions/user_actions.js
```diff
@@ -221,6 +209,37 @@ const userActions = {
...
+
+    const orgGuid = OrgStore.currentOrgGuid;
+
+    AppDispatcher.handleViewAction({
+      type: userActionTypes.CURRENT_USER_ROLES_FETCH,
+      userGuid,
+      orgGuid
+    });
+
+    return cfApi.fetchOrgUserRoles(orgGuid)
```
msecret: This request is already happening in another action. Rather then calling the same thing twice, we should use that action to do what we need.

PR #1112, opened 2017-06-07
Fix user management permission issues

msecret:
Before:
The user objects in `UserStore` had two different data structures related to permissions:

1. The `organization_roles` / `space_roles` array with each role in the array
1. The `roles` object with an array of roles by org / space guid.

This is an example user object:

```js
{
  guid: 'asdflkja',
  organization_roles: ['org_manager'],
  roles: {
    'asdfkjlsdf': ['org_manager']
  }
}
```

The organization/space_roles property was used for displaying what roles a user had checked off on the org and space page. This array of roles was completely replaced each time a new org or space page was loaded, which lead to some potential bugs. The `roles` property with org/space guid was used to check the current user's role. This was incorrectly set based on prior assumptions about how CF api returned things.

This PR changes all role logic to use the 2nd, `roles` with org/space guid data structure. Additionally, it now uses this roles property to get all the users that belong to an org or space. Benefits of this:

- Keeps all role logic in one data structure rather then duplicate logic in different places which can easily get confusing.
- Fixes some bugs related to all users having org manager roles on all orgs. Fixes potential problems with user roles being set strangely based on space and org routes loading for users.

### DIFF
--- a/static_src/actions/user_actions.js
+++ b/static_src/actions/user_actions.js
@@ -88,7 +88,7 @@ const userActions = {
       entityType
     });
 
-    api(
+    return api(
       userGuid,
       entityGuid,
       roles
--- a/static_src/actions/user_actions.js
+++ b/static_src/actions/user_actions.js
@@ -73,19 +73,19 @@ const userActions = {
     });
   },
 
-  addUserRoles(roles, userGuid, entityGuid, resourceType) {
+  addUserRoles(roles, userGuid, entityGuid, entityType) {
     const apiMethodMap = {
       org: cfApi.putOrgUserPermissions,
       space: cfApi.putSpaceUserPermissions
     };
-    const api = apiMethodMap[resourceType];
+    const api = apiMethodMap[entityType];
 
     AppDispatcher.handleViewAction({
       type: userActionTypes.USER_ROLES_ADD,
       roles,
       userGuid,
       entityGuid,
-      resourceType
+      entityType
     });
 
     api(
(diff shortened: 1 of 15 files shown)


rememberlenny:
This is GREAT functionality. I didnt quite get what the two did or why the functions in CF were so different until coming in contact with these problems. Thanks for building this abstraction.

static_src/actions/user_actions.js
```diff
@@ -26,9 +26,9 @@ const userActions = {
     });
   },
 
-  fetchSpaceUsers(spaceGuid) {
+  fetchSpaceUserRoles(spaceGuid) {
```
rememberlenny: This is great. I was confused as to the naming convention broke from the tradition here.

static_src/actions/user_actions.js
```diff
@@ -332,8 +318,6 @@ const userActions = {
       .then(userInfo =>
         Promise.all([
           userActions.fetchUser(userInfo.user_id),
```
rememberlenny: This looks great

static_src/components/users.jsx
```diff
@@ -110,11 +114,12 @@ export default class Users extends React.Component {
     }
 
     let content = (<UserList
-      initialUsers={ this.state.users }
-      initialUserType= { this.state.currentType }
-      initialCurrentUserAccess={ this.state.currentUserAccess }
-      initialEmpty={ this.state.empty }
-      initialLoading={ this.state.loading }
+      users={ this.state.users }
```
rememberlenny: Thanks for making this port

rememberlenny:
Im gonna hit the bed, but this looks good to merge. I'd like to pull off this for my branch.

msecret:
@rememberlenny let's try and work together tomorrow to merge that one in.

static_src/routes.js
```diff
@@ -93,7 +93,8 @@ export function space(orgGuid, spaceGuid, next) {
   spaceActions.fetch(spaceGuid);
   serviceActions.fetchAllInstances(spaceGuid);
   userActions.changeCurrentlyViewedType('space_users');
-  userActions.fetchSpaceUsers(spaceGuid);
+  userActions.fetchOrgUsers(orgGuid);
```
jcscottiii: to be consistent, is there a reason why this wasn't also renamed to be `fetchOrgUserRoles`?

jcscottiii: nevermind, it already exists

jcscottiii:
LGTM. just had one question.

static_src/routes.js
msecret: So I found that there's no CF API equivalent to `fetchSpaceUsers` like `fetchOrgUsers`. So essentially, you have to fetch org users and then fetch space user roles to assign all the users to specific roles for a space.

msecret: I think `fetchOrgUsers` has some more data the `fetchOrgUserRoles`

rememberlenny:
Merged into https://github.com/18F/cg-dashboard/pull/1115

PR #1117, opened 2017-06-13
Fix user management

msecret:
The mapping in role control is no longer needed based on how actions work, so removed. This fixes issue where modifying a role doesn't work.

Speaking of which, there's many functional tests for modifying roles, why didn't any of them fail when the previous change broken them?


jcscottiii:
@msecret there aren't any current functional tests that test the checking of boxes. i just made some to test this PR and it works.  i also just tried it against master (to make sure it should fail and the tests catch it)

here's a screenshot of it running against master and failing:
image

i'll push these tests to this branch

PR #1119, opened 2017-06-14
Fix org and space role association

jcscottiii:
### First commit: Fix functional test server to check for correct roles
This modification makes sure that all requests to do
association/dissociation of roles use the right api key for the role.

This will break tests because currently the code doesn't use the right
api key for role modification but will be fixed in following commits.

### Other commits: fixing the api calls and tests

### DIFF
--- a/static_src/stores/user_store.js
+++ b/static_src/stores/user_store.js
@@ -10,20 +10,6 @@ import BaseStore from './base_store.js';
 import cfApi from '../util/cf_api.js';
 import { userActionTypes } from '../constants.js';
 
-// TODO why is this role mapping needed?
-const resourceToRole = {
-  space: {
-    managers: 'space_manager',
-    developers: 'space_developer',
-    auditors: 'space_auditor'
-  },
-  org: {
-    managers: 'org_manager',
-    billing_managers: 'billing_manager',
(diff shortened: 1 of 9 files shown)


static_src/stores/user_store.js
```diff
@@ -164,6 +150,7 @@ export class UserStore extends BaseStore {
         const orgPermissionsReq = cfApi.deleteOrgUserPermissions(
           action.userGuid,
           action.orgGuid,
+          'users',
```
rememberlenny: @jcscottiii This might be a duplicated line

rememberlenny: https://github.com/18F/cg-dashboard/blob/0ed527b4fe52d824abe738d9b885a0ef0256c6d7/static_src/stores/user_store.js#L153-L154

jcscottiii: it's meant to be duplicated but let me re-work it so it's more clear as to why it's like that.

rememberlenny: Got it

rememberlenny: Looks good overall.

rememberlenny:
@jcscottiii Merged because it does what it is supposed to do.

PR #1122, opened 2017-06-14
Break a >100-character line into two <=100-character lines

fureigh:
Fixes an error thrown by `npm run lint`. Evidently the linter dislikes lines that contain more than 100 characters.


jcscottiii:
just checked CircleCI stopped reporting (but they still ran) on a few other PRs so this managed to slip in. thanks for fixing this!

PR #1123, opened 2017-06-15
[Backend] Add the user check to see if is verified

rememberlenny:


jcscottiii:
It passed. I don't know why it didn't finish reporting. image

PR #1124, opened 2017-06-15
Create the frontend notification when a user is invited/associated to an org

rememberlenny:
click existing

existing


static_src/components/users_invite.jsx
```diff
@@ -13,6 +13,9 @@ import PanelDocumentation from './panel_documentation.jsx';
 import userActions from '../actions/user_actions';
 import { validateString } from '../util/validators';
 
+import style from 'cloudgov-style/css/cloudgov-style.css';
```
jcscottiii: swap the order. `createStyler `should come before `style` alphabetically.

static_src/stores/user_store.js
```diff
@@ -61,6 +61,20 @@ export class UserStore extends BaseStore {
         break;
       }
 
+      case userActionTypes.USER_INVITE_STATUS_UPDATED: {
```
jcscottiii: Do we eventually want to have the similar functionality for inviting users to spaces? If so, we should pass an additional parameter to be able to determine which level (org vs space) are we inviting a user. I'm also fine with leaving it simple like this but just change the action to something like USER_ORG_INVITE_STATUS_UPDATED so that we don't handle the conditional logical in this case statement and we will later create a USER_SPACE_INVITE_STATUS_UPDATED action and just handle which one to send in `receivedInviteStatus` by passing an additional argument there.

after i just typed that, i'm leaning towards the latter.

with all that said, this PR is just for the org but would want it in a way so we can easily throw the space one together if we do plan on that.

thoughts?

jcscottiii:
also, we need tests

rememberlenny:
@jcscottiii I made some updates here to adjust for the org/space discrepancy. 

I created a reference object, that could be moved into the constants, where the language around singular/plural space/orgs can be kept. This is similar to the Rails way of having reference words based on the model names.

You can see it here: https://github.com/18F/cg-dashboard/pull/1124/files#diff-7265241a23b5b3264748ca4d293cf559R72

What do you think of this solution?

static_src/stores/user_store.js
```diff
@@ -61,6 +61,43 @@ export class UserStore extends BaseStore {
         break;
       }
 
+      case userActionTypes.USER_INVITE_STATUS_DISMISSED: {
+        this._inviteNotice = Object.assign({}, "", {
+          contextualMessage: ""
```
jcscottiii: by just reducing the text length to 0, it clears it dismisses the box?

static_src/components/users_invite.jsx
```diff
@@ -93,6 +119,14 @@ export default class UsersInvite extends React.Component {
     } else {
       content = '';
     }
+    if (this.props.notice) {
```
jcscottiii: currently this notice only comes up during the invite process. you might want to move this up to the conditional where we only display this invite notice upon have the right access.

jcscottiii:
@rememberlenny that works for me

static_src/stores/user_store.js
rememberlenny: Changed these up

static_src/components/users_invite.jsx
rememberlenny: Changed these up

static_src/actions/notification_actions.js
```diff
@@ -0,0 +1,29 @@
+
+/*
```
msecret: For this UI, I don't think notification actions and another notification store is required.

We used a separate error store and actions for global errors, relating to the whole application, so one store could decide how to render them. For the notifications here, they're all local to just the user invite section. This means the data for the notification should be stored in the UserStore, not a new notification store.

With this design as it stands, it's currently not separating the notifications in any way. So if the user invite adds a notification and then the app page adds a notification, there's no way of knowing that the one notification belongs on the user page and another belongs on the app page. Separating them would require adding more complex functionality to the notification store for each notification and where in the UI it would belong. Rather then creating a complex data structure like that, keeping a `inviteNotification` variable on the user store would be much simpler to update and access.

If we want global notifications, we should  probably refactor the error store/actions so you can specify the seriousness of the notification (really the error store just holds notifications for global errors occurring) rather then create a new store that's doing pretty much the same thing.

rememberlenny: @msecret Sounds good.

I initially did a local action/userStore based notification. I hit a problem because of the discard notification dispatch event being called during another dispatch event. Specifically, when I had a close button on the notification, the component was choking on render.

I looked into a solution, and all I saw were adding timeout related solutions, so I thought it would be better to have an external store manage the potential notifications associated to cards.

static_src/components/users.jsx
```diff
@@ -22,9 +20,10 @@ const ORG_NAME = 'org_users';
 function stateSetter() {
   const currentOrgGuid = OrgStore.currentOrgGuid;
   const currentSpaceGuid = SpaceStore.currentSpaceGuid;
-  const currentType = UserStore.currentViewedType;
+  const currentType = UserStore.currentlyViewedType;
```
jcscottiii: why was this renamed?

rememberlenny: @jcscottiii I think it was broken. It didnt refer back to anything.

static_src/test/unit/stores/user_store.spec.js
```diff
@@ -715,6 +715,50 @@ describe('UserStore', function () {
     });
   });
 
+  describe('on USER_INVITE_STATUS_DISPLAYED', function() {
+    let notice;
+
+    beforeEach(function() {
+      notice = { noticeType: "finish", description: "There was no cloud.gov account found for undefinedor the user has not verified their account by logging in.They have been sent an email cloud.gov invitation. Their account has been associated to this space and their space roles can be controlled below." };
```
jcscottiii: `undefinedor` - > `undefined or`

static_src/constants.js
```diff
@@ -22,6 +22,8 @@ const entityHealth = keymirror({
   warning: null,
   // Something is definitely not right and requires some attention
   error: null,
+  // Entity process was successful
+  finish: null,
```
jcscottiii: let's change this to `success`

rememberlenny: The styles for `finish` are already in the style library. Im happy to use success, but then we need to make a number of additional changes.

Do you think we should do that?

jcscottiii: leave it as-is, we should revisit it later though along with cg-stlye. because finish != success in all cases.

rememberlenny: 👍

static_src/test/unit/stores/user_store.spec.js
rememberlenny: Fixed this in the following commits.

static_src/stores/user_store.js
```diff
@@ -337,6 +354,23 @@ export class UserStore extends BaseStore {
     return this.get(this._currentUserGuid);
   }
 
+  get currentlyViewedType() {
+    return this._currentViewedType;
+  }
+
+  get viewTypeNouns() {
```
jcscottiii: clean this up. it's in two places

PR #1125, opened 2017-06-21
Lkb email validator

rememberlenny:
screen shot 2017-06-21 at 1 55 39 pm
screen shot 2017-06-21 at 1 55 45 pm

### DIFF
--- a/static_src/util/validators.js
+++ b/static_src/util/validators.js
@@ -43,7 +43,7 @@ export function validateEmail() {
   return function _validateEmail(value, name) {
     if (!(/^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/.test(value))) {
       return {
-        message: `The ${name || ''} field was not filled out`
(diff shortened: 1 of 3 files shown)


jcscottiii:
Let's change the wording to "The value entered is not a valid e-mail address."

jcscottiii:
What about a test of the user invite component? Would be helpful in case the validator was ever renamed or something.

rememberlenny:
@jcscottiii Updated

static_src/util/validators.js
```diff
@@ -31,7 +31,19 @@ export function validateString() {
   return function _validateString(value, name) {
     if (!value || !value.length) {
       return {
-        message: `The ${name || ''} field was not filled out`
+        message: 'The value entered is not a valid e-mail address'
```
jcscottiii: also, i think you changed the message for the other validator

rememberlenny: @jcscottiii Fixed

rememberlenny:
@jcscottiii Finally went green

PR #1130, opened 2017-06-22
Force speed test results to true

rememberlenny:


jcscottiii:
https://github.com/18F/cg-dashboard/blob/master/circle.yml#L43

`cd $WS && npm run test-performance` -> `cd $WS && npm run test-performance || true`

jcscottiii:
i wouldn't mess with the tests themselves, just the the command that runs the tests. we will come back and turn these on so someone can find it easily.

PR #1131, opened 2017-06-22
Only show user invite on org space

rememberlenny:


static_src/test/functional/user_invite.spec.js
```diff
@@ -53,4 +58,22 @@ describe('User roles', function () {
...
+      browser.url(urlOrgXSpaceXX);
+      browser.waitForExist('.test-users');
+    });
+
+    it('should not have the user invite panel', function () {
+      userRoleElement.setAndGetUserRole(cookieManagerOrgXSpaceXX);
+      browser.url(urlOrgXSpaceXX);
+
+      browser.waitForExist('.test-users');
+      var ells = $$('.test-users-invite');
```
jcscottiii: we can use the `waitForExist` with the reverse flag. http://webdriver.io/api/utility/waitForExist.html

jcscottiii: that would probably help with the lint errors as well

rememberlenny: Updated!

rememberlenny:
This should go green and is finished @jcscottiii @el-mapache @fureigh

PR #1133, opened 2017-06-23
Display UI error when user attempts to add/remove a roles without proper permissions

el-mapache:
* Outputs error message via UI when user attempts to change a role while they do not have permission.

screen shot 2017-06-26 at 12 34 53 pm

Currently, this error appears at the top of the user panel — it would make more sense as an inline error, but that will require some additional refactoring of the `error` and `user_role` components. 

* Displays a browser alert when a user attempts to remove their own org manager permissions.

screen shot 2017-06-26 at 12 38 26 pm

* Some minor code refactoring

### DIFF
--- a/static_src/components/user_provider.jsx
+++ b/static_src/components/user_provider.jsx
@@ -1,0 +1,43 @@
+import React from 'react';
+import UserStore from '../stores/user_store.js';
+
+const userProvider = Component => {
+  class UserProvider extends React.Component {
+    getChildContext() {
+      return { currentUser: this.state.currentUser };
+    }
+
+    constructor(props) {
+      super(props);
+
+      this.state = { currentUser: null };
+      this.onChange = this.onChange.bind(this);
+    }
+
+    componentDidMount() {
+      UserStore.addChangeListener(this.onChange);
+    }
+
+    componentWillUnmount() {
+      UserStore.removeChangeListener(this.onChange)
+    }
+
+    onChange() {
+      this.setState({
+        currentUser: UserStore.currentUser
+      })
+    }
+
+    render() {
+      return <Component {...this.props} />
+    }
+  }
+
+  UserProvider.childContextTypes = {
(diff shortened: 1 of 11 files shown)


static_src/components/user_provider.jsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
```
jcscottiii: _please excuse my non-react knowledge_. what's the purpose of this file? i see it keeps the state of the currentUser but couldn't we just have the component get the currentUser from the store directly?

el-mapache: Information on the currently logged in user is something that I think can safely be made global to the entire application. Otherwise, for every component that cares about the current user, we'd have to add a change listener to the store, set state, etc. That gets a bit cumbersome.

In general I think we should be moving towards having less direct store access in components and more props being passed down from container components. I think that was the direction marco wanted to start moving in as well ( see the **container component** section in CONTRIBUTING.md).

In this case, I wasn't sure exactly what component would be the container (likely `Users`), and the currentUser prop would probably end up getting passed down through about 5 components. We can definitely revisit this later whenever we refactor the code to make use of containers more heavily!

jcscottiii: Makes sense! thanks for that!

jcscottiii:
could you add a functional test?

el-mapache:
On it!

el-mapache:
@jcscottiii tests added!

static_src/constants.js
```diff
@@ -252,6 +252,7 @@ const userActionTypes = keymirror({
   USER_ROLES_DELETE: null,
   // Action when user roles are deleted on the server.
   USER_ROLES_DELETED: null,
+  USER_ROLE_CHANGE_ERROR: null,
```
jcscottiii: just add a comment here!

el-mapache: on it!

PR #1135, opened 2017-06-23
Correct validator to use rfc822 email validation

rememberlenny:


rememberlenny:
Fixed the tests here. Ready for review.

static_src/util/validators.js
```diff
@@ -41,7 +41,7 @@ export function validateString() {
 
 export function validateEmail() {
   return function _validateEmail(value, name) {
-    if (!(/^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/.test(value))) {
+    if (!(/^[a-zA-Z0-9.!#$%&amp;'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/).test(value)) {
```
el-mapache: I might make the regex itself a const outside the function so we don't have to create it each time we call the function, and as a bonus we can give it a readable name.

rememberlenny:
@el-mapache Great point. I just stored this as a variable.

static_src/util/validators.js
```diff
@@ -41,7 +41,8 @@ export function validateString() {
 
 export function validateEmail() {
   return function _validateEmail(value, name) {
-    if (!(/^[a-zA-Z0-9.!#$%&amp;'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/).test(value)) {
+    const emailRegex = /^[a-zA-Z0-9.!#$%&amp;'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/;
```
el-mapache: We should move this outside the exported function, this will still re-parse the regex every time the function is called. Its not a big deal performance-wise but I think its good practice

rememberlenny: @el-mapache Just moved up

PR #1138, opened 2017-06-27
Clear `saving` message on error, specs

el-mapache:
**WIP**
* Previously, the `saving` message was sticking around after a user attempted to change a role, but had the action fail. This PR clears that message on failure.

### DIFF
--- a/static_src/stores/user_store.js
+++ b/static_src/stores/user_store.js
@@ -189,9 +189,11 @@ export class UserStore extends BaseStore {
       }
 
       case userActionTypes.USER_ROLE_CHANGE_ERROR: {
+        this._saving = false;
         this._error = Object.assign({}, action.error, {
           description: action.message
         });
+
         this.emitChange();
         break;
       }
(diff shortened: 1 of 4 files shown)


jcscottiii:
nice! i would just add a functional test to make sure we don't still see it! otherwise LGTM

el-mapache:
@jcscottiii I think a functional test case for this scenario is going to be pretty difficult. We would need to open a second browser window with a different logged in user, toggle the role for the first user, then attempt to have the first user try and remove their own org manager role.

I think it would be pretty straightforward to create a `user_list` spec file and add a test there that verifies that the badge is rendered when that component's `saving` prop is set to true, and doesn't render when it is false.

Let me know if that seems reasonable!

jcscottiii:
@el-mapache yep! that totally makes sense!

el-mapache:
@jcscottiii Added a couple of specs to test for the existence of the 'Saving' badge. Let me know what you think.

PR #1143, opened 2017-06-28
Clear notifications upon changing pages

jcscottiii:


el-mapache:
Is the clearErrors() function ever called? I see that `errorActions.clearErrors()` is called in `clearErrors` in the `routes.js` file, but I don't see any reference to the function exported from routes getting called in the code.

jcscottiii:
it's a middleware that runs beforehand https://github.com/18F/cg-dashboard/blob/1.16.5/static_src/main.js#L21

PR #1145, opened 2017-06-28
Bump version to 1.16.7

jcscottiii:
This is to just include the clearing of notifications on user store. I rather not re-push on 1.16.6 since it wasn't technically broken


rememberlenny:
How I feel about this release:

PR #1147, opened 2017-06-29
Remove superfluous props assignment in constructor

el-mapache:
Props get assigned automatically under the hood, and super(props)
handles the case wehre this.props needs to be accessed directly in the
constructor


jcscottiii:
I read some material and it matches that the props assignment already happens

PR #1148, opened 2017-06-30
Break up the render method logic into functions

el-mapache:
* Adds two components to represent a link and button.
* Removes nested if/else statements in favor of smaller named functions

### DIFF
--- a/static_src/test/unit/components/action.spec.jsx
+++ b/static_src/test/unit/components/action.spec.jsx
@@ -4,57 +4,85 @@ import '../../global_setup.js';
 import React from 'react';
 import { shallow } from 'enzyme';
 import Action from '../../../components/action.jsx';
+import Link from '../../../components/action/link.jsx';
+import Button from '../../../components/action/button.jsx';
 
 describe('<Action />', function () {
-  let action, sandbox;
+  let action;
 
-  beforeEach(function () {
-    sandbox = sinon.sandbox.create();
+  describe('default behavior', () => {
+    it('returns a button', () => {
+      action = shallow(<Action />);
+
+      expect(action.find(Button).length).toBe(1);
+    });
   });
 
-  afterEach(function () {
-    sandbox.restore();
-  });
-
-  describe('given type is link', function () {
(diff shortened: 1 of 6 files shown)


msecret:
Great idea @el-mapache! Looking much cleaner

el-mapache:
Thanks! I'm hoping to hit up a couple more spots in the code this week as well.

el-mapache:
@jcscottiii @rememberlenny I commented out one of the specs that tests that the click handler is called from the child component of <Action/>.  I can't get it to pass, although manual testing verifies that it works. If ya'll have any ideas for why it might be failing, they would be much appreciated.

static_src/test/unit/components/action.spec.jsx
```diff
@@ -4,57 +4,85 @@ import '../../global_setup.js';
...
-    describe('on click', function () {
-      beforeEach(function () {
-        action.simulate('click');
-      });
-
-      it('triggers clickHandler', function () {
-        expect(clickHandlerSpy).toHaveBeenCalledOnce();
-      });
+    xit('triggers clickHandler', () => {
+      action.find(Button).simulate('click');
```
msecret: Can't think of why this wouldn't work besides that it's a simulated click. But even then, there's an exact example in the enzyme docs with sinon, pretty much just like this.

I wouldn't say this is the most important thing to test as I don't think there are any logic branches in the functionality that calls the click handler

jcscottiii:
I switched to using `mount` instead of `shallow` then it worked. no clue why though.

el-mapache:
`mount` renders all the child components as well, I was hoping to avoid it, but maybe we have to use it in this case! Thanks for taking a look.

el-mapache:
@jcscottiii is this good to be merged?

el-mapache:
@jcscottiii All updated!

jcscottiii:
one more thing, can we use `mount` so we can unskip this test? https://github.com/18F/cg-dashboard/pull/1148/files#diff-2a387c4f08d0883724c81e45f5f494a4R83

PR #1149, opened 2017-07-03
Move/dedupe logic in routes

el-mapache:
* Remove duplicate call to spaceActions.fetchAll
* Move calls to fetch all orgs and spaces to overview, it is the only
place the user needs that information

### DIFF
--- a/static_src/components/overview_container.jsx
+++ b/static_src/components/overview_container.jsx
@@ -92,21 +92,23 @@ export default class OverviewContainer extends React.Component {
       </span>
     );
 
-    if (state.empty) {
-      content = this.emptyState;
-    } else if (!state.loading && this.state.orgs.length > 0 || this.anyOrgsOpen()) {
-      content = (
-        <div>
-        { state.orgs.map((org) =>
-          <div key={ org.guid } className="test-panel-row-organizations">
-            <OrgQuicklook
-              org={ org }
-              spaces={ this.orgSpaces(org.guid) }
-            />
+    if (!state.loading) {
+      if (state.empty) {
+        content = this.emptyState;
+      } else if (this.state.orgs.length > 0 || this.anyOrgsOpen()) {
+        content = (
+          <div>
+          { state.orgs.map((org) =>
+            <div key={ org.guid } className="test-panel-row-organizations">
+              <OrgQuicklook
+                org={ org }
+                spaces={ this.orgSpaces(org.guid) }
+              />
+            </div>
+          )}
           </div>
-        )}
-        </div>
-      );
+        );
+      }
     }
 
     return (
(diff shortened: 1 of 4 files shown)


el-mapache:
This PR has a couple of broken of tests, but I want to get people's opinions on the changes before I spend the time to fix them. It seems like we didn't need to be making calls to fetch all orgs and all spaces on every request, but I could definitely be missing some important aspect of the app.

jcscottiii:
@el-mapache i would refer to the comment on this commit for the subset of spaces part: https://github.com/18F/cg-dashboard/commit/1ad3a0faf58609aca92f3293ff15cc871856627f

I think that's why spacesFetch was called twice (but I think you can just do it once then just filter by the 10 apps)

Everything else seems fine to me.

el-mapache:
@jcscottiii I took a look at the linked PR, and now I see why we want to fetch the space information (to populate the quicklook component). But, what if the first ten spaces don't include running apps? Or if the first ten spaces all have apps that are running without errors, but the 11th does have an app with an error? 

As another example, here is a view of my dashboard:

screen shot 2017-07-03 at 12 03 27 pm

You can see that my `sandbox-gsa` org has two running apps, but the quicklook doesn't give any indication as to the status of these apps.

I think we could probably re-examine this feature and figure out an efficient way to get statuses for each org. It might be possible to do this without making a ton of requests and generating false negatives for orgs.

msecret:
@el-mapache @jcscottiii true, the way it's working now isn't ideal. One way would be to add UI that highlights whether all the information is complete. Another way would be to not show things that don't have complete information with a show more type of UI.

Another thing to consider: is it usually OK to just download all the information? How many spaces do our customers usually have (that can be found out in admin UI), how much data is it and how long does the request take? can information be presented incrementally as requests are being made?

el-mapache:
I think that if a typical user is only going to have a few spaces, it totally fine to fetch them all. But for large organizations, where a user might be member of multiple cloud.gov orgs, each with multiple spaces, its going to be pretty slow. It currently takes about 10-11 seconds to fetch 10 spaces locally and load the dashboard view. When I removed the space-fetching code, it took about 7-8 seconds.

el-mapache:
Does the api currently expose any way for use to just get highlights? Like get the number of running apps for a space, the number of failed/stopped apps, etc?

msecret:
@el-mapache it's entirely possible that it does. It wouldn't be the first time that we've missed something in the API that we hadn't known or didn't see before. It might be good to just go through the docs, or even ask in one of the cloud.gov community channels.

jcscottiii:
agreed. i would check out https://apidocs.cloudfoundry.org/263/. the api has changed some since the dashboard has been written.

el-mapache:
Doesn't look like they offer filtering by very many parameters, just name, space_guid, org_guid and one or two more.

el-mapache:
closing at this breaks the breadcrumbs feature.

PR #1150, opened 2017-07-05
Add prop-types package and update code

el-mapache:
* React.PropTypes is deprecated and was throwing a warning in the
console.

Just a lil day off coding..

There are a lot of changes but they can be boiled down to removing the `React` prefix from `PropTypes` and importing the `prop-types` library. I used [React codemod](https://github.com/reactjs/react-codemod) so I feel pretty confident about not missing anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "karma-jasmine-sinon": "^1.0.4",
     "lighthouse": "^1.6.3",
     "moxios": "^0.4.0",
+    "prop-types": "^15.5.10",
     "react-addons-test-utils": "^15.4.2",
     "react-dom": "^15.4.2",
     "selenium-standalone": "^5.11.0",
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "cloudgov-dashboard",
-  "version": "1.16.2",
+  "version": "1.16.7",
   "dependencies": {
     "abbrev": {
       "version": "1.1.0",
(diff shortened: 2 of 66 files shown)


package.json
```diff
@@ -114,6 +114,7 @@
     "karma-jasmine-sinon": "^1.0.4",
     "lighthouse": "^1.6.3",
     "moxios": "^0.4.0",
+    "prop-types": "^15.5.10",
```
jcscottiii: would this be a general dependency and not just a dev dependency?

el-mapache: I don't think so, prop types are only checked locally, they don't have any bearing on production code.

el-mapache:
Sorry, I'm not sure what you mean :( The package.json file is reporting the `prop-types` version is 15.5, do you mean I should update react to a newer version?

msecret:
Oh I see what you mean now. That' actually kinda makes sense, because our package.json says use any version of react below v15. So we probably updated our deps and it updated shrinkwrap but doesn't have to update package.json.

el-mapache:
@jcscottiii I think the '^' means that npm will try to update any minor or teeny versions of the package when it can, until the next major release. So, like @msecret said, when I regenerated the shrinkwrap file, npm saw there was a newer version of React (15.5.4) and picked that one over 15.4.2. Since it's a minor version there shouldn't be any breaking changes, and we should be safe to upgrade to that version of React when regenerating dependencies from the shrinkwrap file.

But, if we really want to have consistency, maybe we should peg the React version `package.json` and manually change it whenever we need to update React.

jcscottiii:
That's true.

We can leave it as-is. Let's revisit this part when @msecret revisits the shrinkwrap part and if even matters since the shrinkwrap already pegs it.

PR #1151, opened 2017-07-05
Remove initial props pattern

msecret:
As discussed in [contributing guide](https://github.com/18F/cg-dashboard/blob/master/CONTRIBUTING.md#having-components-with-both-props-and-state), the older pattern of having initial props in components that then get copied into state is redundant. The only place components should have both state and props is when props are for configuration of the component.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "cloudgov-dashboard",
-  "version": "1.16.2",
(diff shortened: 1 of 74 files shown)


msecret:
@el-mapache I kinda see `activity_log` as a container component right now, as nothing else really needs log and event data, so it contains all of it. I would agree that the filtering should happen in the store, as it's not component specific.

PR #1152, opened 2017-07-05
Update contrib document with new and updated information

msecret:


rememberlenny:
Thanks for updating this

PR #1155, opened 2017-07-07
Move render function out of route handlers

el-mapache:
This is a little POC to move `ReactDOM.render` calls out of the route handler functions and let React handle its own updates with a new `RouteProvider` component. The main benefit is that we don't have to re-render the `MainContainer` every time a route changes.

It also could set us up for integrating the router within the flux architecture if we want to do more of that in the future!

Let me know what ya'll think!

### DIFF
--- a/static_src/routes.js
+++ b/static_src/routes.js
@@ -1,6 +1,3 @@
-import React from 'react';
-import ReactDOM from 'react-dom';
-
 import activityActions from './actions/activity_actions.js';
 import AppContainer from './components/app_container.jsx';
 import appActions from './actions/app_actions.js';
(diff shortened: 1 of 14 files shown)


static_src/routes.js
```diff
@@ -57,10 +51,6 @@ export function overview(next) {
     })
     .then(pageActions.loadSuccess, pageActions.loadError);
 
-  ReactDOM.render(<MainContainer>
```
jcscottiii: is there a reason, you removed it from here and moved it up?

jcscottiii:
I like it. just had one question.

static_src/routes.js
el-mapache: No, I had originally moved it inside the callback, which obviously blocked rendering, and when I moved it out stuck it above the `spaceActions` fetch call for no good reason 😄

jcscottiii:
One more question. do you think this pattern should go into the `CONTRIBUTING` guide? (do you think it's something other devs would be likely to not do this if they didn't read it?)

el-mapache:
I'm not sure, I suppose it depends on the future routes added and what the devs' rationale is at that point in time..and also if we all agree splitting the apps concerns this way is worth the changes.

el-mapache:
Seems like this change is at least not _disliked_ by the team, so if someone wants to merge it that is :+1: !

PR #1158, opened 2017-07-12
Break log items up into discrete components

el-mapache:
* Logic is somewhat confusing and activity_log_item knows too much about
content internal to its children


jcscottiii:
just some FYI: as you see, there's two types of activities: events and logs. currently the logs don't work (i.e., if you look at the activities, you'll only see events right now). that's because the internal API for logs has changed (it's a different api than the CF API) But the events come from the CF API.

Either, I'm going to rip out the logs logic (because it keeps changing) or i'll update the code.

that's all just FYI info

el-mapache:
That makes sense, there is an error in the console for the logs endpoint. 

If we change the name of the component to reflect the fact that they are events, should we also change the name of the parent component, `ActivityLogItem`?

jcscottiii:
The parent component holds both the logs (when they work) and events right? I would change it to `AppActivity`

el-mapache:
Oh yeah, I like that name

PR #1159, opened 2017-07-14
Add indicator badge when deleting service instance

el-mapache:
Previously, we didn't give the user any feedback after they clicked the
'confirm delete' button.

Handy gif reference:
delete-service-instance

### DIFF
--- a/static_src/components/service_instance_table.jsx
+++ b/static_src/components/service_instance_table.jsx
@@ -8,10 +8,8 @@ import PanelDocumentation from './panel_documentation.jsx';
 import ServiceInstanceStore from '../stores/service_instance_store.js';
 import SpaceStore from '../stores/space_store.js';
 import { config } from 'skin';
-import createStyler from '../util/create_styler';
 import formatDateTime from '../util/format_date';
 import serviceActions from '../actions/service_actions.js';
-import style from 'cloudgov-style/css/cloudgov-style.css';
 
 function stateSetter() {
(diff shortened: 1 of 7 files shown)


static_src/components/service_instance_table.jsx
```diff
@@ -22,10 +20,13 @@ function stateSetter() {
     serviceInstances,
     currentSpaceGuid,
     loading: ServiceInstanceStore.loading,
-    empty: !ServiceInstanceStore.loading && !serviceInstances.length
+    empty: !ServiceInstanceStore.loading && !serviceInstances.length,
+    updating: ServiceInstanceStore.updating
   };
 }
 
+const propTypes = {};
```
jcscottiii: put this after the imports. (at the top)

jcscottiii:
we need a way to handle the error case of deleting the service. (should turn off _updating). (it's a lot of stuff to add so if you want to make a separate PR, you can do that.) let me know your thoughts.

el-mapache:
@jcscottiii I can add a test to the actions that verifies that the same method is from both `then` and `catch` outcomes of the promise. That will be pretty simple.

We don't really have any kind of error handling around this action, so I would definitely want to have any of that work in a separate ticket/PR

jcscottiii:
@el-mapache that would be really good!

el-mapache:
@jcscottiii Test file updated!

jcscottiii:
@el-mapache could you make a card for taking care of the error case?

el-mapache:
@jcscottiii Definitely!

PR #1162, opened 2017-07-19
Invite users to spaces

el-mapache:
Org managers can now associate users with spaces.

Screenshot:

screen shot 2017-07-20 at 3 53 40 pm

In the case of a space manager who is not an org manager, they are directed to use the cloud foundry api (in the case of associating an existing user) or to ask their org manager to do it for them.

Screenshot:
screen shot 2017-07-20 at 2 55 11 pm

### DIFF
--- a/static_src/actions/user_actions.js
+++ b/static_src/actions/user_actions.js
@@ -125,7 +125,7 @@ const userActions = {
 
   addUserRoles(roles, apiKey, userGuid, entityGuid, entityType) {
     const apiMethodMap = {
-      org: cfApi.putOrgUserPermissions,
+      organization: cfApi.putOrgUserPermissions,
       space: cfApi.putSpaceUserPermissions
     };
     const api = apiMethodMap[entityType];
@@ -163,7 +163,7 @@ const userActions = {
 
   deleteUserRoles(roles, apiKey, userGuid, entityGuid, entityType) {
     const apiMethodMap = {
-      org: cfApi.deleteOrgUserPermissions,
+      organization: cfApi.deleteOrgUserPermissions,
       space: cfApi.deleteSpaceUserPermissions
     };
     const api = apiMethodMap[entityType];
@@ -296,52 +296,55 @@ const userActions = {
   },
 
   createUserAndAssociate(userGuid) {
-    let entityGuid;
     const entityType = UserStore.currentlyViewedType;
+    const orgGuid = OrgStore.currentOrgGuid;
+    let cfApiRequest;
+    let entityGuid;
 
     if (entityType === ORG_NAME) {
-      entityGuid = OrgStore.currentOrgGuid;
+      entityGuid = orgGuid;
+      cfApiRequest = cfApi.putAssociateUserToOrganization.bind(cfApi, userGuid, entityGuid);
+
+      AppDispatcher.handleViewAction({
+        type: userActionTypes.USER_ORG_ASSOCIATE,
+        userGuid,
+        entityType,
+        entityGuid
+      });
     } else {
       entityGuid = SpaceStore.currentSpaceGuid;
+      cfApiRequest = cfApi.putAssociateUserToSpace.bind(cfApi, userGuid, orgGuid, entityGuid);
     }
 
-    AppDispatcher.handleViewAction({
-      type: userActionTypes.USER_ORG_ASSOCIATE,
-      userGuid,
-      entityType,
-      entityGuid
-    });
-
-    return cfApi.putAssociateUserToEntity(userGuid, entityGuid, entityType)
+    return cfApiRequest()
       .then(() => userActions.fetchEntityUsers(entityGuid, entityType))
-      .then(entityUsers => userActions.createdUserAndAssociated(userGuid, entityGuid, entityUsers));
+      .then(entityUsers => {
+        userActions.createdUserAndAssociated(userGuid, entityGuid, entityUsers, entityType);
+      });
   },
 
   fetchEntityUsers(entityGuid, entityType) {
-    let entityUsers;
-    if (entityType === ORG_NAME) {
-      entityUsers = cfApi.fetchOrgUsers(entityGuid);
-    } else {
-      entityUsers = cfApi.fetchSpaceUserRoles(entityGuid);
-    }
-    return Promise.resolve(entityUsers);
+    return cfApi[(entityType === ORG_NAME) ? 'fetchOrgUsers' : 'fetchSpaceUserRoles'](entityGuid);
   },
 
-  createdUserAndAssociated(userGuid, entityGuid, entityUsers) {
-    const user = entityUsers.filter((entityUser) => entityUser.guid === userGuid);
+  createdUserAndAssociated(userGuid, entityGuid, entityUsers, entityType) {
+    const user = entityUsers.filter(entityUser => entityUser.guid === userGuid)[0];
 
-    if (!user[0]) {
+    if (!user) {
       const err = new Error('User was not associated');
       const message = `The user ${userGuid} was not associated in ${entityGuid}.`;
       return Promise.resolve(userActions.userInviteCreateError(err, message));
     }
 
-    AppDispatcher.handleViewAction({
-      type: userActionTypes.USER_ORG_ASSOCIATED,
-      userGuid,
-      entityGuid,
-      user
-    });
+    if (entityType === ORG_NAME) {
+      AppDispatcher.handleViewAction({
+        type: userActionTypes.USER_ORG_ASSOCIATED,
+        userGuid,
+        entityGuid,
+        user
+      });
+    }
+
     return Promise.resolve(user);
   },
 
@@ -433,12 +436,14 @@ const userActions = {
     // TODO add error action
     return userActions
       .fetchCurrentUserInfo()
-      .then(userInfo =>
-        Promise.all([
-          userActions.fetchUser(userInfo.user_id),
-          userActions.fetchCurrentUserUaaInfo(userInfo.user_id)
-        ])
-      )
+      .then(userInfo => {
+        const userId = userInfo.user_id;
+
+        return Promise.all([
+          userActions.fetchUser(userId),
+          userActions.fetchCurrentUserUaaInfo(userId)
+        ]);
+      })
       // Grab user from store with all merged properties
       .then(() => UserStore.currentUser)
       .then(userActions.receivedCurrentUser);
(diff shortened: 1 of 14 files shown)


static_src/components/users_invite.jsx
```diff
@@ -65,55 +58,60 @@ export default class UsersInvite extends React.Component {
...
+
     if (err.message) {
       return `${message}: ${err.message}.`;
     }
+
     return message;
   }
 
+  get entityType() {
+    return /space/.test(this.props.inviteEntityType) ? 'space' : 'organization';
```
el-mapache: I think, in a future PR, we will want to have some kind of service class or utility code that takes the raw name of the entity and gives us the human-friendly version.

el-mapache:
@jcscottiii So given that, will we need to figure out what that UI looks like before this PR will be valuable?

jcscottiii:
@el-mapache in the meantime, you can do a check and show error.
Given I'm a space manager, if I try to invite a user to my space who currently does not exist in the org, show an error asking for an org manager to first invite them to the org. 

Given I'm a space manager, if i try to invite a user to my space who current does exist in my org, the user is added to my space.

el-mapache:
@jcscottiii Just so I am totally clear: we want to prevent the invite from ever hitting the UAA API, correct?

rememberlenny:
Jumping on this

static_src/components/users.jsx
```diff
@@ -116,75 +122,85 @@ export default class Users extends React.Component {
     return entityGuid;
   }
 
-  _onChange() {
-    this.setState(stateSetter());
+  get currentUserIsOrgManager() {
+    const { currentUser } = this.state;
+    const { currentOrgGuid } = OrgStore;
+
+    // TODO this is not 100% robust.
```
jcscottiii: why isn't it robust?

static_src/actions/user_actions.js
```diff
@@ -257,25 +253,27 @@ const userActions = {
...
 
   fetchEntityUsers(entityGuid, entityType) {
-    let entityUsers;
-    if (entityType === ORG_NAME) {
-      entityUsers = cfApi.fetchOrgUsers(entityGuid);
-    } else {
-      entityUsers = cfApi.fetchSpaceUserRoles(entityGuid);
-    }
-    return Promise.resolve(entityUsers);
+    return cfApi[(entityType === ORG_NAME) ? 'fetchOrgUsers' : 'fetchSpaceUserRoles'](entityGuid);
```
jcscottiii: while this makes the code more concise, i wonder if this takes away from the clarify and future expansion of the code (stuff to do in the branches)

static_src/components/user_list.jsx
```diff
@@ -134,8 +134,6 @@ export default class UserList extends React.Component {
 
     if (this.props.empty) {
       content = this.emptyState;
-    } else if (this.props.users.length === 1) {
```
jcscottiii: why did this get removed?

static_src/components/users.jsx
```diff
@@ -116,75 +122,85 @@ export default class Users extends React.Component {
...
+      <Notification
+        message={ inviteNotices.description }
+        actions={ [] }
+        onDismiss={ this.onNotificationDismiss }
+        status="finish"
+      />
+    );
+  }
+
+  get userInvite() {
```
jcscottiii: (at first glance, i could be wrong), will this come up on the org page for a non admin, non org manager user? (this text is catered to the space page)

static_src/components/users_invite.jsx
```diff
@@ -65,55 +58,68 @@ export default class UsersInvite extends React.Component {
...
+
+    if (error.message) {
+      return `${message}: ${error.message}.`;
     }
+
     return message;
   }
 
+  get entityType() {
+    return this.isOrganization ? 'organization' : 'space';
```
jcscottiii: const values for organization and space

static_src/components/users_invite.jsx
```diff
@@ -65,55 +58,68 @@ export default class UsersInvite extends React.Component {
...
+  }
+
+  get isOrganization() {
+    return /org/.test(this.props.inviteEntityType);
```
jcscottiii: is there a comparison we can do without a regex?

static_src/actions/user_actions.js
el-mapache: If we were doing more things in the branch, I would advocate to move the code into separate methods to maintain readability. I think this is ok as is, for now.

static_src/components/users_invite.jsx
jcscottiii: i only ask because, we should know for sure what values `inviteEntityType` could be

static_src/components/user_list.jsx
el-mapache: I didn't think the `onlyOneState` makes sense anymore. If the user is an orga manager, we will want to show the invite component, and if they are a space manager, we want to show the message telling them to ask their org manager to invite users. 

Maybe we need to show it regardless.

static_src/components/users_invite.jsx
el-mapache: We could compare against the exact string that `inviteEntityType` is. I didn't use a const because we don't currently (I dont think?) have a consistent way that we are handling those in the app.

static_src/components/users.jsx
el-mapache: left this in by mistake!

static_src/components/users.jsx
el-mapache: It will, I need to add the `entityType` interpolation we will in the `UserInvite` component. I might just move this to a separate component so there isn't a ton of logic in the method.

static_src/components/users_invite.jsx
el-mapache: I'll just add the constant for now, I'll make a ticket to go through an audit our use of strings/constants, and where we put the constants e.g. all in the `constants.js` file, or in additional domain specific files, etc.

static_src/components/user_list.jsx
jcscottiii: leave it in for now. we can revisit this when we develop the UI for the space managers to move exiting org users into the space.

if it's a hassle to maintain that separate `onlyOneState` state, let's talk about it then though.

static_src/components/users_invite.jsx
jcscottiii: >  I'll make a ticket to go through an audit our use of strings/constants, and where we put the constants e.g. all in the constants.js file

That's a great idea!

el-mapache:
@jcscottiii Updated with feedback + rebased with master

static_src/actions/user_actions.js
```diff
@@ -296,52 +296,55 @@ const userActions = {
...
   },
 
-  createdUserAndAssociated(userGuid, entityGuid, entityUsers) {
-    const user = entityUsers.filter((entityUser) => entityUser.guid === userGuid);
+  createdUserAndAssociated(userGuid, entityGuid, entityUsers, entityType) {
+    const user = entityUsers.filter(entityUser => entityUser.guid === userGuid);
 
-    if (!user[0]) {
+    if (!user.length) {
```
jcscottiii: hmm, i rather keep this checking for the zero index. you can have an empty array which has a length of zero

el-mapache: Hmm I'm not sure I understand. If the array has a length of zero it is going to evaluate as false in the if statement.

jcscottiii:
@el-mapache whoops you were right about that one comment. i meant to say if you think `users[0]` makes it more clear that you need that element.

el-mapache:
@jcscottiii Ahh ok. I think maybe calling it users is misleading, we are filtering all the users down to just the current user. What do you think about grabbing the first element immediately and just naming the  variable `user`? Then we can just write:
```
if (!user)
```
which is probably clearer.

static_src/test/unit/components/users.spec.jsx
```diff
@@ -1,52 +1,83 @@
...
+  describe('with a user', () => {
+    beforeEach(() => {
+      UserStore._data = Immutable.fromJS([user]);
       users = shallow(<Users />);
     });
 
-    describe('when at org level', function () {
-      beforeEach(function () {
+    describe('when at org level', () => {
+      it('doesnt have permissions to edit users', () => {
```
jcscottiii: i know this was here before, but let's change this statement (or change the test, but probably the former). it doesn't test the permissions of the users.

static_src/test/unit/components/users.spec.jsx
```diff
@@ -1,52 +1,83 @@
...
         users.setState({ currentType: 'org_users' });
+        const actual = users.instance().entityType;
+
+        expect(actual).toEqual('organization');
       });
+    });
 
-      it('doesnt have permissions to edit users', function () {
+    describe('when at space level', () => {
+      it('doesnt have permissions to edit users', () => {
```
jcscottiii: same as https://github.com/18F/cg-dashboard/pull/1162/files#r129148710

static_src/test/unit/stores/user_store.spec.js
```diff
@@ -134,22 +135,25 @@ describe('UserStore', function () {
       expectedUsers = [
         {
           guid: userGuidA,
-          roles: { [spaceGuid]: ['space_developer'] }
+          roles: { [spaceGuid]: ['space_developer'] },
+          space_roles: [ 'space_developer' ]
```
jcscottiii: is there a reason we need this new attribute `space_roles` when we already have a mapping between the space guid and the roles that person has for the guid?

jcscottiii:
sorry i didn't go through everything last time. just finished now. left one main question.

jcscottiii:
@el-mapache re: https://github.com/18F/cg-dashboard/pull/1162#issuecomment-317547523 i like that much better

static_src/test/unit/stores/user_store.spec.js
el-mapache: Its not a new attribute, its an old one. `space_roles` is returned when we fetch the user from the API. We do a transformation that adds the `roles` key, but in order to pass a deep equality test, the object needs to retain the `space_roles` key. 

We could edit the code to the delete `space_roles` from the underlying object, but I don't think it hurts to have the extra field.

jcscottiii: ahh didn't know it was already there. in that case, we can leave it. i think we need to revisit the whole topic of the post-processing of data (like this transformation). but another day!

el-mapache:
@jcscottiii alright, updated based on current round of feedback!

jcscottiii:
:shipit: once the tests pass

jcscottiii:
@el-mapache thanks for squashing the commits!

rememberlenny:
Thanks for this! It looks great!

PR #1164, opened 2017-07-24
Add a "Remove all space roles" button

rememberlenny:
remove

### DIFF
--- a/static_src/constants.js
+++ b/static_src/constants.js
@@ -270,6 +270,10 @@ const userActionTypes = keymirror({
   USER_INVITE_ERROR: null,
   // Action to dismiss an user list notification.
   USER_LIST_NOTICE_DISMISSED: null,
+  // Action to remove all roles for space user.
+  USER_REMOVE_ALL_SPACE_ROLES: null,
+  // Action to removed all roles for space user.
+  USER_REMOVED_ALL_SPACE_ROLES: null,
   // Action to delete a user from an org.
   USER_DELETE: null,
   // Action when a user was deleted from an org on the server.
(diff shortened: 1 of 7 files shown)


static_src/constants.js
```diff
@@ -270,6 +270,10 @@ const userActionTypes = keymirror({
   USER_INVITE_ERROR: null,
   // Action to dismiss an user list notification.
   USER_LIST_NOTICE_DISMISSED: null,
+  // Action to remove all roles for space user.
+  USER_REMOVE_ALL_SPACE_ROLES: null,
+  // Action to removed all roles for space user.
```
jcscottiii: `Action when all roles removed from space user`

static_src/stores/user_store.js
```diff
@@ -164,6 +164,13 @@ export class UserStore extends BaseStore {
         break;
       }
 
+      case userActionTypes.USER_REMOVED_ALL_SPACE_ROLES: {
```
jcscottiii: only need one more test. that it changes the list of users upon removing the person.

jcscottiii: there's similar tests for the store for the add or delete cases.

static_src/constants.js
```diff
@@ -270,6 +270,10 @@ const userActionTypes = keymirror({
   USER_INVITE_ERROR: null,
   // Action to dismiss an user list notification.
   USER_LIST_NOTICE_DISMISSED: null,
+  // Action when all roles for space user are removed.
```
jcscottiii: oops i think i told you to put that comment on the wrong constant. i meant the one below should be the `when all`...

jcscottiii: this first one should be `Action to...`

rememberlenny: Fixed!

PR #1169, opened 2017-07-26
Restore entity specific button text

el-mapache:
Fixes button text! My dev env is down right now, so I can't supply a screenshot, unfortunately.


jcscottiii:
Let's add a test to catch this.

PR #1172, opened 2017-07-28
Create space user's dropdown form for adding other org users

rememberlenny:
record


static_src/actions/user_actions.js
```diff
@@ -12,7 +12,10 @@ import UserStore from '../stores/user_store';
 import OrgStore from '../stores/org_store';
 import SpaceStore from '../stores/space_store';
 
+const ORG_ENTITY = 'organization';
+const SPACE_ENTITY = 'space';
 const ORG_NAME = OrgStore.cfName;
+// const SPACE_NAME = SpaceStore.cfName;
```
jcscottiii: why was this commented out?

static_src/components/users_parent_entity_user_selector.jsx
```diff
@@ -0,0 +1,147 @@
...
+    parentEntityUsers: props.parentEntityUsers,
+    parentEntityGuid: props.parentEntityGuid,
+    parentEntity: props.parentEntity,
+    currentEntityGuid: props.currentEntityGuid,
+    currentEntity: props.currentEntity,
+    error: props.error
+  };
+}
+
+export default class UsersParentEntityUserSelector extends React.Component {
```
jcscottiii: we might want to change the name of this. at first glance it's not that clear.

static_src/actions/user_actions.js
rememberlenny: It has a linking error because it's not used.

static_src/components/users_parent_entity_user_selector.jsx
rememberlenny: I totally agree. Originally it was supposed to be a module. Not it only does one thing, check org users, if user is in a space.

static_src/components/users.jsx
```diff
@@ -186,6 +199,34 @@ export default class Users extends React.Component {
...
 
+  get userParentEntityUserSelector() {
+    if (!this.isSpace) {
+      return null;
+    }
+
+    if (!this.currentUserIsSpaceManager) {
+      return (
+        <PanelDocumentation>
+          ORG AND SPACE MANAGERS CAN DO THIS.
```
jcscottiii: Not sure you meant to have this text here.

static_src/components/users_parent_entity_user_selector.jsx
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  _onSubmitForm(errs, values) {
+    const entityType = this.state.currentEntity;
+    const entityGuid = this.state.currentEntityGuid;
+    const apiKey = 'auditors';
+    const roles = 'space_auditor';
+    if (values.userGuid) {
+      const userGuid = values.userGuid.value;
+      userActions.addUserRoles(roles, apiKey, userGuid, entityGuid, entityType);
```
jcscottiii: i think in order to prevent confusion, we should only show org users currently without space roles. then here, upon successful submission, we should remove the user from the list so we don't see them here.

static_src/components/users.jsx
rememberlenny: Hahaha. No it should be :)

static_src/components/users_parent_entity_user_selector.jsx
rememberlenny: I was working on last night, but didn't know quite how the EventEmitter on the UserStore worked. I need to dig in a bit.

Basically, the logic to do this works.

static_src/components/users_parent_entity_user_selector.jsx
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+  get errorMessage() {
```
jcscottiii: test for this

static_src/components/users_parent_entity_user_selector.jsx
```diff
@@ -0,0 +1,147 @@
...
+    const message = error.contextualMessage;
+
+    if (error.message) {
+      return `${message}: ${error.message}.`;
+    }
+
+    return message;
+  }
+
+  get invitationMessage() {
```
jcscottiii: test for this

static_src/components/users_parent_entity_user_selector.jsx
```diff
@@ -0,0 +1,147 @@
...
+    const parentEntity = this.props.parentEntity;
+    const currentEntity = this.props.currentEntity;
+
+    return `Invite an existing user in this ${parentEntity}` +
+      ` to this ${currentEntity}.`;
+  }
+
+  get userSelector() {
```
jcscottiii: test for this

static_src/actions/user_actions.js
el-mapache: lets delete it!

rememberlenny: 👍

static_src/components/users_parent_entity_user_selector.jsx
rememberlenny: Copy!

static_src/test/unit/stores/user_store.spec.js
```diff
@@ -682,6 +682,19 @@ describe('UserStore', function () {
     });
   });
 
+  describe('getAllInOrgAndNotSpace()', function() {
+    it('should find all users that are in an org, without the space users', function() {
```
el-mapache: I am probably misreading this spec, but from the description, shouldn't the `testUser` be excluded here? They don't have a `space_roles` key on their object, they do have a space role in their `roles`.

Should the corresponding method in `UserStore` also check for the existence of space roles within the `roles` object?

static_src/components/users_parent_entity_user_selector.jsx
```diff
@@ -0,0 +1,153 @@
...
+};
+const defaultProps = {
+  orgUsersSelectorDisabled: false,
+  currentUserAccess: false,
+  error: {}
+};
+
+
+function stateSetter(props) {
+  return {
```
msecret: Why is state needed in this component if everything is passed down from props? https://github.com/18F/cg-dashboard/blob/master/CONTRIBUTING.md#components

rememberlenny: Its not. This is a great document. Thanks.

static_src/components/users_org_user_selector.jsx
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  componentDidMount() {
+    FormStore.create(USERS_PARENT_ENTITY_USER_FORM_GUID);
+  }
+
+  _onSubmitForm(errs, values) {
+    const { currentEntity } = this.props;
+    const { currentEntityGuid } = this.props;
+    const apiKey = AUDITOR_NAME;
```
el-mapache: Should this be renamed to `endpoint`? When I read api key, I think of an access key, or a secret. I know you didn't introduce this pattern, but maybe we can start renaming now!

jcscottiii: i like the idea of renaming it! let's move it to `resource` (i think of endpoint as being the full path). 

- the person who introduced the pattern lol

jcscottiii: there's another refactor while we are talking about it, (but outside the scope of this PR)
addUserRoles currently only takes one role but it and the argument are plural `roles`

static_src/components/users.jsx
```diff
@@ -186,6 +199,33 @@ export default class Users extends React.Component {
...
+    if (!this.currentUserIsSpaceManager) {
+      return (
+        <PanelDocumentation>
+          Org Managers and Space Managers can add current organization users into this space.
+        </PanelDocumentation>
+      );
+    }
+
+    return (
+      <OrgUsersSelector
```
jcscottiii: this component has a lot of attributes. for a lot of attributes, you can make this generic like `UserSelector`.
For OrgUserSelector, you probably wouldn't need all this. I'm leaning to keeping all of these and just rename to `UserSelector` and we can refactor for more generic cases later on.

rememberlenny: On it!

static_src/actions/user_actions.js
```diff
@@ -146,7 +148,7 @@ const userActions = {
     });
   },
 
-  addUserRoles(roles, apiKey, userGuid, entityGuid, entityType) {
+  addUserRoles(resource_role_name, user_role, userGuid, entityGuid, entityType) {
```
jcscottiii: ahh sorry for the confusing conversation.
we were saying `apiKey` -> `resource`
and you can revert `resource_role_name` to `roles` and leave `roles` as-is for now.
we will address the naming then.

el-mapache: We should stick to camel case as well

static_src/components/users.jsx
```diff
@@ -186,6 +199,33 @@ export default class Users extends React.Component {
...
+      return (
+        <PanelDocumentation>
+          Org Managers and Space Managers can add current organization users into this space.
+        </PanelDocumentation>
+      );
+    }
+
+    return (
+      <UsersSelector
+        orgUsersSelectorDisabled={ this.state.orgUsersSelectorDisabled }
```
jcscottiii: also, this can be renamed to `usersSelectorDisabled`

static_src/components/users_selector.jsx
```diff
@@ -0,0 +1,123 @@
...
+import PanelDocumentation from './panel_documentation.jsx';
+import userActions from '../actions/user_actions';
+import { validateString } from '../util/validators';
+
+const AUDITOR_NAME = 'auditors';
+const SPACE_AUDITOR_NAME = 'space_auditor';
+const USERS_PARENT_ENTITY_USER_FORM_GUID = 'users-parent-entity-users-form';
+
+const propTypes = {
+  orgUsersSelectorDisabled: PropTypes.bool,
```
jcscottiii: this can be renamed to `selectorDisabled` or `usersSelectorDisabled`

fureigh:
@rememberlenny: Is the class in line 78 of https://github.com/18F/cg-dashboard/blob/339846c962783db8afa63f054be202af00b1a0e0/static_src/components/users_selector.jsx meant to be "test-users"? If so, LGTM.

rememberlenny:
@fureigh It was a copy over. Ill change for specificity.

rememberlenny:
@fureigh Its a react syntax. If you take a look at the other components, you will see that they all have a `test-` prefix class. This allows us to target the elements in a test.

Issue #790, opened 2016-12-03
Dashboard should be generic


jmcarp:
In principle, nothing about this app should be specific to cloud.gov--so this should be a cloud foundry dashboard, not a cloud.gov dashboard. Toward that goal, we should remove any hard-coded references to cloud.gov here, either linking to cloud foundry materials or allowing operators to override text, logos, etc. I'm glad to see that @tessereth has started this work in https://github.com/AusDTO/cg-dashboard/pull/2.

* [ ] Link to cloud foundry docs instead of docs.cloud.gov
* [ ] Allow operators to override templates
* [ ] Rename project to cf-dashboard

msecret:
Will discuss this with Bret based on Pivotal discussion of making dashboard more transparent with CF.

jonathaningram:
I don't mean to bring up old issues, but I was looking through what issues are still around and noticed this. As you know, lots of the work cloud.gov.au is doing is about making the project more generic and so pretty soon most of this ticket will be done anyway.

I just wanted to chime in and briefly comment on the task "Rename project to cf-dashboard".

At this stage cloud.gov.au is calling this "console", not "dashboard". It matches console.cloud.google.com and console.aws.amazon.com. For me, this is a console not a dashboard. For me a dashboard is a page showing charts and stats or the home page of most web applications. I don't want to bikeshed (really, dashboard is fine) but if the project does undergo a rename, perhaps consider `cf-console`.

PR #1173, opened 2017-07-28
Display error when service instance creation fails

el-mapache:
Screenshots!

screen shot 2017-07-28 at 4 46 12 pm

screen shot 2017-07-28 at 4 45 43 pm

### DIFF
--- a/static_src/actions/service_actions.js
+++ b/static_src/actions/service_actions.js
@@ -134,7 +134,8 @@ const serviceActions = {
 
     return cfApi.createServiceInstance(name, spaceGuid, servicePlanGuid)
       .then(serviceInstance => serviceActions.fetchInstance(serviceInstance.guid))
-      .then(serviceActions.createdInstance, serviceActions.errorCreateInstance);
+      .then(serviceActions.createdInstance)
+      .catch(serviceActions.errorCreateInstance);
   },
 
   createdInstance(serviceInstance) {
@@ -146,10 +147,13 @@ const serviceActions = {
     return Promise.resolve(serviceInstance);
   },
 
-  errorCreateInstance(err) {
+  errorCreateInstance(error) {
+    const { response } = error;
+    const safeError = (response && response.data) || { code: 500 };
+
     AppDispatcher.handleServerAction({
       type: serviceActionTypes.SERVICE_INSTANCE_CREATE_ERROR,
-      error: err
+      error: safeError
     });
 
     return Promise.resolve();
--- a/static_src/test/unit/actions/service_actions.spec.js
+++ b/static_src/test/unit/actions/service_actions.spec.js
@@ -357,8 +357,7 @@ describe('serviceActions', function() {
   });
 
   describe('createdInstance()', function() {
-    it('should dispatch a server event of type instance created with service',
-        function() {
+    it('dispatchs a server event of type instance created with service', () => {
       var expectedInstance = { guid: 'asdfas' };
 
       let expectedParams = {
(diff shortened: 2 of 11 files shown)


el-mapache:
Due to the way the form component works, server errors are displayed before the 'There were errors submitting the form'.

static_src/test/unit/actions/service_actions.spec.js
```diff
@@ -373,19 +372,47 @@ describe('serviceActions', function() {
...
-      assertAction(spy, serviceActionTypes.SERVICE_INSTANCE_CREATE_ERROR,
-                   expectedParams);
+    describe('server fault', () => {
+      it('dispatches the correct error type and a code 500', () => {
+        const originalError = { message: 'Bad error', stack: [] };
+        const actual = { code: 500 };
+
+        serviceActions.errorCreateInstance(originalError);
+
+        const actionInfo = spy.getCall(0).args[0];
```
el-mapache: I'm not suing the test helpers here as I was having a hard time getting the test to validate when the arguments passed to the dispatch function are objects.

static_src/test/unit/stores/service_instance_store.spec.js
```diff
@@ -286,16 +286,35 @@ describe('ServiceInstanceStore', function() {
...
-  describe('on instance create error', function() {
-    it('should set create error to passed in error', function() {
-      var expectedError = { status: 500 };
+  describe('on SERVICE_INSTANCE_CREATE_ERROR', () => {
+    it('should set `createError` based on error received', () => {
+      const serverError = { code: 500 };
+      const argumentError = { response: { data: { error_code: 'CF-MessageParseError' } } };
+      const spaceError = { response: { data: { error_code: 'CF-ServiceInstanceInvalid' } } };
+      const serverErrorMsg =
+        'Error #500: please contact cloud.gov support for help troubleshooting this issue.';
```
jcscottiii: let me know if you think this would be something for another PR:  would be nice to remove the `cloud.gov` hard coding so that in the future we can achieve https://github.com/18F/cg-dashboard/issues/790 eventually

el-mapache: Yes I agree. Do we have any concept of the app environment baked into the build process or anything? I think we can have webpack set a global variable during the build process that gets injected into the app code. Then 'cloud.gov' could just be whatever platform the dashboard is running on.

static_src/actions/service_actions.js
```diff
@@ -146,10 +147,13 @@ const serviceActions = {
     return Promise.resolve(serviceInstance);
   },
 
-  errorCreateInstance(err) {
+  errorCreateInstance(error) {
+    const { response } = error;
+    const safeError = (response && response.data) || { code: 500 };
```
jcscottiii: coming from a non-JS perspective:

it looks like `(response && response.data)`  could evaluate to true or false.
if true, safeError would be set to true.

if false, it would set safeError to {code: 500};

Now your tests, tell me it doesn't work that way, but just could be a little confusing at first glance.

jcscottiii: also, don't you also need`error_code` to make sure this is a safe error?

el-mapache: Using this construct in JS, if the first argument evaluates to `true`, and the second argument is also true, the second argument will be executed. In this instance, executing means storing the value of the second argument in a variable, but you could, for example,  pass a function as the second argument.

If the second argument is also false, the fallback will be executed instead.

I concede that this might look a little confusing but it is a pretty common in JS.

If I misunderstood the question, and you already knew this, sorry 😬 !

el-mapache: @jcscottiii For your second comment, `error_code` needn't be present, the store checks for the existence of one and returns the generic 500 error message if it isn't present.

jcscottiii: Regarding the first one, I've seen it done like this with ternary operators. `condition ? happy case : sad case` but sounds good.

jcscottiii: Regarding the second one: ahh i gotcha.

el-mapache:
I mistakenly misnamed this PR's branch, it covers error handling when creating a service instance, not binding a service instance.

PR #1178, opened 2017-08-02
Fix user state bug

jcscottiii:
Pairing with @el-mapache
we found that if you navigate to the org or space page from home, you
are able to successfully change people's roles. However, if you refresh
the page, it will not work because the currentUser context is not set
upon refresh.

This patch removes this bit which regulated state updates


el-mapache (approved):
Looks good!

jcscottiii:
No test for this one, but made sure we tackle it in the acceptance testing https://github.com/18F/cg-dashboard/issues/1179

PR #1182, opened 2017-08-03
Fix invite flow depending on permissions

jcscottiii:
## Before this PR: 
- The help message for asking people to invite users still said that you can only invite existing org users via the CLI. This was no longer the case when the UserSelector was introduced. So it needed to be updated.
- An org manager could not see the UsersSelector. So they couldn't invite existing org users to the space. So we need to allow org managers to see the selector too. 
image
- In the case when the user was not an org manager nor a space manager inside a space, they saw two panels with similar info. It was confusing. So let's reduce that to one (yes similar screenshot to the point above that show the same thing).
image
- when a space manager was inside their space, they saw the old message saying they could not invite space users and instead needed to use the CLI. Let's get rid of that. 
image

### Help Message for a user has been split into three cases
1) When on the org page without org manager permissions
image

2) When on the space page as an space manager
image

3) When on the space space not as a space manager
(I didn't bother to go into the difference between existing org users vs new org users because that might be confusing to a non-space manager/ non-org-manager user.)
image

(The cases when the user is an org manager on either space or org page, there's no help text to ask someone for help or to do things via the CLI)

### Org Managers can invite existing org users to space

(This was because the space page didn't load the org roles)
image
 

The first commit is the tests adds coverage for the desired behavior mentioned above (CI will fail)
The second commit fixes the code to match the behavior.
(A third commit may be required to fix any functional tests + comments from reviewers)

### DIFF
--- a/static_src/components/users.jsx
+++ b/static_src/components/users.jsx
@@ -24,6 +24,14 @@ const ORG_MANAGER = 'org_manager';
 const SPACE_MANAGER = 'space_manager';
 const ORG_ENTITY = 'organization';
 const SPACE_ENTITY = 'space';
+const ORG_INVITE_HELP = 'Only an Org Manager can new invite users to this ' +
+  'organization via the dashboard. Speak to your Org Manager if you need to ' +
+  'add a user to this organization';
+const SPACE_INVITE_SPACE_MANAGER_HELP = 'As an Space Manager, you can invite existing ' +
+  'organization users into your space. If you wish to invite a person who is ' +
+  'not in the organization into your space, please ask an Org Manager';
+const SPACE_INVITE_HELP = 'If you wish to invite users into this space, please ' +
+  'ask an Org Manager or a Space Manager';
 
 function stateSetter() {
   const currentOrgGuid = OrgStore.currentOrgGuid;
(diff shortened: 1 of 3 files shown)


jcscottiii:
@brittag this addresses number one of your comments here https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-6974

rememberlenny:
This is really great.

static_src/components/users.jsx
```diff
@@ -176,19 +175,45 @@ export default class Users extends React.Component {
...
+      );
+    }
+    // When on the space page, likewise, an org manager should always see the
+    // invite form. Else, let's dig into what to display.
+    if (this.isSpace && !this.currentUserIsOrgManager) {
+      // if the user is a space manager, let them know that they can invite
+      // existing org users but not new ones.
+      if (this.currentUserIsSpaceManager) {
+        return (
+          <PanelDocumentation>
```
rememberlenny: In the future, I can imagine there only needing to be one <PanelDocumentation> return statement. All the other content can be compiled as a raw variable.

static_src/components/users.jsx
```diff
@@ -176,19 +175,45 @@ export default class Users extends React.Component {
...
+          </PanelDocumentation>
+        );
+      }
+      // Else, just tell the user to invite a space manager or org manager.
+      // Let's not confuse the regular user with the difference between
+      // new and existing org users.
+      // We can figure out wording later.
+      return (
+        <PanelDocumentation>
+          If you wish to invite users into this space, please ask an
```
brittag: We should be consistent about capitalizing Org Manager and Space Manager throughout the copy (see https://github.com/18F/cg-product/blob/master/ContentGuide.md for list of consistent capitalization/spacing/etc. to use for this kind of thing)

brittag:
Cool! We should be consistent about capitalizing Org Manager and Space Manager throughout the copy (see https://github.com/18F/cg-product/blob/master/ContentGuide.md for list of consistent capitalization/spacing/etc. to use for this kind of thing).

static_src/components/users.jsx
jcscottiii: i agree but we need to first clean up this decision tree of branches otherwise it could get confusing assigning / reassigning variables.

static_src/components/users.jsx
```diff
@@ -24,6 +24,14 @@ const ORG_MANAGER = 'org_manager';
 const SPACE_MANAGER = 'space_manager';
 const ORG_ENTITY = 'organization';
 const SPACE_ENTITY = 'space';
+const ORG_INVITE_HELP = 'Only an Org Manager can new invite users to this ' +
```
rememberlenny: Good move. We might _also_ want to consider moving all strings into a separate file that gets imported.

jcscottiii: agreed. there's a constants.js file too we should probably move things there eventually but should make we group them accordingly.

el-mapache: @rememberlenny I actually have a ticket somewhere in the backlog to start rounding up strings and splitting them into constant files! 😄

static_src/components/users.jsx
```diff
@@ -24,7 +24,14 @@ const ORG_MANAGER = 'org_manager';
 const SPACE_MANAGER = 'space_manager';
 const ORG_ENTITY = 'organization';
 const SPACE_ENTITY = 'space';
-const cfCliLink = 'https://docs.cloudfoundry.org/adminguide/cli-user-management.html#space-roles';
+const ORG_INVITE_HELP = 'Only an Org Manager can new invite users to this ' +
```
msecret: Is there a reason these long strings aren't being done with `dedent` and template strings? I can't remember it if there is and wanted to learn for other projects.

jcscottiii: there was this commit that scarred me after it took a while finding the problem lol https://github.com/18F/cg-dashboard/commit/77a78f381efa28b1ebc80090cf1270ac3271bc51

el-mapache: In some cases template strings introduce unnecessary spaces, and will fail tests if the strings aren't formatted identically. 

If the strings are all moved into constant files, the issue should go away?

PR #1183, opened 2017-08-04
Add Pagination for org users, org roles and org calls

jcscottiii:
The tests do not check if the appropriate action is still called but instead now I check for a function.
Eventually, we want to move the actions out of the cf api util file because it shouldn't be so tightly coupled with the actions.

addresses number 2 in https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-6974

## More about the PR

This PR turned out to be a mini refactor as well.
### Use `fetchAllPages` in cf api file
This function was already available but we just weren't using it. We need to revisit the naming of `fetchMany` which actually doesn't get all the pages.
In process of moving to this other function, I removed the use of `userActions` in the CF API file because CF API should not depend on the userActions file. Instead of using one of the userAction promises, I just return the data (via `Promise.resolve`) This decopuling is better but now since the userAction doesn't happen in CF API now, I would need to chain the CF API call with the receivedCall. However, before this PR, all the async work happened in the stores which is an [anti-pattern](https://github.com/18F/cg-dashboard/blob/master/CONTRIBUTING.md#doing-async-work-in-the-stores) so I fixed that. 

### Less Async Work in Store
As mentioned beforehand, now that we removed the userActions from the CF API, the single calls in the store (e.g. `this.load([cfApi.fetchOrgUsers(action.orgGuid)]);`) would now need to be chained with their appropriate userAction
```
return cfApi.fetchOrgUsers(orgGuid)
       .then((users) => userActions.receivedOrgUsers(users, orgGuid));
```
But instead of doing that in the store, just moved those chains to the Fetch actions.

### Change how we track loading state
Previous to this PR, since the store did the async work, it also controlled the loading. (It happens in `this.load` and there's a counter that the base class `BaseStore` has which controls whether it's loading). Now that the actions do the async work, we need a new way to track loading state. The OrgStore and SpaceStore already have `get loading()` but UserStore did not and it would cause the page to stay in an infinite loading spinner. So there's now a `get loading()` for the UserStore that the `userActions` can use and is better than the internal counter.

#### Misc changes
1) For the overview route, the page only loads when `pageActions.loadSuccess` runs. Now that there's more async calls, wrapped them all in a `Promise.all`

2) I moved the loading of all the orgs and spaces to only happen on the overview page. As a result, the app page needed org info so I added an extra `orgActions.fetch(orgGuid);` to the app route

### DIFF
--- a/static_src/test/unit/routes.spec.js
+++ b/static_src/test/unit/routes.spec.js
@@ -8,7 +8,7 @@ import orgActions from '../../actions/org_actions';
 import spaceActions from '../../actions/space_actions';
 import userActions from '../../actions/user_actions';
 import windowUtil from '../../util/window';
-import { checkAuth } from '../../routes';
+import { checkAuth, overview } from '../../routes';
 
 
 describe('routes', function () {
@@ -20,6 +20,19 @@ describe('routes', function () {
 
   afterEach(function () {
     sandbox.restore();
+  });
+
+  describe('overview()', function () {
+    beforeEach(function (done) {
+      sandbox.stub(orgActions, 'fetchAll').returns(Promise.resolve());
+      sandbox.stub(spaceActions, 'fetchAll').returns(Promise.resolve());
+      overview(done);
+    });
+
+    it('fetches page data', function () {
+      expect(orgActions.fetchAll).toHaveBeenCalledOnce();
+      expect(spaceActions.fetchAll).toHaveBeenCalledOnce();
+    });
   });
 
   describe('checkAuth()', function () {
@@ -49,8 +62,6 @@ describe('routes', function () {
 
       it('fetches page data', function () {
         expect(userActions.fetchCurrentUser).toHaveBeenCalledOnce();
-        expect(orgActions.fetchAll).toHaveBeenCalledOnce();
-        expect(spaceActions.fetchAll).toHaveBeenCalledOnce();
       });
     });
 
@@ -73,8 +84,6 @@ describe('routes', function () {
 
       it('fetches page data', function () {
         expect(userActions.fetchCurrentUser).toHaveBeenCalledOnce();
-        expect(orgActions.fetchAll).toHaveBeenCalledOnce();
-        expect(spaceActions.fetchAll).toHaveBeenCalledOnce();
       });
 
       it('calls userActions.fetchCurrentUser() with guids for filtered API calls', function () {
@@ -112,8 +121,6 @@ describe('routes', function () {
 
       it('fetches page data', function () {
         expect(userActions.fetchCurrentUser).toHaveBeenCalledOnce();
-        expect(orgActions.fetchAll).toHaveBeenCalledOnce();
-        expect(spaceActions.fetchAll).toHaveBeenCalledOnce();
       });
 
       it('calls noticeError action', function () {
@@ -147,8 +154,6 @@ describe('routes', function () {
 
       it('does not fetch page data', function () {
         expect(userActions.fetchCurrentUser).not.toHaveBeenCalled();
-        expect(orgActions.fetchAll).not.toHaveBeenCalled();
-        expect(spaceActions.fetchAll).not.toHaveBeenCalled();
       });
     });
   });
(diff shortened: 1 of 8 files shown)


rememberlenny:
Not sure if you saw the failing test @jcscottiii

jcscottiii:
yeah i saw. it's the functional tests.

static_src/test/unit/util/cf_api.spec.js
```diff
@@ -935,17 +935,15 @@ describe('cfApi', function() {
...
+      expect(actual).toEqual(jasmine.any(Function));
     });
   });
 
   describe('fetchOrgUserRoles()', function() {
     it(`should call fetch org user roles with org guid and received org user
         roles action and org guid`, function() {
       var expectedOrgGuid = 'zkjvczcvzwexdvzdfa',
-          spy = sandbox.stub(cfApi, 'fetchMany');
+          spy = sandbox.stub(cfApi, 'fetchAllPages');
```
fureigh: @jcscottiii Should this be `sandbox.spy`?

fureigh: (Capturing answer from convo: 👌 to change it. I'll pop that once my updated dev environment is squared away.)

jcscottiii: yes! thanks!

static_src/util/cf_api.js
```diff
@@ -368,15 +368,13 @@ export default {
...
-                          orgGuid);
+    return this.fetchAllPages(`/organizations/${orgGuid}/users`,
+              (results) => userActions.receivedOrgUsers(results, orgGuid));
   },
 
   fetchOrgUserRoles(orgGuid) {
-    return this.fetchMany(`/organizations/${orgGuid}/user_roles`,
-                          userActions.receivedOrgUserRoles,
-                          orgGuid);
+    return this.fetchAllPages(`/organizations/${orgGuid}/user_roles`,
```
el-mapache: Two space indentation here as well!

static_src/util/cf_api.js
```diff
@@ -368,15 +368,13 @@ export default {
    * @param {String} orgGuid - The guid of the org that the users belong to.
    */
   fetchOrgUsers(orgGuid) {
-    return this.fetchMany(`/organizations/${orgGuid}/users`,
-                          userActions.receivedOrgUsers,
-                          orgGuid);
+    return this.fetchAllPages(`/organizations/${orgGuid}/users`,
```
el-mapache: This is super nit-pickey, but we have a lot of crazy spacing in this code. I would just indent the second line two spaces from the previous one.

static_src/util/cf_api.js
```diff
@@ -247,8 +247,8 @@ export default {
   },
 
   fetchOrgs() {
-    return http.get(`${APIV}/organizations`)
-      .then(res => this.formatSplitResponses(res.data.resources))
+    return this.fetchAllPages('/organizations',
+      (res) => res)
```
el-mapache: Honestly I would leave the parens out of this function. We have inconsistent usage of `(data) => data` and `data => data` throughout the app, and the consensus in the JS community is to omit them, especially for one-line functions. This can be seen in the airbnb js style guide, which is pretty much the gold standard for JS style-guides (and the one we use in this app)!

static_src/util/cf_api.js
jcscottiii: done!

static_src/util/cf_api.js
jcscottiii: done

static_src/test/unit/routes.spec.js
```diff
@@ -22,6 +22,19 @@ describe('routes', function () {
     sandbox.restore();
   });
 
+  describe('overview()', function () {
+    beforeEach(function (done) {
+      sandbox.stub(orgActions, 'fetchAll').returns(Promise.resolve());
```
fureigh: @jcscottiii These are both properly stubs, right?

jcscottiii: yep! since I want to change what they return instead of calling the real thing.

jcscottiii:
@fureigh @rememberlenny @el-mapache this is ready now. LMK what you think!

fureigh:
:shipit: 🎉

PR #1187, opened 2017-08-09
you need to get the spaces first on the org page

jcscottiii:
this case only happens when you refresh. before this PR, if you refreshed the org page, nothing would show. this PR fixes that by enforcing that the spaces be loaded. you have to load them first sequentially before the rest of the promises in case it doesn't finish first.

### DIFF
--- a/static_src/routes.js
+++ b/static_src/routes.js
@@ -62,14 +62,16 @@ export function org(orgGuid, next) {
   spaceActions.changeCurrentSpace();
   appActions.changeCurrentApp();
 
-  orgActions.toggleSpaceMenu(orgGuid);
-  orgActions.fetch(orgGuid);
-  cfApi.fetchSpaces().then(() => spaceActions.fetchAllForOrg(orgGuid));
-  userActions.changeCurrentlyViewedType('org_users');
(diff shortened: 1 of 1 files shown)


static_src/routes.js
```diff
@@ -62,14 +62,16 @@ export function org(orgGuid, next) {
...
-  cfApi.fetchSpaces().then(() => spaceActions.fetchAllForOrg(orgGuid));
-  userActions.changeCurrentlyViewedType('org_users');
-  userActions.fetchOrgUsers(orgGuid);
-  userActions.fetchOrgUserRoles(orgGuid);
-  routerActions.navigate(OrgContainer);
-  next();
+  spaceActions.fetchAll().then(() => {
+    orgActions.toggleSpaceMenu(orgGuid);
+    orgActions.fetch(orgGuid);
+    cfApi.fetchSpaces().then(() => spaceActions.fetchAllForOrg(orgGuid));
```
el-mapache: Does this mean we are fetching spaces two extra times per request?

jcscottiii: good catch! we shouldn't. updating

jcscottiii: updated

PR #1190, opened 2017-08-10
Split out space roles into own property for user

jcscottiii:
Desired behavior: ability to reflect users in the store correctly so that the drop down and user list render correctly with users w/o permissions to the space and users w/ permissions to the space.

Problem found:
When we load the page, the space roles are loaded correctly in the .`roles` property of the user but they are not loaded into the same layout as the org permissions.
Org permissions were mapped from a org guid to an array of strings.
The Space permissions were just an array of strings (only showing perms for the current space)
On the space page, both were loaded into `.roles`. 
The logic to add and remove roles depending on a map `user.roles[guid]`
It worked for org stuff but since it wouldn't find the right role for spaces, it wouldn't actually add or delete to the store (but upon refresh, fresh API calls would get the new data)

Solution:
## Part 1 
Added conditional to `mergeRoles` which creates a `.space_roles` property on the user in the user store and splits it out from `.roles`.
- Reason: now that we have both the org roles and the space roles loaded at the same time on the space page, I chose to keep the similar logic of using a map so the code wouldn't get too complicated (i started down that route initially but it was weird) 

We also, couldn't map into the same `roles` because there could be a org with the same guid as the space guid
if guid-2 and guid were equal, it could get weird.
Example
  ```
  {
  'guid': 'user-guid',
  'roles': { 'guid' : ['org-manager'], 'guid-2': ['space-manager']}
  }
  ```

As a result of this introduction of `space_roles`, you'll see that there's multiple checks for whether we are on the space view or not (determines whether it will use `roles` vs `space_roles` property of the user)
## Part 2
Created `deleteProp` so that the remove all users from space would just drop the `space_roles` property from the user which would make them automagically disappear.
Previous to this, the user itself was just being deleted from the store but we want to keep the user but just remove their space_roles

### DIFF
--- a/static_src/components/user_role_list_control.jsx
+++ b/static_src/components/user_role_list_control.jsx
@@ -82,7 +82,12 @@ export default class UserRoleListControl extends React.Component {
   }
 
   roles() {
-    const roles = this.props.user.roles;
+    let roles;
+    if (this.props.userType == 'space_users') {
+      roles = this.props.user.space_roles;
+    } else {
+      roles = this.props.user.roles;
+    }
     return roles ?
       (roles[this.props.entityGuid] || []) :
       []
--- a/static_src/stores/base_store.js
+++ b/static_src/stores/base_store.js
@@ -78,6 +78,17 @@ export default class BaseStore extends EventEmitter {
     return !Immutable.is(this._data, c);
   }
 
+  deleteProp(guid, prop, cb = defaultChangedCallback.bind(this)) {
+    const index = this._data.findIndex((d) =>
+      d.get('guid') === guid
+    );
+
+    if (index === -1) return cb(false);
+
+    this._data = this._data.deleteIn([index, prop]);
+    return cb(true);
+  }
+
   delete(guid, cb = defaultChangedCallback.bind(this)) {
     const index = this._data.findIndex((d) =>
       d.get('guid') === guid
--- a/static_src/stores/user_store.js
+++ b/static_src/stores/user_store.js
@@ -66,10 +66,10 @@ export class UserStore extends BaseStore {
 
         // Force an update to the cached list of users, otherwise mergeRoles
         // won't be able to find the new user
-        this.mergeMany('guid', users, () => {});
-        this.mergeMany('guid', this.mergeRoles(users, spaceGuid, 'space_roles'), () => {});
+        const updatedUsers = this.mergeRoles(users, spaceGuid,
+          'space_roles');
+        this.mergeMany('guid', updatedUsers, () => { });
         this.emitChange();
-
         break;
       }
 
@@ -107,17 +107,8 @@ export class UserStore extends BaseStore {
       case userActionTypes.USER_ROLES_ADDED: {
         this._saving = false;
         const user = this.get(action.userGuid);
-        const addedRole = action.roles;
-        if (user) {
-          if (!user.roles) user.roles = {};
-          const updatedRoles = new Set(user.roles[action.entityGuid] || []);
-          updatedRoles.add(addedRole);
-          user.roles[action.entityGuid] = Array.from(updatedRoles);
-
-          this.merge('guid', user, () => {});
-        }
-
-        this.emitChange();
+        this.addUserRole(user, action.entityType, action.entityGuid,
+          action.roles, () => this.emitChange());
         break;
       }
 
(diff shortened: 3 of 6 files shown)


static_src/components/users_selector.jsx
```diff
@@ -11,6 +11,8 @@ import { Form, FormSelect } from './form';
 import PanelDocumentation from './panel_documentation.jsx';
 import userActions from '../actions/user_actions';
 import { validateString } from '../util/validators';
+import SpaceStore from '../stores/space_store.js';
```
el-mapache: Sorry for the drive-by comment, I happened to be creeping by the PR list....

The parent component of `UsersSelector` already imports both the User and Space stores, maybe we can leverage that to pass the data this component needs as props? The `stateSetter` pattern should really only be used when there isn't a clear way to get props down from a parent component

static_src/components/user_role_list_control.jsx
```diff
@@ -82,7 +82,12 @@ export default class UserRoleListControl extends React.Component {
   }
 
   roles() {
-    const roles = this.props.user.roles;
+    let roles;
+    if (this.props.userType == 'space_users') {
```
jcscottiii: going to move this `'space_users'` into a constant in the next PR because it's pretty widely used and don't want to create it in this PR and cause confusion.

static_src/components/users_selector.jsx
jcscottiii: i appreciate the feedback. will do! just got the first version working, refactoring now.

static_src/stores/user_store.js
```diff
@@ -137,11 +142,21 @@ export class UserStore extends BaseStore {
         const user = this.get(action.userGuid);
         const deletedRole = action.roles;
         if (user) {
-          const roles = user.roles && user.roles[action.entityGuid];
-          if (roles) {
-            const idx = deletedRole && roles.indexOf(deletedRole);
-            if (idx > -1) {
-              roles.splice(idx, 1);
+          if (action.entityType === 'space') {
+            const roles = user.space_roles[action.entityGuid];
```
jcscottiii: need to add check of `user.space_roles`

el-mapache:
So, I don't think its terrible to have the separate `space_roles` object. But, does it seems like too much over head to have a data structure like:
```
roles: {
  orgs: {
     guid: 'string',
     guid: 'string'
  },
  spaces: {
    guid: 'string',
    guid: 'string
  }
}
```

This might be more of a v2 refactor (or not, since, I think, objects should generally be as flat as possible). Having a separate object just for space roles is definitely the easiest thing right now.

jcscottiii:
i like that idea but let's save that for the v2.

rememberlenny:
Testing now locally

static_src/stores/user_store.js
```diff
@@ -109,14 +109,19 @@ export class UserStore extends BaseStore {
         const user = this.get(action.userGuid);
         const addedRole = action.roles;
         if (user) {
-          if (!user.roles) user.roles = {};
-          const updatedRoles = new Set(user.roles[action.entityGuid] || []);
-          updatedRoles.add(addedRole);
-          user.roles[action.entityGuid] = Array.from(updatedRoles);
-
+          if (action.entityType === 'space') {
+            if (!user.space_roles) user.space_roles = {};
```
el-mapache: How about breaking these up into a method? Both paths seem fairly similar, and a descriptive method call never hurts readability.

rememberlenny:
Looks good! Confirmed its working.

static_src/stores/base_store.js
```diff
@@ -78,6 +78,17 @@ export default class BaseStore extends EventEmitter {
...
   }
 
+  deleteProp(guid, prop, cb = defaultChangedCallback.bind(this)) {
+    const index = this._data.findIndex((d) =>
+      d.get('guid') === guid
+    );
+
+    if (index === -1) return cb(false);
+
+    this._data = this._data.deleteIn([index, prop]);
```
rememberlenny: This was the key thing missing before. I got so caught up with the `_data` that I didnt consider the issue was in how the `_data` was being rendered. Thanks for synthesizing this.

static_src/stores/user_store.js
```diff
@@ -66,10 +66,10 @@ export class UserStore extends BaseStore {
 
         // Force an update to the cached list of users, otherwise mergeRoles
         // won't be able to find the new user
-        this.mergeMany('guid', users, () => {});
-        this.mergeMany('guid', this.mergeRoles(users, spaceGuid, 'space_roles'), () => {});
+        const updatedUsers = this.mergeRoles(users, spaceGuid,
+          'space_roles');
```
rememberlenny: And that was it.

static_src/test/unit/stores/base_store.spec.js
```diff
@@ -214,6 +214,80 @@ describe('BaseStore', () => {
     });
   });
 
+  describe('deleteProp', function() {
```
rememberlenny: 👍

fureigh:
Seconding @el-mapache (https://github.com/18F/cg-dashboard/pull/1190#issuecomment-321638309).

For v2, what if we prepended `org-` or `space-` to each GUID instead? ex:

```
roles: {
  'org-[guid]': 'string',
  'org-[guid]': 'string',
  'space-[guid]': 'string',
  'space-[guid]': 'string'
}
```

PR #1191, opened 2017-08-10
Fetch Services and Plans once in <Marketplace/> component

el-mapache:
Rather than perform multiple fetches and rely on state in component, fetch services and associated plans once in <Marketplace />, and pass those down to child components.

I _think_ this also resolves the bug where a user on a space page sees a service, but no associated plans.

### DIFF
--- a/static_src/components/service_plan_list.jsx
+++ b/static_src/components/service_plan_list.jsx
@@ -48,10 +48,6 @@ export default class ServicePlanList extends React.Component {
     return columns;
   }
 
-  get rows() {
-    return this.props.plans;
-  }
-
   cost(plan) {
     const cost = ServicePlanStore.getCost(plan);
     if (plan.free || cost === 0) return 'Free';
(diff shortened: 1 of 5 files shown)


static_src/components/service_plan_list.jsx
```diff
@@ -9,11 +9,14 @@ import serviceActions from '../actions/service_actions.js';
...
 const propTypes = {
+  plans: PropTypes.array,
   serviceGuid: PropTypes.string
 };
 
-function stateSetter(serviceGuid) {
-  const servicePlans = ServicePlanStore.getAllFromService(serviceGuid).sort((a, b) => {
+const empty = plans => !ServicePlanStore.loading && !plans.length;
+
+const sortPlansByCost = (plans) =>
```
jcscottiii: when is this used?

el-mapache: Right you are! Adding it in.

el-mapache: @jcscottiii updated!

PR #1192, opened 2017-08-11
Restore call to `receivedSpaceUserRoles` after inviting user to space

el-mapache:
**WIP**

Fixes bug where user was invited to space but not populated in the space user list. There is still an outstanding bug where user email is not populated in list.


el-mapache:
Closing in favor of #1194

PR #1194, opened 2017-08-11
fix space user from automatically being populated upon invite

jcscottiii:
in prod, if you try to invite a user to the space by e-mail, they will
not be populated in the user list automatically. you need to refresh to
see them.

### DIFF
--- a/static_src/constants.js
+++ b/static_src/constants.js
@@ -260,15 +260,13 @@ const userActionTypes = keymirror({
   USER_INVITE_STATUS_UPDATED: null,
   // Action to trigger when user list notice is created.
   USER_LIST_NOTICE_CREATED: null,
-  // Action to trigger email sent to user with cloud.gov invite url
-  // and associate user to organization.
+  // Action to associate user to an organization.
   USER_ORG_ASSOCIATE: null,
-  // Action when user is associated to organization.
+  // Action when user is associated to an organization.
   USER_ORG_ASSOCIATED: null,
-  // Action to trigger email sent to user with cloud.gov invite url
-  // and associate user to space.
+  // Action to associate user to a space.
   USER_SPACE_ASSOCIATE: null,
-  // Action when user is associated to space.
+  // Action when user is associated to a space.
   USER_SPACE_ASSOCIATED: null,
   // Display the user associated to org.
   USER_ASSOCIATED_ORG_DISPLAYED: null,
--- a/static_src/constants.js
+++ b/static_src/constants.js
@@ -260,10 +260,16 @@ const userActionTypes = keymirror({
   USER_INVITE_STATUS_UPDATED: null,
   // Action to trigger when user list notice is created.
   USER_LIST_NOTICE_CREATED: null,
-  // Action to trigger email sent to user with cloud.gov invite url.
+  // Action to trigger email sent to user with cloud.gov invite url
+  // and associate user to organization.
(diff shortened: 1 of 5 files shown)


static_src/constants.js
```diff
@@ -260,10 +260,16 @@ const userActionTypes = keymirror({
...
-  // Action to trigger email sent to user with cloud.gov invite url.
+  // Action to trigger email sent to user with cloud.gov invite url
+  // and associate user to organization.
   USER_ORG_ASSOCIATE: null,
-  // Action to associate user to organization on the server.
+  // Action when user is associated to organization.
   USER_ORG_ASSOCIATED: null,
+  // Action to trigger email sent to user with cloud.gov invite url
+  // and associate user to space.
+  USER_SPACE_ASSOCIATE: null,
```
el-mapache: I think this comment is slightly misleading, since this action type doesn't trigger the email invite as a result of it being dispatched.

jcscottiii: you're right. i just copied the one above. will change both of them.

jcscottiii:
`userAction. fetchSpaceUserRoles` is called in routes.js @el-mapache

el-mapache:
I think we should still be able to combine the calls somehow? `fetchSpaceUserRoles` doesn't actually get roles, it gets all the users in a given space.

jcscottiii:
that one call gets both the users and their roles. `userAction.fetchSpaceUserRoles` calls `cfApi.fetchSpaceUserRoles` (by listing the roles, you get all the users as well. as a side note: we should probably do this one call for the org stuff, currently it gets the org users in one call then the roles in another call, then merges)

image

https://apidocs.cloudfoundry.org/268/spaces/retrieving_the_roles_of_all_users_in_the_space.html

el-mapache:
Right, I understand that, I guess I was just saying that the function is not well named. The spaces API specifically doesn't let you get all the users, it only has a function to get the users and their roles within the space. This isn't necessarily calling out this code specifically, the method is called the same thing in the cloud foundry api and it seems confusing.

el-mapache:
Like you said, Orgs also exposes the same call to get users and their roles for orgs, so we should for sure use that going forward.

jcscottiii:
re: https://github.com/18F/cg-dashboard/pull/1194#issuecomment-322196056
maybe rename the userAction to `fetchSpaceUsersAndRoles`?

> This PR adds some new methods to associate users to spaces, but what about the existing store code that handles the SPACE_USER_ROLES_RECEIVED action type? Do we need both ways to associate the user to the space?

We need both. The associated action toggles `_inviteInputActive` (even though it's not being used which is another story for another PR). These types of indication of states depending on if you're fetching or inviting someone new are exclusive (when we use them). BUT the merge logic that they both use, that could be consolidated to a new shared function.

jcscottiii:
@el-mapache i don't see a comment there?
> The comment on line 67 in user_actions doesn't apply anymore, since the method call it was commenting has been removed.

el-mapache:
@jcscottiii Whoops! I typed `user_actions` but it should be `user_store`

el-mapache:
@jcscottiii So you are saying that we were previously overloading `SPACE_USER_ROLES_RECEIVED` and we should have had a discrete action type just for associating users to spaces all along? I can get behind that.

jcscottiii:
@el-mapache should be ready to go!

PR #1197, opened 2017-08-18
Create default text for specified services with multiple required params

rememberlenny:
screen shot 2017-08-18 at 1 44 27 pm

PR #1199, opened 2017-08-21
New button text around multiple param service instances

rememberlenny:
screen shot 2017-08-21 at 5 44 35 pm

### DIFF
--- a/static_src/components/service_plan.jsx
+++ b/static_src/components/service_plan.jsx
@@ -12,6 +12,12 @@ const propTypes = {
   })
 };
 
+// This should be removed when solution is setup
+// for service instancemultiple params
+const CF_CLI_SERVICE_LIST = [
+  'cdn-route', 'cloud-gov-identity-provider', 'cloud-gov-service-account'
+];
+
 class ServicePlan extends React.Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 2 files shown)


el-mapache:
This looks fine, is there going to be a corresponding PR to show a link to the documentation?

static_src/test/unit/components/service_plan.spec.jsx
```diff
@@ -55,3 +55,56 @@ describe('<ServicePlan />', () => {
     expect(props.onAddInstance.calledWith(props.plan.guid)).toBe(true);
   });
 });
+
+describe('<ServicePlan with multi param plan />', () => {
+  it('renders an action button for cdn-route', () => {
```
el-mapache: I think all of these test cases could be collapsed into a single `it` block with a loop, almost all of the code is identical except for the names for the services.

`it('renders a link to documentation')` or something similar.

static_src/test/unit/components/service_plan.spec.jsx
```diff
@@ -55,3 +55,56 @@ describe('<ServicePlan />', () => {
     expect(props.onAddInstance.calledWith(props.plan.guid)).toBe(true);
   });
 });
+
+describe('<ServicePlan with multi param plan />', () => {
```
el-mapache: Nitpick, I think this should be 'with multi param plan', since 'ServicePlan' is already a top level description block.

static_src/components/service_plan.jsx
```diff
@@ -12,6 +12,12 @@ const propTypes = {
   })
 };
 
+// This should be removed when solution is setup
+// for service instancemultiple params
+const CF_CLI_SERVICE_LIST = [
```
jcscottiii: is there some way we can use a common structure between this one and the one you introduced in this other PR? https://github.com/18F/cg-dashboard/pull/1197/files#diff-49e04869fdff69f2ac117d46c205eaf9R23

rememberlenny: It is, but do you think its worth it considering these will be removed?

PR #1200, opened 2017-08-23
Adjust copy explaining org users

el-mapache:
* Let managers know that users with no roles continue to exist in the
organization

screen shot 2017-08-23 at 10 11 31 am

### DIFF
--- a/static_src/components/user_list.jsx
+++ b/static_src/components/user_list.jsx
@@ -83,7 +83,8 @@ export default class UserList extends React.Component {
         To invite a user and give them roles, see <a href={ config.docs.invite_user }>Managing Teammates</a>.&nbsp;
         <b>
           Removing all roles does not remove a user from an organization.
-          Users with no roles will remain in the organization, but they will be unable to perform any actions.
+          Users with no roles will remain in the organization until you remove them,
+          but they will be unable to perform any actions.
         </b>
       </span>
     );
--- a/static_src/components/user_list.jsx
+++ b/static_src/components/user_list.jsx
@@ -75,16 +75,27 @@ export default class UserList extends React.Component {
     return (this.props.userType === 'org_users') ? 'Organization' : 'Space';
   }
 
+  get inviteDocumentation() {
+    if (!config.docs.invite_user) return null;
+
+    return (
+      <span>
+        To invite a user and give them roles, see <a href={ config.docs.invite_user }>Managing Teammates</a>.&nbsp;
+        <b>
+          Removing all roles does not remove a user from an organization.
+          Users with no roles will remain in the organization, but they will be unable to perform any actions.
+        </b>
+      </span>
+    );
+  }
+
   get documentation() {
     return (
       <PanelDocumentation description>
         <p>
           { this.userTypePretty } Managers can change these roles. For details
             about these roles, see <a href="https://docs.cloudfoundry.org/concepts/roles.html#roles">Cloud Foundry roles and permissions</a>.
-          { config.docs.invite_user &&
-            <span> To invite a user and give them roles, see <a href={ config.docs.invite_user }>Managing Teammates</a>.
-            </span>
-          }
+          { this.inviteDocumentation }
         </p>
       </PanelDocumentation>
     );


el-mapache:
Is this copy approaching what we want? The section is already a bit wordy and I'm hesitant to add multiple paragraphs explaining the difference in command-line generated user lists and the one visible on the dashboard.

suprenant:
@el-mapache I think this is close, and – yeah – we don't want it to be too wordy. What do you think of this? 

_Organization Managers can change these roles. For details about these roles, see Cloud Foundry roles and permissions. For more information about inviting users and assigning roles, see Managing Teammates. Removing all roles does not remove a user from an organization. Users with no roles will remain in the organization until you remove them, but they will be unable to perform any actions._

suprenant:
@femmebot I believe there's a compliance reason that we don't automatically delete users even if they have no roles, and users do inherit an org role if they are added to a space within that org. I believe the above will alleviate at least some of the confusion we're seeing so suggest we move forward with this fix and adjust again later if we're still hearing customers ask about this. 

Only displaying the list to users with appropriate roles/permissions to add/assign/remove could be a separate story. I'll look into whether we've considered this before and will write a new story if we haven't.

el-mapache:
@femmebot It seems like a conscious choice to display the list even if you aren't an org manager, maybe because anyone can view the list of users via the command line? 

Non-managers can't edit anything in the user list and also see less of an explanation, maybe we should just hide all the documentation if the list can't be interacted with?

suprenant:
@femmebot you may be right. Let's move forward with this implementation for now so we can work on other stories – we can always revisit and make further refinements later.

el-mapache:
@femmebot Based on the docs linked in the ticket, the root of the confusion seems to stem from the two different user enumeration commands that the CLI provides. I still think a decent solution might involve some kind of flagging or separation of users with no roles, although I'm not sure what that looks like.

Do you all feel like this is good enough to try out and see if confusion is reduced at all?

suprenant:
Some separation/flagging is a nice idea. @el-mapache @femmebot okay, if you want to take one more day and see where that can get us, let's do it.

el-mapache:
@suprenant I don't know if another day is necessarily going to produce a solution that is any better than the one outlined in the ticket. I think this ticket can be closed, and a new one can be opened to focus on figuring out what the root of this problem is and how we can best address it.

suprenant:
@el-mapache OK, let's call this one done, then, and I'll flag the above as possible follow up work for another time.

brittag:
The root problem is that Cloud Foundry itself behaves in an unexpected way - it doesn't remove users from the org when you remove all their roles. See the discussions linked from the [bottom of this doc](https://docs.google.com/document/d/19oa16uu3WGoXHezCtAnMD7h9TGZmcg_my6llEX03vko/edit#heading=h.eikp32gh9cpx) - https://github.com/cloudfoundry/cli/issues/781

brittag:
Also people with org user status but no org roles or space roles can still run the commands `cf org-users -a [org]` and `cf org-users [org]` and see results...

el-mapache:
@brittag I actually wouldn't expect that behavior. I would expect a user is only going to be removed when I explicitly delete their record. For example, I might create a user in the organization but only intend to give them space roles.

el-mapache:
@brittag Yes, but having an account with cloud.gov isn't the same as belonging to an organization. You can be a member of cloud.gov and belong to no organizations, and read all the same information, correct?

brittag:
@el-mapache If you have a cloud.gov account but no "org user" status on any orgs, you can't see any results from running `cf org-users -a [org]` and `cf org-users [org]` - you just get "Organization [org] not found"

el-mapache:
Ok, that makes sense. I think I misread your above comment. I still don't consider it a fault of the API if users don't understand that roles and associations are things that a user has, rather than the _actual_ user.

brittag:
Users don't expect that an "org user" remains after you remove their org roles because the [CF user-facing docs page about roles/permissions](https://docs.cloudfoundry.org/concepts/roles.html) doesn't document the existence of "org user" status (how it gets created, what it can do, how you remove it, etc.), and if you use the CF CLI default command of `cf org-users [org]`, you don't see "org users", just people with org roles. I'd love to get this fixed in the CF docs and CF CLI but that's a bigger project.

brittag:
Correct, no compliance requirements here, just trying to explain CF behavior.

brittag:
Filed a related PR just in case it helps: https://github.com/cloudfoundry/docs-cloudfoundry-concepts/pull/75

PR #1204, opened 2017-08-25
remove disabled coveralls.

jcscottiii:
We aren't using it. so no need for it


jcscottiii:
@el-mapache good catch will remove!

jcscottiii:
@el-mapache done!

PR #1207, opened 2017-08-25
Remove the cost column from service plans

rememberlenny:
Favro https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-5649


suprenant:
@rememberlenny since we groomed this story, I learned that we do have plans to start charging for some of these services (maybe as soon as a few months away), so "Free" could be confusing to users once this changes. So, I suggest we just remove the Cost column for now, and we can potentially add it back with more accurate information in the future.

jcscottiii:
Looks good. Just squash the commits and fix the tests

PR #1209, opened 2017-08-26
Create an check before displaying user data without a username

rememberlenny:
# Problem

Users that have been deleted, but still have roles associate with orgs/spaces show up as blank.

# Solution

Short term: Dont show users in the user list if they dont have a username.

Longer term: Change the functions that "delete" users and make sure to run a check for any associated org/space roles.

screen shot 2017-08-26 at 5 51 06 pm

### DIFF
--- a/static_src/components/user_list.jsx
+++ b/static_src/components/user_list.jsx
@@ -126,47 +126,49 @@ export default class UserList extends React.Component {
         { this.documentation }
         <ComplexList>
           { this.props.users.map((user) => {
-            let actions;
-            if (this.props.onRemove) {
-              let button = <span></span>;
-              if (this.props.currentUserAccess) {
-                if (this.props.userType === 'org_users') {
-                  buttonText = 'Remove User From Org';
-                } else if (this.props.userType === 'space_users') {
-                  buttonText = 'Remove All Space Roles';
+            if (!!user.username) {
+              let actions;
+              if (this.props.onRemove) {
+                let button = <span></span>;
+                if (this.props.currentUserAccess) {
+                  if (this.props.userType === 'org_users') {
+                    buttonText = 'Remove User From Org';
+                  } else if (this.props.userType === 'space_users') {
+                    buttonText = 'Remove All Space Roles';
+                  }
+                  button = (
+                    <Action
+                      style="base"
+                      clickHandler={ this._handleDelete.bind(this, user.guid) }
+                      label="delete">
+                      <span>{ buttonText }</span>
+                    </Action>
+                  );
                 }
-                button = (
-                  <Action
-                    style="base"
-                    clickHandler={ this._handleDelete.bind(this, user.guid) }
-                    label="delete">
-                    <span>{ buttonText }</span>
-                  </Action>
+                actions = (
+                  <ElasticLineItem align="end">
+                    { button }
+                  </ElasticLineItem>
                 );
               }
-              actions = (
-                <ElasticLineItem align="end">
-                  { button }
-                </ElasticLineItem>
+
+              return (
+                <ElasticLine key={ user.guid }>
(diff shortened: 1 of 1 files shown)


jcscottiii:
I disagree with the long term solution
> Longer term: Change the functions that "delete" users and make sure to run a check for any associated org/space roles.

The dashboard itself doesn't delete users from the CF or UAA databases.
(I looked into why we see those blank users and it's because the user exists in the CF database but not in the UAA database, which is why it can't find the username).
There's something external that goes on that gets a user into the state mentioned above.

Long term, we should have a cron job that watches for that (outside the scope of the dashboard) and it will clean up these half users.

static_src/components/user_list.jsx
```diff
@@ -126,47 +126,49 @@ export default class UserList extends React.Component {
...
+                    buttonText = 'Remove User From Org';
+                  } else if (this.props.userType === 'space_users') {
+                    buttonText = 'Remove All Space Roles';
+                  }
+                  button = (
+                    <Action
+                      style="base"
+                      clickHandler={ this._handleDelete.bind(this, user.guid) }
+                      label="delete">
+                      <span>{ buttonText }</span>
```
el-mapache: This span isn't strictly necessary

static_src/components/user_list.jsx
```diff
@@ -126,47 +126,49 @@ export default class UserList extends React.Component {
...
-              let button = <span></span>;
-              if (this.props.currentUserAccess) {
-                if (this.props.userType === 'org_users') {
-                  buttonText = 'Remove User From Org';
-                } else if (this.props.userType === 'space_users') {
-                  buttonText = 'Remove All Space Roles';
+            if (!!user.username) {
+              let actions;
+              if (this.props.onRemove) {
+                let button = <span></span>;
```
el-mapache: I'd suggest just going with `null` for empty nodes.

static_src/components/user_list.jsx
```diff
@@ -126,47 +126,49 @@ export default class UserList extends React.Component {
...
           { this.props.users.map((user) => {
-            let actions;
-            if (this.props.onRemove) {
-              let button = <span></span>;
-              if (this.props.currentUserAccess) {
-                if (this.props.userType === 'org_users') {
-                  buttonText = 'Remove User From Org';
-                } else if (this.props.userType === 'space_users') {
-                  buttonText = 'Remove All Space Roles';
+            if (!!user.username) {
```
el-mapache: Keeping a map here is going to create a sparse array with `undefined`s where no JSX is returned. I would filter the list first on the `user.username` check. Doing that also provides the extra benefit of reducing the levels of nested `if` checks in the function!

static_src/components/user_list.jsx
```diff
@@ -126,47 +126,49 @@ export default class UserList extends React.Component {
...
+                if (this.props.currentUserAccess) {
+                  if (this.props.userType === 'org_users') {
+                    buttonText = 'Remove User From Org';
+                  } else if (this.props.userType === 'space_users') {
+                    buttonText = 'Remove All Space Roles';
+                  }
+                  button = (
+                    <Action
+                      style="base"
+                      clickHandler={ this._handleDelete.bind(this, user.guid) }
```
el-mapache: Calling a `bind` directly inside the render function has performance implications, as the function will be rebound (creating a memory allocation + running the garbage collector) every time the render function runs. Additionally, since we are always assigning a new value to the `clickHandler` property, the component will always be re-rendered.

We should probably refactor this. I think simplest way would be to break the logic contained within the `map`'s callback out into a component.

jcscottiii:
After talking to @brittag last week, we found out that this blank user is a separate problem. The blank user problem that she was seeing is different but was fixed also. Closing this.

PR #1210, opened 2017-08-26
Create the new USWDS banner

rememberlenny:
Based on the component mentioned here: https://components.standards.usa.gov/components/detail/banner.html

save


static_src/components/disclaimer.jsx
```diff
@@ -15,20 +15,52 @@ export default class Disclaimer extends React.Component {
 
   render() {
     let flagImg;
+    let dotGovIconImg;
+    let httpsIconImg;
+    const dotGovIcon = require('cloudgov-style/img/icon-dot-gov.svg');
+    const dotGovAlt = 'Dot gov';
+    dotGovIconImg = <img className={ this.styler('usa-banner-icon','usa-media_block-img') } src={ dotGovAlt } src={ dotGovIcon }</img>;
```
jcscottiii: i would be interested to hear @el-mapache's opinion on the `styler`. i have seen him remove that. wondering if we can do it here in this PR?

el-mapache: I think we can remove the styler. I think the main thing it offers over the `classnames` package is that it de-dupes class names by default, something that I haven't seen come up at all in the code base.

el-mapache: I think we can just put this image tag inline, no need to save it in a variable.

static_src/components/disclaimer.jsx
```diff
@@ -15,20 +15,52 @@ export default class Disclaimer extends React.Component {
...
+    const httpsIcon = require('cloudgov-style/img/icon-https.svg');
+    const httpsAlt = 'Https';
+    httpsIconImg = <img className={ this.styler('usa-banner-icon','usa-media_block-img') } src={ httpsAlt } src={ httpsIcon }</img>;
```
el-mapache: As above, I don't think we need to store this tag in a variable.

I think the first `src` attribute should be `alt`.

Also, image tags are self closing, you can omit the `</img>`

rememberlenny: Sounds good. I was following the pattern on the previous template. I'll update accordingly.

el-mapache:
@rememberlenny I can help finish this up, if that is alright

rememberlenny:
Please do!

--
Leonard Bogdonoff
rememberlenny.com | @rememberlenny

el-mapache:
@rememberlenny Do we need to copy the icons over to the cloud.gov style repo? Should we make some new css components for these content blocks as well?

Also, how do we get the chevron icons to display? The guide says to use `svg` tags use namespaced xml, which React won't parse.

el-mapache:
Here are some POC screenshots, the css isn't correct yet:

screen shot 2017-08-31 at 10 33 01 am

screen shot 2017-08-31 at 10 33 10 am

static_src/components/disclaimer.jsx
```diff
@@ -1,33 +1,102 @@
...
+    });
   }
 
   render() {
+    const { expanded } = this.state;
+    const hidden = !expanded;
+    const panelClass = classnames('usa-banner-content usa-grid usa-accordion-content', {
+      hide: hidden
+    });
+    const bannerBarClass = classnames('usa-banner usa-disclaimer disclaimer-no_sidebar', {
```
el-mapache: One question, why do we need the expanded class here?

rememberlenny: That was initially taking care of the chevon.

PR #1211, opened 2017-08-28
Help users find docs about services

brittag:
In order to help dashboard users use services successfully, here are a few copy suggestions for review (I didn't test these locally, so they probably need adjustment):

* At the top of the marketplace section, link to a page with docs about the services (https://cloud.gov/docs/services/). This may help people find the correct price information about services (since the incorrect price column is being removed in https://github.com/18F/cg-dashboard/pull/1207).
* At the top of the service instance table, update the old location of the marketplace (left instead of below), and remove the old note about binding using the CLI (since you can do that on the dashboard).
* Update a few of the docs links to their current locations (the redirects were working, but it's nice to link to the current location).

### DIFF
--- a/static_src/components/marketplace.jsx
+++ b/static_src/components/marketplace.jsx
@@ -63,9 +63,10 @@ export default class Marketplace extends React.Component {
       <PanelDocumentation description>
         <p>
           Use this marketplace to create service instances for apps in this space. Then bind service instances to apps.
+          See <a href="https://cloud.gov/docs/services/">docs for these services</a>, and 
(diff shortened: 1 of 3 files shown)


brittag:
cc @suprenant since this is a followup to a tiny chat we had!

jcscottiii:
Snapshot of how it looks: 
image

jcscottiii:
I'm not sure what else @suprenant is looking for but I'll let him review before we merge.

suprenant:
Looks good to me.

PR #1212, opened 2017-08-29
Update to 1.16.10

jcscottiii:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cloudgov-dashboard",
-  "version": "1.16.9",
+  "version": "1.16.10",
```
el-mapache: Just wondering, at what point do we roll this over to the next minor version? Do we stop at 10 or do we just stop when we feel like we've added a more significant piece of functionality?

jcscottiii: we'll bump it next major feature for the dashboard.

el-mapache: But what about the next minor version? For example, when does it become `1.17.0` vs `1.16.11`?

jcscottiii: the next major feature will go to 1.17.0. Right now since we're doing mostly bug fixes, that's why i'm only bumping the patch number.
More about it here: https://github.com/18F/cg-dashboard/blob/adae679b1a7bb0611e2e02f5851cdfd9f786b350/RELEASE.md#updating-version-number

(Also, I think we may have messed up because we should've bumped minor when we did the user invite flow a few PIs ago)

el-mapache: Sorry, I guess my question and/or example weren't clear. I was asking about minor, not major versions. But it seems like it's entirely up to the person doing the version bumping. Thanks!

rememberlenny: @el-mapache To your question, I would guess we go to `1.16.11`.

PR #1219, opened 2017-09-05
Act as trusted proxy for client addresses to CF API proxy

wjwoodson:


controllers/secure.go
```diff
@@ -105,8 +105,17 @@ func (c *SecureContext) submitRequest(rw http.ResponseWriter, req *http.Request,
 		request.Header.Set("Content-Type", contentHeader)
 	}
 
-	// Set headers for requests to CF API proxy
+	// Get RemoteAddr from the request
 	clientIP := strings.Split(req.RemoteAddr, ":")[0]
+
+	// We are behind a gorouter, nginx, and ELB. gorouter and ELB set addresses
+	// in X-Forwarded-For. The 2nd from last should be "true" egress IP.
+	xff := req.Header["X-Forwarded-For"]
```
jmcarp: I was thinking of using the first non-localhost address:

```go
for idx := len(xff) - 1; idx >= 0; idx-- {
  if xff[idx] == "127.0.0.1" {
    continue
  }
  return xff[idx]
}
```

controllers/secure.go
```diff
@@ -103,6 +105,17 @@ func (c *SecureContext) submitRequest(rw http.ResponseWriter, req *http.Request,
...
 	}
+
+	// Get RemoteAddr from the request
+	clientIP, err := GetClientIP(req)
+	if err == nil {
+		// Set headers for requests to CF API proxy
+		request.Header.Add("X-Client-IP", clientIP)
+		request.Header.Add("X-TIC-Secret", c.Settings.TICSecret)
+	} else {
+		log.Println(err)
```
jcscottiii: will want to do something like
```go
 		rw.WriteHeader(http.StatusInternalServerError)
 		rw.Write([]byte("unknown error. try again"))
```

you can replace the "unknown error" part with the actual error with some context: "Unable to get Client IP: " + err.Error()

controllers/secure.go
```diff
@@ -103,6 +105,17 @@ func (c *SecureContext) submitRequest(rw http.ResponseWriter, req *http.Request,
 	if contentHeader := req.Header.Get("Content-Type"); len(contentHeader) > 0 {
 		request.Header.Set("Content-Type", contentHeader)
 	}
+
+	// Get RemoteAddr from the request
+	clientIP, err := GetClientIP(req)
```
jcscottiii: thinking about local development. not sure we should do the computation for TIC unless we have provided a TIC Secret.

helpers/settings.go
```diff
@@ -119,6 +121,9 @@ func (s *Settings) InitSettings(envVars EnvVars, env *cfenv.App) error {
 	if len(envVars.SMTPHost) == 0 {
 		return errors.New("Unable to find '" + SMTPHostEnvVar + "' in environment. Exiting.\n")
 	}
+	if len(envVars.TICSecret) == 0 {
```
jcscottiii: if we are making TIC secret mandatory, we need to find a way to replicate it locally. but i don't think we can due to our setup vs PCFDev.
Also, if we make it mandatory, we need to add it here too https://github.com/18F/cg-dashboard/blob/master/docker-compose.yml#L31

Whether or not it's mandatory could also be resolved by answering this comment: https://github.com/18F/cg-dashboard/pull/1219/files#r137849061

controllers/secure.go
```diff
@@ -107,13 +107,18 @@ func (c *SecureContext) submitRequest(rw http.ResponseWriter, req *http.Request,
...
-	clientIP, err := GetClientIP(req)
-	if err == nil {
-		// Set headers for requests to CF API proxy
-		request.Header.Add("X-Client-IP", clientIP)
-		request.Header.Add("X-TIC-Secret", c.Settings.TICSecret)
-	} else {
-		log.Println(err)
+	if c.Settings.TICSecret != "" {
+		clientIP, err := GetClientIP(req)
+		if err != nil {
```
jcscottiii: `if err != nil {` -> `if err != nil || clientIP == ""`

jmcarp: But `GetClientIP` can return an empty string and a `nil` error. This happens when the proxied request has no `RemoteAddr`, when we generate a request like this: https://github.com/18F/cg-dashboard/blob/master/controllers/uaa.go#L280-L282.

jmcarp:
WDYT @jcscottiii ?

helpers/testhelpers/testhelpers.go
```diff
@@ -88,6 +90,23 @@ var ValidTokenData = map[string]interface{}{
 	"token": oauth2.Token{Expiry: time.Time{}, AccessToken: "sampletoken"},
 }
 
+// EchoResponseHandler is a normal handler for responses received from the proxy requests.
+func EchoResponseHandler(rw http.ResponseWriter, response *http.Response) {
+	for header := range response.Header {
+		// fmt.Println("DEBUG GENERIC HANDLER", header, response.Header.Get(header))
```
jcscottiii: remove this comment

jcscottiii:
@jmcarp @wjwoodson sorry about the delay. just one more comment. otherwise LGTM

jmcarp:
Thanks @jcscottiii, updated to drop that comment.

Issue #757, opened 2016-11-22
Fix memory leak on space detail view


jmcarp:
On some space detail views, I see a `possible EventEmitter memory leak detected` warning in the console--for example, https://dashboard.fr.cloud.gov/#/org/4175a5b2-78a4-42b0-94c2-280f8be41e7f/spaces/4a543094-ecff-43b5-a4a2-43886a892e4a/apps/cded9067-5b8e-4fcb-9acc-24d44b88983c.

<img width="891" alt="screen shot 2016-11-21 at 7 43 40 pm" src="https://cloud.githubusercontent.com/assets/1633460/20506543/dbc7117c-b022-11e6-9a3f-e0e8757def9d.png">

We may have unused event listeners hanging around that aren't being cleaned up, in which case we should drop them when we're done using them. If not, we can bump the limit on listeners: http://www.jongleberry.com/understanding-possible-eventemitter-leaks.html

msecret:
Thanks @jmcarp, I've seen this before and keep forgetting to write it somewhere to track it. I've investigated a bit and I don't think theres a leak. I think it's just a lot of listeners due to how the flux system works. We should probably set the max to more then 11.

PR #1223, opened 2017-09-06
Resolve console warning when activity log component unmounts

el-mapache:
* Previously, the activity log component listened to the AppStore's
change event, which fires each time a new route is visited. Because the
activity log could unmount on a route transition, this led to a race
condition in which react would begin processing the `stateSetter`
function after the component had been removed from the node tree.

### DIFF
--- a/static_src/components/activity_log.jsx
+++ b/static_src/components/activity_log.jsx
@@ -1,7 +1,6 @@
 
 import PropTypes from 'prop-types';
 import React from 'react';
-
 import Action from './action.jsx';
 import AppActivity from './app_activity/app_activity.jsx';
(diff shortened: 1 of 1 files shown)


el-mapache:
Just a little cleanup/refactoring!

jcscottiii:
@el-mapache LGTM just curious do have the exact warning? is it related to https://github.com/18F/cg-dashboard/issues/757

el-mapache:
No, I haven't seen that error in the console! This warning is:

screen shot 2017-09-07 at 10 15 19 am

We get them for the `Breadcrumbs` and `Users` component as well, but those are going to be more work to resolve.

PR #1228, opened 2017-09-12
Update tests to send well-formed GUID, better compare JSON responses, fix escaped URLs in tests

aeijdenberg:
This PR should fix the broken tests from yesterday - sorry about that.

(is there a way to set CI up to test PRs before merging?)

### DIFF
--- a/helpers/testhelpers/testhelpers.go
+++ b/helpers/testhelpers/testhelpers.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cloudfoundry-community/go-cfenv"
 	"github.com/gocraft/web"
 	"github.com/gorilla/sessions"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 	"golang.org/x/oauth2"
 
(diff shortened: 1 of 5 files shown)


helpers/testhelpers/testhelpers.go
```diff
@@ -286,10 +287,39 @@ func PrepareExternalServerCall(t *testing.T, c *controllers.SecureContext, testS
 	return nil, nil, nil
 }
 
+// comparePossiblyJSONStrings will try to parse both as JSON. If it succeeds, it will marshal them both out and compare.
+// Since JSON doesn't actually have a canonical form, this may be flaky, but likely less flaky than previous behavior.
+// If anything fails, it will compare the original string as strings
+func comparePossiblyJSONStrings(a, b string) bool {
+	fallback := (a == b)
+
+	var oa, ob interface{}
```
jcscottiii: we can use testify's `JSONEq` https://github.com/stretchr/testify/blob/890a5c3458b43e6104ff5da8dfa139d013d77544/assert/assertions.go#L1128

jcscottiii:
> (is there a way to set CI up to test PRs before merging?)

whoops. simple circleci mis-configuration. enabled now

aeijdenberg:
That was a whole lot harder than it should have been. :)  It wasn't possibly to simply drop-in the `assert.JSONEq()` as it would fail when we weren't comparing JSON.

Refactored to work, and rebased on current `master`.

PR #1231, opened 2017-09-20
move footnote annotation inside anchor tag

jcscottiii:
to be consistent with other footnote e-mails


rememberlenny:
Missed this one. Thank you.

PR #1235, opened 2017-10-03
Allow all environment variables to be set in common way

aeijdenberg:
This PR allows all environment variables to be sourced from the `dashboard-ups`, falling back to `os.LookupEnv()`. This should not affect the behavior of existing servers.

This PR also makes adding new environment variables require much less copy/paste by adding `.MustGet(key)` for mandatory fields, and `.BoolGet(key)` for `bool` fields and `.Get(key, defaultVal)` for fields that should have default values.


helpers/env_vars.go
```diff
@@ -53,27 +60,96 @@ var (
...
+	EnvVar string
+}
+
+// Error returns an error string
+func (err *ErrMissingEnvVar) Error() string {
+	return fmt.Sprintf("missing env variable: %s", err.EnvVar)
+}
+
+// NewEnvLookupFromMap creates a lookup based on a map
+func NewEnvLookupFromMap(m map[string]string) EnvLookup {
```
jcscottiii: this should be moved to the testhelpers package (it's only used in tests)

aeijdenberg:
Fixed as suggested, thanks for the review @jcscottiii .

helpers/env_vars.go
```diff
@@ -53,27 +60,88 @@ var (
...
-	SMTPPort        string
-	SMTPUser        string
-	SMTPPass        string
-	SMTPFrom        string
-	TICSecret       string
+	path []EnvLookup
+}
+
+// Get returns value for key if present, else returns defaultVal if not found
+func (el *EnvVars) Get(key, defaultVal string) string {
```
jonathaningram: @aeijdenberg I was just thinking:

- `String` was a better name than `Get`: so `env.String("FOO", "bar")`
- Then you have `env.Bool("ENABLE_THIS")`
- And `env.MustString("FOO")`

For me `env.String(...)`, `env.Bool(...)` are more intuitive (and consistent) from the outside than `env.Get(...)` and also matches the way the `flag` package does this.

Anyway, just a suggestion.

helpers/env_vars.go
```diff
@@ -53,27 +60,88 @@ var (
...
+	if !found {
+		panic(&ErrMissingEnvVar{EnvVar: key})
+	}
+	return rv
+}
+
+// BoolGet returns true if and only the value "true" or "1" is set for the key, else defaults to false
+func (el *EnvVars) BoolGet(key string) bool {
+	val, _ := el.load(key)
+	return val == "true" || val == "1"
```
jonathaningram: @aeijdenberg if you wanted to be a little more flexible here you could use `strconv.ParseBool`: https://golang.org/src/strconv/atob.go?s=351:391#L1

helpers/env_vars.go
```diff
@@ -53,27 +60,88 @@ var (
...
+		return defaultVal
+	}
+	return rv
+}
+
+// MustGet will panic if value is not set, otherwise it returns the value.
+func (el *EnvVars) MustGet(key string) string {
+	rv, found := el.load(key)
+	if !found {
+		panic(&ErrMissingEnvVar{EnvVar: key})
```
jonathaningram: @aeijdenberg you could remove the `EnvVar` stutter: 

```go
// ErrMissingEnvVar is panicked if a MustGet fails.
type ErrMissingEnvVar struct {
	// Name is the name of the env var that was not found
	Name string
}
```

Minor, but I used `name` not `key` - I thought the former was the correct terminology, but yeah it's minor.

helpers/env_vars.go
aeijdenberg: Fixed as suggested.

helpers/env_vars.go
aeijdenberg: Fixed as suggested - noting this is a slight change from previous behaviour - it now accepts a few additional values for `true`, and will now panic if set but not parsable as `true` or `false`.

helpers/env_vars.go
aeijdenberg: Fixed as suggested.

PR #1242, opened 2017-10-04
Fix `NewEnvLookupFromCFAppNamedService()` to work outside of CF

aeijdenberg:


jcscottiii:
Also, is there a use case for us to skip this check? The docker-compose.yml env vars help simulate the cfenv.App. (Your use-case could be that you don't use the docker containers but just want to clarify).

aeijdenberg:
Thanks for the review @jcscottiii - fixed to check the error code.

Yes I think the use-case is local development without needing Docker. I think that particularly with not needing Redis, and ignoring the SMTP part, it's convenient to be able to just run app without needing Docker (though now I've seen it, that file does look convenient).

PR #1243, opened 2017-10-04
app: add env panel with CRUD for user-defined env

jonathaningram:
Also move ErrorMessage to SystemErrorMessage.

The former ErrorMessage was specifically designed to say "The system
encountered an error" and also "try again". This is too specific for
ErrorMessage to be a generic component to render error messages.
Instead, the old ErrorMessage is moved to SystemErrorMessage and
ErrorMessage now simply renders the error message given to it.

Screenshots (we're in an app page):

image

System-defined:

image

User-defined:

image

Add (invalid):

image

Edit:

image

Delete:

image


static_src/components/env_panel.jsx
```diff
@@ -0,0 +1,219 @@
...
+              />
+            )}
+            {showAddForm && updateError && <ErrorMessage error={updateError} />}
+            <ComplexList>{this.renderUserVarItems()}</ComplexList>
+          </div>
+        )}
+        <Section styler={styler}>
+          <Header styler={styler}>System-defined environment variables</Header>
+          {this.renderShowHideAction(showEnv, this.handleShowEnvClick)}
+          {showEnv && (
```
el-mapache: Let's add a function here to make this more readable `this.displayEnvironment(showEnv)` or something along those lines.

static_src/components/env_panel.jsx
```diff
@@ -0,0 +1,219 @@
...
+              )}
+            </pre>
+          )}
+        </Section>
+      </div>
+    );
+  }
+}
+
+EnvPanel.propTypes = {
```
el-mapache: Here also, proptype defs at the top of the file, assignment at the bottom.

static_src/components/env_var_list_item.jsx
```diff
@@ -0,0 +1,155 @@
...
+import ElasticLineItem from './elastic_line_item.jsx';
+import Loading from './loading.jsx';
+import ErrorMessage from './error_message.jsx';
+import EnvVarForm from './env_var_form.jsx';
+
+export default class EnvVarListItem extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {};
```
el-mapache: Lets define the state shape here.

static_src/components/error_message.jsx
```diff
@@ -1,71 +1,53 @@
...
+      >
+        {this.renderMessage()}
       </div>
     );
   }
 }
 
-ErrorMessage.propTypes = propTypes;
-ErrorMessage.defaultProps = defaultProps;
+ErrorMessage.propTypes = {
```
el-mapache: Please toss these back up top.

static_src/stores/env_store.js
```diff
@@ -0,0 +1,58 @@
...
+    return {
+      ...error,
+      errorCode
+    };
+  }
+
+  _registerToActions(action) {
+    switch (action.type) {
+      case envActionTypes.ENV_FOR_APP_FETCH:
+        cfApi.fetchEnv(action.appGuid);
```
el-mapache: In general, we try to follow these guidelines for our stores: https://gist.github.com/twokul/9501770. 

We definitely have a few places where the code doesn't match this guide, and we are actively working to make those areas of the codebase follow the linked conventions!

static_src/components/env_var_form.jsx
```diff
@@ -0,0 +1,104 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import { generateId } from '../util/element_id';
+import PanelActions from './panel_actions.jsx';
+import Action from './action.jsx';
+
+export default class EnvVarForm extends Component {
```
el-mapache: Can this component make use of the existing `Form` component? Or is that component too restrictive for this use case? Let us know, so we can improve it!

static_src/stores/env_store.js
jonathaningram: @el-mapache I've had a read of that gist, but I'm not sure how it relates to the line of code you're referencing - it may just be me, but the gist talks about V8 internals but it doesn't seem to link it back to how one should write one's JS code so V8 is happier?

Can you give me a little more info?

static_src/components/env_var_form.jsx
jonathaningram: I didn't want to use use the existing `Form` component for a few reasons:

1. I don't really want to store this state in redux - it's local state that doesn't need to be in redux and I wanted to keep things simple.
2. I have some pending ideas that I'd like to propose for forms, inputs, labels, etc. which are also related to accessibility and I wanted to do that in a separate PR.
3. Because the validation for env vars is (arguably) so simple (only name is required), I haven't even bothered to do any client-side validation since the API picks up the error straight up if you try and submit with an empty name. Of course, I'm not saying we should never do it, but I just felt it was overkill at this stage for what is really just a basic two-input form.

Having said that, if you prefer to use it now I can mash it in, otherwise I'm happy to revisit in a subsequent form proposal 👍

jonathaningram:
> This is awesome! Thanks so much for your interest and contributions, our dedicated front-end team hovers between 1-2 people at any given time so these PRs are definitely appreciated.
> 
> I've left a few additional comments, mostly about style conventions. Additionally, it would be awesome if all components related to the env panel could be moved into a single folder. I've been trying to organize components into folders when I get time, and we'd like to continue doing that going forward.

Thanks so much for your _super_ responsive feedback @el-mapache ! I'm really excited to be helping out and contributing to the frontend.

I've addressed all of it and commented on two of your comments.

static_src/stores/env_store.js
el-mapache: Ha sorry! That was in my copy buffer, totally unrelated to this work. Here is the actual link: https://github.com/18F/cg-dashboard/blob/master/CONTRIBUTING.md#working-with-reactflux

This will make a lot more sense in context.

jonathaningram: Thanks, that makes more sense :) I've refactored the store. I think it's more along the lines of what you wanted and I am much more happy with it. I can see room for extracting some of this out as it could become a bit too boilerplate-y, but for now I think it's OK. See my commit message for other changes I made when I refactored the store.

el-mapache:
Looks good, thanks!

mogul:
💯  🎉  all around for international collaboration on common challenges via open source. The dream is alive, thanks so much 🇦🇺 ! 
❤️  🇺🇸

Issue #1234, opened 2017-10-03
Remove session backend dependency and other session management tidy-up


aeijdenberg:
We'd like to submit a PR to remove the dependency on Redis for a session backend. In addition to making the dashboard more complex to deploy, a database of refresh tokens is a juicy target and we'd like to avoid storing them if possible.

As background, the session currently contains basically just a serialized OAuth2 token.

In theory, we could use the gorilla session library (which is already a dependency) to simply sign and encrypt this value and rely on the browser to store it for us, however from a practical standpoint this currently fails to work because the serialized token is too large (greater than 4096 bytes) and thus won't work reliably between browsers (the gorilla library won't even try).

Newer versions of UAA (since May 2016) support [opaque tokens](https://www.cloudfoundry.org/opaque-access-tokens-cloud-foundry/) for both access and refresh tokens which are much smaller and if used make this small enough to store within the user session.

However, after experimenting with this, we found that while UAA (who issues the opaque tokens) can accept them, the CF API (`cloud_controller_ng`) currently cannot accept an opaque access token as it is built to decode and verify a JWT only and expects the access token to be in this form.

We can however use a hybrid approach. Request from UAA opaque tokens and receive an opaque refresh token and access token. We then discard the access token, and use the opaque refresh token to request a JWT access token (and discard the new JWT refresh token returned), and then store in the session the JWT access token along with the opaque refresh token, which together are small enough to remain below the session size limit giving us the best of both worlds.

Creating issue to track discussion around this - expect to see some PRs forth coming.

In addition the PRs will address a few other issues encountered along the way:

- Separation of the CSRF token key, a session authentication key, and a session encryption key.
- Actually update the access key in the session after a refresh token is fetched - currently once an access token is expired, every request will trigger a new access token be fetched, when it should be saved back into the session.
- Allow for all environment variables to be set via a user-provided-service - currently some are, some aren't, and allowing all makes deployment easier. Possibly it's distracting to put this in the same PR, but it sure made it easier for us to test so I'll likely start that way and we can split out if needed.

jcscottiii:
@aeijdenberg this is very exciting. Looking forward to the PRs!

aeijdenberg:
@jcscottiii - the first part if I've done it right should be entirely unexciting and can be found here in pull request #1235.

It's probably the largest change but will make the other bits easier. We found that using CUPS made it easier for us to deploy to multiple environments, so this PR allows all environment variables to be set in a common method.

[...]

PR #1244, opened 2017-10-04
Add support for opaque tokens and cookiestore "backend"

aeijdenberg:
This is the intended final part of this set of changes for https://github.com/18F/cg-dashboard/issues/1234.

It adds a `SESSION_BACKEND=cookiestore` which requires a `SESSION_ENCRYPTION_KEY` which in turn enables use of opaque refresh tokens, which is enough to keep our cookie at around 2922 bytes (under the IE limit of 4096).

This is probably easier reviewed once the other PRs are in, and if happy with the direction, would probably benefit from some end to end tests which I can look at adding tomorrow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,9 @@ cf create-user-provided-service dashboard-ups -p @<(cat <<EOF
   "CONSOLE_CLIENT_ID": "your-client-id",
   "CONSOLE_CLIENT_SECRET": "your-client-secret",
   "CSRF_KEY": "$(openssl rand -hex 32)",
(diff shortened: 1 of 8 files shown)


aeijdenberg:
I've added a commit that switches the `SESSION_BACKEND` to `cookiestore` in the tests, and verifies that the `/ping` returns as expected.

It's not clear the best way to add much additional testing (without mocking out a full UAA) - it's a pretty vanilla use of the sessions library, and I can't find other tests that check on cookie setting, or that test UAA token issuance (which is external to us anyway).

Let me know your thoughts.

PR #1251, opened 2017-10-05
Add ability to skip tests that require Docker

aeijdenberg:
Set env variable SKIP_DOCKER=1, and then tests that require Docker are skipped.

Why? In some CI environments it is problematic to allow Docker containers to be created, this change gives the operator the choice to accept the risk of these tests failing.


aeijdenberg:
I'm pretty sure that test breakage is actually unrelated to this change, however I'll note the irony that the test that failed has failed due to a Docker issue, which would have been skipped if the flag that this patch introduces had been enabled.

I'll push the same patch a 2nd time, to trigger another build to see if that was just a temporary flakiness or not.

Issue #1229, opened 2017-09-14
Consider using prettier to format JS (and also CSS and JSON files)


jonathaningram:
Prettier (https://github.com/prettier/prettier) is the defacto frontend code formatter with support for:

- JavaScript, including ES2017
- JSX
- Flow
- TypeScript
- CSS, LESS, and SCSS
- JSON
- GraphQL

It is essentially the equivalent of `gofmt` and removes a significant amount of cognitive load when writing and reading a codebase. Prettier has excellent editor support (e.g. format on save), and is also super easy to run over a codebase where there is no editor support. There are many other benefits that we could discuss, but I won't list them here.

Instead, I'll just ask: would you (and your project) be interested in adopting prettier? This would require some changes to your eslint settings, but style and formatting should (arguably) not be part of linting anyway (since you can just resolve it automatically by just running a formatter over the codebase).

Note: the aim is _to replace/remove_ the parts of Airbnb Javascript that are related to formatting and styling, _not_ the other "best practice" parts of it, for example "use const", "use the object spread operator over Object.assign"

Background: as Adam mentioned in https://github.com/18F/cg-dashboard/issues/1225#issuecomment-328736812 we're considering using this codebase as a frontend to one of our CF environments within cloud.gov.au.

jcscottiii:
@jonathaningram sorry for the delay. I've been asking around 18F about people's thoughts on this tool. (The frontend world isn't my expertise)

Overall, people said no to prettier. (it's relatively new in the JS world, worrying about how well it catches issues that are not just formatting, etc.)

Also, I did some research and saw things like this:
https://github.com/airbnb/javascript/issues/1307
https://github.com/airbnb/javascript/issues/1548

With all that said, I'm all for increasing dev productivity. I would open to seeing a PR. I would share it around internally and get thoughts on it. If all looks well, we could merge it! But warning, there's no guarantee that it would be accepted.
If you plan on trying it out, we can keep this issue open to track it.

jonathaningram:
@jcscottiii thanks for your response. As a semi aside: I have a PR for a frontend feature that we're currently reviewing internally. Will hopefully be PRing it soon. It uses prettier on all the new JS code written in it.

Prettier 1.0 has been out for a while and is ready for production. As for who is using it - most of the libs/tools in this cf-dashboard JS stack are using it:

![image](https://user-images.githubusercontent.com/626664/31149418-98f87eaa-a8db-11e7-8367-4ff4a278d004.png)

[...]

PR #1252, opened 2017-10-05
Add prettier and run on codebase

jonathaningram:
As I promised in #1229, this PR adds support for prettier but does not run it across the codebase. If this gets approved I can either run it across the codebase in a subsequent commit.

I really hope this gets accepted! 🤞 In my env var task, my dev workflow was hindered because I couldn't just hit `cmd+s` and go on my way - I had to deal with eslint complaining that I had wrong spacing here and there and had to manually fix those issues. By the way, the code that we pushed in my env var PR was actually all prettier formatted, I just didn't have the "format on save" enabled for the project.

Notes and comments:

1. `.prettierrc.json`: normally I'd not even have this file at all. The only reason I have it is so that I can override the `singleQuote` setting to match Airbnb. However, like `gofmt`, the point is that you don't need to configure it - you just format and forget. I'm super happy and _encourage_ us to remove this entirely and only use Airbnb as a base for non-style things. If you do approve this PR, please advise if you're happy for me to remove this option override (like `gofmt`, I just want to format things one way without thinking about having this particular option enabled for this particular project).
2. You can see how we get formatting for CSS, JS, JSX and JSON all "out of the box" - basically our entire frontend stack can now be auto-formatted.
3. We can format `.eslintrc` too (it's just JSON). In a separate PR I'll look to upgrade eslint and associated packages. I'd do this PR because eslint now recommends to use `.eslintrc.json` as the filename. And then once we do that, it will be another JSON file in the repo that will "magically" just become auto-formatted.
4. There are some unrelated spelling error fixes in a separate commit in this PR...

Look forward to hearing back and thanks! 😎


jonathaningram:
Thanks @jcscottiii awesome!!! I'll leave the actual running of the formatting until my other ones are merged so I can avoid conflicts.

jonathaningram:
@jcscottiii this is rebased and ready to merge. Will be no more conflicts unless you merge #1262 first, but it doesn't bother me which one is merged first.

Also note that prettier can now format markdown files too, hence the changes in this PR.

Also worth noting is that I've disabled a bunch of eslint rules for now as we transition and upgrade airbnb/javascript.

.eslintrc
```diff
@@ -8,12 +8,35 @@
...
+    "global-require": 0,
+    "import/imports-first": 0,
+    "import/newline-after-import": 0,
+    "import/no-extraneous-dependencies": 0,
+    "import/no-named-as-default": 0,
+    "import/no-unresolved": [2, { "ignore": ["skin"] }],
+    "import/prefer-default-export": 0,
+    "jsx-a11y/href-no-hash": 0,
+    "jsx-a11y/img-has-alt": 0,
+    "no-console": 0,
```
jcscottiii: some of these are fine during the transition but ones like this one `no-console` and ones like `prefer-const` `no-undef` seem like we should never have to turn these off.

jcscottiii:
@jonathaningram i think i need a better understanding in terms of what the path to re-enabling these rules are. there's quite a few.

also, is it just possible to use https://github.com/prettier/eslint-config-prettier which will stop the conflict between prettier and eslint without us having to manually turn off rules?

.eslintrc
jonathaningram: Yep we never want to turn those off. In other words, we should always prefer const, for example.

I have grouped all of the ones we need to revisit after my TODO comment.

jonathaningram:
> @jonathaningram i think i need a better understanding in terms of what the path to re-enabling these rules are. there's quite a few.
> 
> also, is it just possible to use https://github.com/prettier/eslint-config-prettier which will stop the conflict between prettier and eslint without us having to manually turn off rules?

@jcscottiii I think the most obvious path after this is merged is to just go through each rule one by one. Re-enable the rule, run the linter and fix the issues, then provide a PR. If a rule only results in one error maybe it's worth fixing more than one in that PR.

I could have tried to fix all of these in this PR, but I didn't want it to drag out too long so that's why I decided to disable the rules as part of a transition period. I'm pretty sure ones like `prefer-const` can even be fixed by running `eslint --fix` so some changes should be quick and automatic.

And yup, we're using `eslint-config-prettier` 😄  that is what is silencing all of the airbnb formatting rules. To be clear, the reason we need to silent these extra rules is that we _had_ to upgrade to a newer version of airbnb (not the latest one available yet, just a newer one). We had to upgrade because of the few eslint configs and plugins that also needed to update. So it was about finding a bunch of deps that would work together.

jonathaningram:
@jcscottiii ok have rebased this since the latest merge.

jonathaningram:
@jcscottiii just checking that you're not waiting on anything from me to merge this?

jcscottiii:
@jonathaningram I appreciate the work going into this PR. unfortunately, I don't want to merge it as-is because if we disable rules now and slowly turn them back on in later PRs, in between then, other developers (e.g. we have a new developer coming on our team) might come and start contributing bad code because of those turned off rules.

jonathaningram:
@jcscottiii I'm happy to disable the rules and fix the code that fails those rules, it just means the PR will be bigger and viewing changes might be hard. Having said that, if I fix the code that fails those rules in a separate commit in this PR you should be able to just view that single commit's changes.

Alternatively, I'm able to do a PR almost immediately after this is merged, so there'll really only be a small time in between. Up to you of course.

jcscottiii:
@jonathaningram  that's a good point. if it will be too bloated, create a bunch of smaller PRs that use this one as the base for the first one (first PR will use this as base, the PR after that will use the first PR as base, and so on to create a long chain of PRs. also it can be many chains if the changes aren't that dependent on one another) until all of those rules are re-enabled then I can review and merge them

jonathaningram:
@jcscottiii ok I decided to go ahead and just fix as many lint issues as I could. You'll notice I've still got some overrides in place. As follows:

```json
{
  "import/imports-first": 0,
  "import/no-extraneous-dependencies": 0,
  "import/no-named-as-default": 0,
  "react/jsx-filename-extension": 0
}
```

And test eslint file:

```json
{
  "jasmine/new-line-before-expect": 0,
  "jasmine/new-line-between-declarations": 0
}
```

1. `import/no-extraneous-dependencies`: it's because we are importing `skin` and `dashboard` modules that don't exist in package.json. This eslint rule doesn't have the feature to ignore project-level modules so we have no choice but to turn this off for now.
2. `import/no-named-as-default`: a bit difficult to fix in the codebase for this PR. The current codebase is already failing this test anyway, so I don't think it needs to be done in this PR.
3. `react/jsx-filename-extension`: we'll need to revisit and rename files accordingly. Note: pretty sure later airbnb versions have turned this off, so I don't want to make any changes on this right now, especially renaming all the files...
4. The 3 other rules I've turned off are just formatting so we can do this in a subsequent PR. 

Thanks for you time :)

jonathaningram:
@jcscottiii I've just rebased again after my previous merge.

jonathaningram:
@pburkholder I've rebased and Circle is passing. Are you able to merge this without Codeclimate passing? I don't think this was in place when I started this PR, plus this PR doesn't really change any code, it' just formatting it - I didn't want to go down a rabbit whole massive PR that refactored all the code :)

pburkholder:
Let's make the security-related checks go away, then we'll address the PR on its own merits.

And yes, this wasn't in place before.

pburkholder:
If you cherry-pick 47387115acd64535dacd135764c17c2b7664fb3f then this should be mergeable and addresses the CodeClimate security findings.

jonathaningram:
@pburkholder yep I'll look into it. Looks like a few commits/changes have gone into master related to package.json/npm-shrinkwrap.json so the rebase might be a bit intensive, but stay tuned.

pburkholder:
Thanks for the help Jonathan. We had to turn on CodeClimate, but it looks like we're really close to getting this in. I won't have time to work on this today, if you can take a stab at it, that would be great. P.

pburkholder:
Hi Jonathan -- how can I help get this across the finish line? Happy to pair if you want to.

jonathaningram:
@pburkholder I believe this is ready. There are still 27 CC issues but they are just odd rules like too many returns, not security issues. Let me know if I missed something?

jonathaningram:
@pburkholder yay!

PR #1255, opened 2017-10-09
app: add panel to list UPS instances

### LINKED ISSUE #1001: User provided services do not appear in dashboard
User provided services do not appear in the dashboard. Here, the dashboard app itself should have a `dashboard-ups` user provided service but does not.

screenshot from 2017-02-07 11-31-12

**Repro**
1. Create a user provided service `cf cups my-service`
1. Go to the space page in the dashboard

**Expected**
`my-service` should appear in the service instances panel

**Actual**
`my-service` is not listed.

jonathaningram:
Edit: fixes #1001

This panel appears under the "Services" panel:

image

There's no CRUD for this one so far: just listing out the UPSI that are bound to the app and those that are in the space:

image

It's worth noting that the second section is not "unbound UPSI" - it's showing all UPSI in the space that are _not_ bound to the app. In fact, it might be showing UPSI that are bound to _another_ app. It would be nice to make this a bit better, but we're limited by what the CF API returns (it doesn't tell us if the UPSI is bound). So without doing some transformations ourselves, I found the best way to do this quickly was to just label the second section as "Others in the space".

I imagine it's useful to know the GUID of the UPSI too - for now I've just put it in the `title` attribute in order to avoid cluttering the UI with GUIDs.

### DIFF
--- a/static_src/actions/env_actions.js
+++ b/static_src/actions/env_actions.js
@@ -1,6 +1,6 @@
 import AppDispatcher from '../dispatcher';
 import { envActionTypes } from '../constants';
-import cfApi from '../util/cf_api.js';
+import cfApi from '../util/cf_api';
 
 const fetchEnvSuccess = (appGuid, env) => {
   AppDispatcher.handleServerAction({
--- a/static_src/actions/upsi_actions.js
+++ b/static_src/actions/upsi_actions.js
@@ -11,8 +11,6 @@ export const fetchAllSuccess = items => {
   return Promise.resolve({ items });
 };
 
-fetchAllSuccess.exportTest = true;
-
 export const fetchAllFailure = err => {
   AppDispatcher.handleServerAction({
     type: upsiActionTypes.UPSI_FETCH_ALL_FAILURE
@@ -20,8 +18,6 @@ export const fetchAllFailure = err => {
 
   return Promise.resolve({ err });
 };
-
-fetchAllFailure.exportTest = true;
 
 export const fetchAllForSpaceSuccess = (spaceGuid, items) => {
   AppDispatcher.handleServerAction({
@@ -33,8 +29,6 @@ export const fetchAllForSpaceSuccess = (spaceGuid, items) => {
   return Promise.resolve({ spaceGuid, items });
 };
 
-fetchAllForSpaceSuccess.exportTest = true;
-
 export const fetchAllForSpaceFailure = (spaceGuid, err) => {
   AppDispatcher.handleServerAction({
     type: upsiActionTypes.UPSI_FETCH_ALL_FOR_SPACE_FAILURE,
@@ -43,8 +37,6 @@ export const fetchAllForSpaceFailure = (spaceGuid, err) => {
 
   return Promise.resolve({ spaceGuid, err });
 };
-
-fetchAllForSpaceFailure.exportTest = true;
 
 export default {
   fetchAll() {
--- a/static_src/actions/upsi_actions.js
+++ b/static_src/actions/upsi_actions.js
@@ -1,0 +1,67 @@
+import AppDispatcher from '../dispatcher';
+import { upsiActionTypes } from '../constants';
+import cfApi from '../util/cf_api';
+
+const fetchAllSuccess = items => {
+  AppDispatcher.handleServerAction({
+    type: upsiActionTypes.UPSI_FETCH_ALL_SUCCESS,
+    items
+  });
+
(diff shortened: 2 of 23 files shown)


static_src/components/upsi/upsi_panel.js
```diff
@@ -0,0 +1,95 @@
...
+          <ElasticLine key={guid}>
+            <ElasticLineItem title={`ID: ${guid}`}>{name}</ElasticLineItem>
+            <ElasticLineItem align="end" />
+          </ElasticLine>
+        ))}
+      </ComplexList>
+    );
+  }
+
+  renderOther() {
```
jcscottiii: renderOther is kind of generic.
change renderOther to something like renderUnbound

jonathaningram: How about `renderOthersInSpace`? It's not quite "unbound" per:

> It's worth noting that the second section is not "unbound UPSI" - it's showing all UPSI in the space that are not bound to the app. In fact, it might be showing UPSI that are bound to another app. It would be nice to make this a bit better, but we're limited by what the CF API returns (it doesn't tell us if the UPSI is bound). So without doing some transformations ourselves, I found the best way to do this quickly was to just label the second section as "Others in the space".

jonathaningram:
@jcscottiii ready for review again.

static_src/actions/env_actions.js
```diff
@@ -1,6 +1,6 @@
 import AppDispatcher from '../dispatcher';
 import { envActionTypes } from '../constants';
-import cfApi from '../util/cf_api.js';
```
jcscottiii: is there a reason your removed the .js?

jonathaningram:
> @jonathaningram you don't need to rethrow the error. In Redux, you create a matching error handling action.
> Now, if the actions aren't doing appropriate things, that's a different problem.

@jcscottiii see Dan Abramov's comments on this:

1. https://github.com/facebook/react/issues/7617#issuecomment-247710003
2. https://twitter.com/dan_abramov/status/776891811222675457?lang=en
3. https://twitter.com/dan_abramov/status/770914221638942720?lang=en

The first one is the best reference, and I'll paste in the most important paragraphs here:

Problem:

image

Even something as simple as an undefined variable gets swallowed using `.catch` which makes for very hard development workflow.

Solution:

image

I can update the code to not rethrow, but change to using the second argument to `.then` instead of using `.catch`. I was only rethrowing because I wanted to show the error and wanted to change the current behaviour as little as possible. This was/is essentially a stepping stone.

Related: that this problem is slightly improved in React 16, so we should look at upgrading in the near future.

jcscottiii:
@jonathaningram let's leave it for react 16 (just create an issue about it). since react 16 will kind of take care of those weird cases like undefined variables, i would like to keep the actions handling all other errors.

jonathaningram:
@jcscottiii ok reverted the error handling change. Ready for review.

static_src/actions/upsi_actions.js
```diff
@@ -0,0 +1,75 @@
...
+export const fetchAllSuccess = items => {
+  AppDispatcher.handleServerAction({
+    type: upsiActionTypes.UPSI_FETCH_ALL_SUCCESS,
+    items
+  });
+
+  return Promise.resolve({ items });
+};
+
+fetchAllSuccess.exportTest = true;
```
jcscottiii: what are these (this line and the other `exportTest`) needed for?

jonathaningram: I'll remove it for now. It's an idea I was toying with.

Rationale: testing JS modules often means you need to export certain things so that the test file can import them to test. Unlike in Go where it can test both exported and unexported members, with JS modules, we're forced to export all the things if we want to test them. However, I don't want callers to use these. E.g. `fetchAllSuccess` should never be directly called. So I toyed with the idea of setting `exportTest = true` as an indicator that it shouldn't be used. Of course, to actually enforce it we would need tooling: e.g. maybe eslint can error when it's imported and used. Also even though the bandwidth would be minimal, we could also use a babel plugin to strip them out. I didn't really look for existing solutions to this - I just made this up as I was writing the test.

Anyway, I'll just remove it for now, but it's something that we could revisit later out of interest.

jonathaningram: OK @jcscottiii removed.

jonathaningram:
@jcscottiii let me know if there's anything pending with this? I feel like we're good to go but I may have missed a change request.

PR #1257, opened 2017-10-10
Make breadcrumbs component dumb

jonathaningram:
This also fixes the "calling setState on an unmounted component"
warning.

Also, instead of using CSS classes to target DOM nodes in integration
tests, we use `data-test` attributes. We can later strip these out of
the build using a babel plugin, if desired. Note: there are other ways
of doing this, but for the time being the `data-test` attribute best
matches the current code with the least number of changes and is clear
enough. It's also a recommended approach by Kent C Dodds, if that holds
any weight.


static_src/components/app_container.jsx
```diff
@@ -7,7 +7,7 @@ import { appHealth, worstAppInstanceState } from '../util/health';
 import { appStates } from '../constants';
 import { config } from 'skin';
 import AppStore from '../stores/app_store.js';
-import Breadcrumbs from './breadcrumbs.jsx';
+import Breadcrumbs from './breadcrumbs';
```
jcscottiii: `breadcrumbs.js`?

jcscottiii: `./breadcrumbs/breadcumbs.js`?

static_src/components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+
+import { appPropType } from '../../stores/app_store';
+import { orgPropType } from '../../stores/org_store';
+import { spacePropType } from '../../stores/space_store';
+import Item from './breadcrumbs_item';
```
jcscottiii: Let's keep the name the same as what's in the imported file.
`Item` -> `BreadcrumbsItem`

static_src/stores/org_store.js
```diff
@@ -1,15 +1,16 @@
-
-/*
```
jcscottiii: is there a reason you got rid of this?

jonathaningram: @jcscottiii I will revert - I just removed it because it didn't really tell us anything we don't already know: `org_store.js` is a store for org data :)

static_src/components/breadcrumbs/breadcrumbs.js
jonathaningram: @jcscottiii I was just trying to remove some stutter but I do concede that keeping `BreadcrumbsItem` makes searching easier (just at the expense of stutter) - but all good, I will update.

jonathaningram:
> Just two things! Otherwise, LGTM. thanks for this!

Thanks so much for your review - changes made, ready for review again.

PR #1259, opened 2017-10-11
Move Go templates into `templates dir for skinning

jonathaningram:
By moving all the templates into the one directory, we can now override
these templates in our skins by setting the TEMPLATES_PATH environment
variable.

For example in my local instance, the env var looks like:

```sh
export TEMPLATES_PATH="/path/to/our/skin/templates"
```

This change also allows us to fix tests that were trying to construct paths
with BASE_PATH when using `testdata/*` directly is enough (because Go
runs its tests with the current working directory the same as the
current file under test).

Also moved `index.html` to `web` subdir so that it matches the `mail` subdir.


helpers/env_vars.go
```diff
@@ -50,6 +50,8 @@ var (
 	LegacySessionKeyEnvVar = "SESSION_KEY"
 	// BasePathEnvVar is the path to the application root
```
jcscottiii: I think we can get rid of BasePath now :)

helpers/settings.go
```diff
@@ -44,6 +45,8 @@ type Settings struct {
 	LogURL string
 	// Path to root of project.
```
jcscottiii: same as https://github.com/18F/cg-dashboard/pull/1259/files#r144411138

helpers/env_vars.go
jonathaningram: Unfortunately I don't think we can. The reason is that `CreateRouterWithMockSession` tries to load the templates. Since `CreateRouterWithMockSession` is called by an array of tests in various packages, the location of the templates directory is not consistent. Recall this is because when `go test` runs your package tests it uses the working directory of the current test. E.g. when running `...cg-dashboard/controllers` tests, the CWD is `/Users/me/go/src/github.com/18F/cg-dashboard/controllers` and so we need that base path variable to be able to consistently locate the templates directory.

I'm sure we could solve this another way. For example the tests in the `controllers` package could pass a path like `../templates`, but I just tried this out and I think it's more trouble than it's worth. Let me know if you want me to try and solve it.

Actually, come to think of it, this env var is only for testing, so I think I'm going to prefix it with `TEST_` to make it perfectly clear why it exists and that the main app does not need to use it.

jonathaningram:
@jcscottiii OK this one is ready for review again - see my comment above in answer to your question and let me know what you think.

PR #1272, opened 2017-10-28
Fix incorrect loading state calculation in app page

jonathaningram:
The app page is nested under an org and a space.

Thus the `loading` flag should be set to true if any of the org, space or app are loading (plus if the quota is loading).

We want this because we don't want the `AppContainer` component to try and render anything that depends on the org, space or app if those objects are not yet loaded.

By the way, this is necessary as part of the upgrade to React 16.


jcscottiii:
thanks @jonathaningram !

PR #1274, opened 2017-10-31
Use non-US dates

jonathaningram:
Pulls out the non-US dates change from https://github.com/18F/cg-dashboard/pull/1256

### DIFF
--- a/static_src/util/format_date.js
+++ b/static_src/util/format_date.js
@@ -3,6 +3,15 @@ import moment from 'moment-timezone';
 const makeInvalidDateTimeValueError = str =>
   new Error(`could not format invalid date/time value: ${str}`);
 
+/**
+ * formatDateTime returns a formatted date/time string.
+ *
+ * @params {String} date/time as a string
+ * @params {String} timezone abbreviation - defaults to "Etc/UTC"
+ *
+ * @returns {String} A "MMM DD YYYY hh:mma z" formatted string
+ *                   e.g. "Jul 13 2015 09:02pm PDT"
+*/
 const formatDateTime = (str, timezone = 'Etc/UTC') => {
   if (!str) {
     throw makeInvalidDateTimeValueError(str);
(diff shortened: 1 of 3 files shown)


static_src/util/format_date.js
```diff
@@ -1,18 +1,17 @@
 import moment from 'moment-timezone';
 
-const invalidMsg = 'Invalid datetimes cannot be formatted.';
+const makeInvalidDateTimeValueError = str =>
+  new Error(`could not format invalid date/time value: ${str}`);
 
-/**
```
jcscottiii: no need to remove valid comments

static_src/util/format_date.js
```diff
@@ -1,18 +1,17 @@
...
- * Returns a formated datetime string
- * @params {String} datetime as a string
- * @params {String} timezone abbreviation - defaults to UTC
- *
- * @returns {String} A "MM/DD/YYYY H:mma z" formatted string
- *                    i.e. 03/21/2016 10:39am PDT
-*/
-export default function formatDateTime(dateString, timezone = 'Etc/UTC') {
-  const d = moment(dateString);
+const formatDateTime = (str, timezone = 'Etc/UTC') => {
```
jcscottiii: `str` -> `dateString`

static_src/util/format_date.js
jonathaningram: I can revert. One of the reason these kinds of comments are not great is that they become outdated unless updated when the function is updated. When I went to update the comment I realised I was describing what the ~4 line function was doing, so it seemed redundant, so I removed it.

static_src/util/format_date.js
jonathaningram: I named it `str` because technically the caller can pass any string that is not a date, e.g. `formatDateTime('random non date string')` and `str` encapsulates that an arbitrary string was passed. Per https://golang.org/src/time/format.go?s=23626:23672#L762

I can change it though.

jcscottiii: Thanks for that. I can see the case for leaving it as str. Ignore this and just bring back the comment

jonathaningram:
@jcscottiii OK, reinstated the comment. Ready for review again!

PR #1275, opened 2017-11-01
Rename CF_SKIN to SKIN_NAME

jonathaningram:
The CF prefix is redundant. Env var will also then match `SKIN_PROVIDES_TRANSLATIONS` if we add that.

Cherry-picked from https://github.com/18F/cg-dashboard/pull/1248

Because the default (`cg`) skin does not need to provide translation files, the `SKIN_NAME` (formerly `CF_SKIN`) variable has been passed into `process.env`. This is deliberate because now any i18n module can remove dead code at build time. Particularly, if we use i18next, the i18next-xhr-backend module can be entirely removed for cloud.gov.

See https://github.com/18F/cg-dashboard/pull/1248#issue-263012641

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,10 +3,11 @@ const webpack = require('webpack');
 
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 
-const PRODUCTION = (process.env.NODE_ENV === 'prod');
-const TEST = (process.env.NODE_ENV === 'test');
+const PRODUCTION = process.env.NODE_ENV === 'prod';
+const TEST = process.env.NODE_ENV === 'test';
 const CG_STYLE_PATH = process.env.CG_STYLE_PATH;
-const CF_SKIN = process.env.CF_SKIN || 'cg';
+
+const SKIN_NAME = process.env.SKIN_NAME || 'cg';
 
 const srcDir = './static_src';
 const compiledDir = './static/assets';
@@ -14,10 +15,7 @@ const compiledDir = './static/assets';
 const config = {
   bail: false,
 
-  entry: [
-    'babel-polyfill',
-    `${srcDir}/main.js`
-  ],
+  entry: ['babel-polyfill', `${srcDir}/main.js`],
 
   output: {
     path: path.resolve(compiledDir),
@@ -49,7 +47,8 @@ const config = {
         test: /\.(svg|ico|png|gif|jpe?g)$/,
         loader: 'url-loader?limit=1024&name=img/[name].[ext]'
       },
-      { test: /\.(ttf|woff2?|eot)$/,
+      {
+        test: /\.(ttf|woff2?|eot)$/,
         loader: 'url-loader?limit=1024&name=font/[name].[ext]'
       }
     ]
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -3,21 +3,19 @@ const webpack = require('webpack');
...
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 
-const PRODUCTION = (process.env.NODE_ENV === 'prod');
-const TEST = (process.env.NODE_ENV === 'test');
+const PRODUCTION = process.env.NODE_ENV === 'prod';
+const TEST = process.env.NODE_ENV === 'test';
 const CG_STYLE_PATH = process.env.CG_STYLE_PATH;
-const CF_SKIN = process.env.CF_SKIN || 'cg';
+
+process.env.SKIN_NAME = process.env.SKIN_NAME || 'cg';
```
jcscottiii: You can make this as a const variable like the rest of the variables instead of overwriting `process.env.SKIN_NAME`.

In regards to your comment in the PR description, you make it globally available via `webpack.DefinePlugin` down below.

jonathaningram: @jcscottii by the way, the significance of setting directly into `process.env.SKIN_NAME` is that we want to treat the non-existence of the env var as if it was set to `cg`. This is because we always have to have a skin. I never want to get to a point in the codebase where I have to do `process.env.SKIN_NAME || "cg"`. Thus to avoid this, at the very top we make sure it's always set to something.

With that, I've changed to this:

```js
process.env.SKIN_NAME = process.env.SKIN_NAME || 'cg';
const { SKIN_NAME } = process.env;
```

However, the local var is probably a bit redundant because `process.env.SKIN_NAME` is pretty short as it is. Either way, I've made the change.

jcscottiii: @jonathaningram i'm just wondering if we can set it once via the `DefinePlugin` that we set below. more docs [here](https://webpack.js.org/plugins/define-plugin/#usage). If it works like i think, once it's there, we don't need to keep redefining the default skin in process.env because it's globally available. BUT i'm not sure if I'm missing something. that's why I asked:

> is there no way for i18next to use the globally exposed variable exposed via the DefinePlugin function below?

the only reason, I can anticipate this not working is if DefinePlugin doesn't take affect by the time `i18next` does it thing.
if i'm correct, you can get rid of those two lines and just have: `const SKIN_NAME = process.env.SKIN_NAME || 'cg';`
if i'm incorrect and i18next can't use the global from DefinePlugin, what you had beforehand was better (now, like you mention, it's redundant).

jonathaningram:
@jcscottiii ok I have changed per your comments.

To answer your question - the define plugin will run before any user land code, so yup the internationalisation code will see the resolved/replaced environment variable value.

BTW, the rationale for my change was the following. At the moment, in webpack.config.js, we are doing the following:

1. Pull out some env vars into local variables/consts. E.g. `const X = process.env.X`
2. Use local variables in webpack.config.js (e.g. pass `X` to a loader)
3. Put the local variable back into `process.env` by way of the define plugin.

I wanted to circumvent these 3 roundabout steps and always just use `process.env` as the source of truth rather than defining local variables. But anyway, that was just my rationale, it's not an issue 😎

PR #1278, opened 2017-11-08
api: don't handle response code in submitRequest

jonathaningram:
End handlers should be able to set the HTTP response code. E.g.
logMessageResponseHandler should be able to set the response code to 500
if there is an error.

### DIFF
--- a/controllers/log.go
+++ b/controllers/log.go
@@ -31,10 +31,12 @@ func (c *LogContext) RecentLogs(rw web.ResponseWriter, req *web.Request) {
 func (c *LogContext) logMessageResponseHandler(rw http.ResponseWriter, response *http.Response) {
 	messages, err := c.ParseLogMessages(&(response.Body), response.Header.Get("Content-Type"))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


jcscottiii:
@jonathaningram this looks good. as a side note, are you using the /log endpoints successfully now? in our cloud foundry setup loggregator is gone and we need to pull logs from doppler now so right now logs do not work in our dashboard deployment. there's a need to move to support that instead but if you're using loggregator, we will need to support that too instead of ripping that out.

jonathaningram:
> @jonathaningram this looks good. as a side note, are you using the /log endpoints successfully now? in our cloud foundry setup loggregator is gone and we need to pull logs from doppler now so right now logs do not work in our dashboard deployment. there's a need to move to support that instead but if you're using loggregator, we will need to support that too instead of ripping that out.

@jcscottiii Correct, it does not work for us either. In my local branch I've made a change to indicate to the user that logs aren't being fetched (previously it just silently failed which I didn't like). Haven't PRd it yet though:

image

PR #1281, opened 2017-11-13
mail: allow sending a text version of mail

jonathaningram:
Extracted from https://github.com/18F/cg-dashboard/pull/1280

Cleaned up the mailer tests.


jcscottiii:
@jonathaningram this looks good!
i have a question: we are thinking about moving completely to text only e-mails and ripping out the HTML template. now if you still want the html template, we can keep it. Just want to plan accordingly.

mogul:
I think we want to keep HTML mails, just make the default [no frills](https://www.gkogan.co/blog/dont-design-emails/)!

jonathaningram:
@jcscottiii yep I think we should keep the ability to send HTML emails too.

On a semi-related note (because you mentioned that you might be making a change to sending text-only emails), you might like to take a look at these links for a PR that I have yet to make, but probably will at some stage:

- https://github.com/18F/cg-dashboard/commit/d919da5655fea68e2cbbb8d5d008f0c93642420f
- https://github.com/govau/ci-console/blob/master/skin/email-templates/src/emails/UserAssociatedWithOrganization.js
- https://github.com/govau/emailtemplate

The first link shows how I've incorporated a new `emailtemplate` package into the dashboard so that skins can provide the HTML, text and subject as separate files (basically, how to get the hard-coded "cloud.gov" subject out of Go code).

The second link shows how we are actually building our HTML email templates - in short, we use React to write out a HTML file, and the node stack also writes out the subject and text files.

The third link shows the actual Go package. It has documentation/examples so hopefully it's clear once you click through.

Let me know what you think!

PR #1297, opened 2017-12-09
Update for CodeClimate2, provide test guidance

pburkholder:
This repo can now be tested with `codeclimate` locally.

Also, added some guidance to CONTRIBUTING.md on how to do
testing locally.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,8 +1,6 @@
 ---
 version: 2
 plugins:
-  duplication:
-    enabled: false
   govet:
     enabled: true
   golint:
@@ -18,11 +16,6 @@ plugins:
     enabled: false
   csslint:
     enabled: false
-
-ratings:
-  paths:
-  - "**.go"
-  - "**.js"
 
 exclude_patterns:
   - static_src/test/*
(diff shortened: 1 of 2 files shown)


.codeclimate.yml
```diff
@@ -1,6 +1,6 @@
 ---
-engines:
-
+version: 2
+plugins:
   duplication:
```
LinuxBozo: `duplication` is one of the things that is now part of the stock checks in version 2, so you should be able to remove this.

.codeclimate.yml
```diff
@@ -24,5 +24,15 @@ ratings:
   - "**.go"
```
LinuxBozo: Not specific to this line, but the section `ratings`. This has also been deprecated in version 2 and can be removed. All files are rated now by default.

.codeclimate.yml
```diff
@@ -1,6 +1,8 @@
 ---
 version: 2
 plugins:
+  duplication:
```
LinuxBozo: Instead of re-enabling this, you'll want to disable these checks for version 2.
```
checks:
  similar-code:
    enabled: false
  identical-code:
    enabled: false
```

https://docs.codeclimate.com/docs/duplication-concept

PR #1303, opened 2018-01-30
Update button label and in-app docs for restage button

brittag:
[From support chat](https://gsa-tts.slack.com/archives/C09CR1Q9Z/p1517246989000096): "looks like a restart actually calls restage in the CF api: https://github.com/18F/cg-dashboard/blob/46216daafa964c3faffa28af9837e09c87d46eac/static_src/util/cf_api.js#L386-L391 ." "So the web `Restart app` is a `cf restage`, and `cf restart` is a different action?"

That's confusing, so this is a tiny effort to fix the label on the button. I didn't test this locally, just poked around to edit the text strings.

cc @LinuxBozo @konklone

### DIFF
--- a/static_src/components/app_container.jsx
+++ b/static_src/components/app_container.jsx
@@ -156,7 +156,7 @@ export default class AppContainer extends React.Component {
     let loading;
 
     let handler = this.handleRestart;
-    let actionText = "Restart app";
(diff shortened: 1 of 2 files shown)


konklone:
I didn't test this locally or anything, but :+1: to the intent.

PR #1326, opened 2020-01-24
Update config.yml

hillaryj:
Update circleci's config.yml from the migrated v1->v2 to one that actually builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
     parallelism: 1
 
     docker:
-      - image: circleci/golang:1.9-node-browser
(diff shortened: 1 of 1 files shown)


pburkholder:
I spent a _lot_ of time trying to get cg-dashboard to build with a lightweight image and ended up pretty despondent in the process. The project is an ugly mongrel with a lot of code that depends on Go conventions, and Node conventions, and I found that starting with either a Go or a Node image got me nowhere. (That my CircleCI skills are pretty rudimentary didn't help any)

Given that cg-dashboard will probably only have 6 more builds remaining in its life, if it's possible to continue with a massive all-in-one docker image that might be the way to go.

